PR #1, opened 2025-01-09
tests: add integration tests

sd2k:
This updates the Grafana service in the docker-compose file to have some
dashboards and datasources provisioned, then adds some integration tests
for each tool to the test suite. These can be run by adding the
`--integration-tests` flag to the `pytest` invocation.

There are also tests for cloud-only functionality (i.e. Grafana Incident)
which can be run by adding the `--cloud-only` flag to the `pytest`
invocation. ~These aren't currently run in CI as I need to set up a suitable cloud instance.~

### DIFF
--- a/.github/workflows/cloud.yml
+++ b/.github/workflows/cloud.yml
@@ -34,5 +34,6 @@ jobs:
       - name: Run tests
         run: uv run pytest tests --cloud
         env:
-          GRAFANA_URL: ${{ secrets.CLOUD_GRAFANA_URL }}
+          # E.g. https://mcptests.grafana-dev.net.
+          GRAFANA_URL: ${{ vars.CLOUD_GRAFANA_URL }}
           GRAFANA_API_KEY: ${{ secrets.CLOUD_GRAFANA_API_KEY }}


.github/workflows/cloud.yml
```diff
@@ -0,0 +1,38 @@
...
+        run: uv sync --all-extras --all-groups
+
+      - name: Run ruff
+        run: uv run ruff check .
+
+      - name: Run tests
+        run: uv run pytest tests --cloud
+        env:
+          GRAFANA_URL: ${{ secrets.CLOUD_GRAFANA_URL }}
+          GRAFANA_API_KEY: ${{ secrets.CLOUD_GRAFANA_API_KEY }}
```
annanay25: can we add a comment on which instance this is? will be helpful in debugging.

tests/tools/incident_test.py
```diff
@@ -0,0 +1,22 @@
+import json
+
+from mcp_grafana.tools.incident import list_incidents, ListIncidentsArguments
+
+from . import mark_cloud
+
+pytestmark = mark_cloud
```
annanay25, line 9:
```suggestion
# All tests in this file require a cloud Grafana instance and
# an asynchronous event loop.
pytestmark = mark_cloud
```

just for completeness!

.github/workflows/cloud.yml
sd2k: Done, switched to a variable instead of a secret too :+1:

PR #6, opened 2025-01-10
tests: add more varied tests for Prometheus queries

sd2k:
Also fixes some minor issues along the way.

### DIFF
--- a/src/mcp_grafana/tools/prometheus.py
+++ b/src/mcp_grafana/tools/prometheus.py
@@ -22,8 +22,8 @@ async def query_prometheus(
     datasource_uid: str,
     expr: str,
     start_rfc3339: str,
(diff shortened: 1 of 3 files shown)


src/mcp_grafana/tools/prometheus.py
```diff
@@ -38,8 +38,13 @@ async def query_prometheus(
     step_seconds: The time series step size in seconds. Ignored if `query_type` is 'instant'.
     query_type: The type of query to use. Either 'range' or 'instant'.
     """
+    if query_type == "range" and (end_rfc3339 is None or step_seconds is None):
```
annanay25, line 41: hmmm. should we add some checks on `expr`/`start_rfc3339` for an instant query too?

PR #8, opened 2025-01-10
refactor: rename some Prometheus-based tools to 'list' rather than 'get'

sd2k:
Several of these tools actually return lists of things, so for
consistency let's rename them to 'list' rather than 'get'.

This also changes some of the tools to have default values for the
'limit' parameter to limit the amount of data returned to the LLM.

### DIFF
--- a/src/mcp_grafana/tools/prometheus.py
+++ b/src/mcp_grafana/tools/prometheus.py
@@ -137,7 +137,7 @@ async def list_prometheus_label_names(
     matches: Optionally, a list of label matchers to filter the results by.
     start: Optionally, the start time of the time range to filter the results by.
     end: Optionally, the end time of the time range to filter the results by.
-    limit: Optionally, the maximum number of results to return.
+    limit: Optionally, the maximum number of results to return. Defaults to 100.
     """
     response = await grafana_client.list_prometheus_label_names(
         datasource_uid,
@@ -167,7 +167,7 @@ async def list_prometheus_label_values(
     matches: Optionally, a list of selectors to filter the results by.
     start: Optionally, the start time of the query.
     end: Optionally, the end time of the query.
-    limit: Optionally, the maximum number of results to return.
+    limit: Optionally, the maximum number of results to return. Defaults to 100.
     """
     response = await grafana_client.list_prometheus_label_values(
         datasource_uid,
--- a/tests/tools/prometheus_test.py
+++ b/tests/tools/prometheus_test.py
@@ -48,14 +48,14 @@ async def test_list_prometheus_metric_metadata():
 
 
 async def test_list_prometheus_metric_names():
-    # Test listing list of available metric names
+    # Test listing available metric names
     metric_names = await list_prometheus_metric_names("robustperception", ".*")
     assert isinstance(metric_names, list)
     assert len(metric_names) > 0
 
 
 async def test_list_prometheus_label_names():
-    # Test listting list of label names for a metric
+    # Test listing label names for a metric
     label_names = await list_prometheus_label_names(
         "robustperception",
         [Selector(filters=[LabelMatcher(name="job", value="node")])],
@@ -66,7 +66,7 @@ async def test_list_prometheus_label_names():
 
 
 async def test_list_prometheus_label_values():
-    # Test listting values for a specific label
+    # Test listing values for a specific label
     label_values = await list_prometheus_label_values("robustperception", "job")
     assert isinstance(label_values, list)
     assert len(label_values) > 0


src/mcp_grafana/tools/prometheus.py
```diff
@@ -87,39 +89,46 @@ async def get_prometheus_metric_metadata(
     )
 
 
-async def get_prometheus_metric_names(
+async def list_prometheus_metric_names(
     datasource_uid: str,
     regex: str,
+    limit: int = 10,
```
annanay25, line 95: Wondering if we should retain the default behaviour of this particular function to be the same as the Prometheus API, ie. return all results by default. Maybe we can set default to `-1`? That would ensure it works as-expected for everyone to begin with.

src/mcp_grafana/tools/prometheus.py
```diff
@@ -130,7 +139,7 @@ async def get_prometheus_label_names(
     end: Optionally, the end time of the time range to filter the results by.
     limit: Optionally, the maximum number of results to return.
```
annanay25:
```suggestion
    limit: Optionally, the maximum number of results to return. Defaults to 100.
```

src/mcp_grafana/tools/prometheus.py
```diff
@@ -160,7 +169,7 @@ async def get_prometheus_label_values(
     end: Optionally, the end time of the query.
     limit: Optionally, the maximum number of results to return.
```
annanay25:
```suggestion
    limit: Optionally, the maximum number of results to return. Defaults to 100.
```

tests/tools/prometheus_test.py
```diff
@@ -41,23 +41,22 @@ async def test_query_prometheus_range(step_seconds):
...
-    metadata = await get_prometheus_metric_metadata("robustperception", 10)
+    metadata = await list_prometheus_metric_metadata("robustperception", 10)
     assert 0 < len(metadata) <= 10
 
 
-async def test_get_prometheus_metric_names():
-    # Test getting list of available metric names
-    metric_names = await get_prometheus_metric_names("robustperception", ".*")
+async def test_list_prometheus_metric_names():
+    # Test listing list of available metric names
```
annanay25:
```suggestion
    # Test listing available metric names
```

tests/tools/prometheus_test.py
```diff
@@ -41,23 +41,22 @@ async def test_query_prometheus_range(step_seconds):
...
     assert isinstance(metric_names, list)
     assert len(metric_names) > 0
-    assert "up" in metric_names  # 'up' metric should always be present
 
 
-async def test_get_prometheus_label_names():
-    # Test getting list of label names for a metric
-    label_names = await get_prometheus_label_names(
+async def test_list_prometheus_label_names():
+    # Test listting list of label names for a metric
```
annanay25:
```suggestion
    # Test listing label names for a metric
```

tests/tools/prometheus_test.py
```diff
@@ -66,8 +65,8 @@ async def test_get_prometheus_label_names():
     assert "instance" in label_names  # 'instance' is a common label
 
 
-async def test_get_prometheus_label_values():
-    # Test getting values for a specific label
-    label_values = await get_prometheus_label_values("robustperception", "job")
+async def test_list_prometheus_label_values():
+    # Test listting values for a specific label
```
annanay25:
```suggestion
    # Test listing values for a specific label
```

tests/tools/prometheus_test.py
```diff
@@ -41,23 +41,22 @@ async def test_query_prometheus_range(step_seconds):
...
 
-async def test_get_prometheus_metric_names():
-    # Test getting list of available metric names
-    metric_names = await get_prometheus_metric_names("robustperception", ".*")
+async def test_list_prometheus_metric_names():
+    # Test listing list of available metric names
+    metric_names = await list_prometheus_metric_names("robustperception", ".*")
     assert isinstance(metric_names, list)
     assert len(metric_names) > 0
-    assert "up" in metric_names  # 'up' metric should always be present
```
annanay25, line 55: i thought this was a nice assertion, but all good

sd2k: It doesn't appear now we have the limit 😅

src/mcp_grafana/tools/prometheus.py
sd2k: I'm slightly concerned about blowing up the context window if we return everything... maybe we could provide a more high level tool which does more interesting filtering though/pagination?

annanay25: Ok I think its fine to proceed with the limit then 👍

PR #11, opened 2025-01-16
ci: remove publishing to testpypi

sd2k:
We can't run uv publish on every commit because the version in
pyproject.toml doesn't change that frequently, and the command
fails if we try to upload different files to the previous version.

The only way this would work is if we changed the version prior
to running uv publish, which would result in us running out of
space in testpypi, so just don't bother now we have the publish
pipeline working.


annanay25 (approved):
sounds good 👍

PR #19, opened 2025-02-06
feat: add Dockerfile for mcp-grafana

sd2k:
The Docker image uses the SSE transport by default and
exposes port 8000.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,0 +1,28 @@
+FROM python:3.13-slim AS builder
+COPY --from=ghcr.io/astral-sh/uv:latest /uv /uvx /bin/
+
+# Change the working directory to the `app` directory
+WORKDIR /app
+
+# Install dependencies
+RUN --mount=type=cache,target=/root/.cache/uv \
+  --mount=type=bind,source=uv.lock,target=uv.lock \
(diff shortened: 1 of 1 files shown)


sd2k:
We probably want a GitHub action to push this to the Docker hub too, I'll add that in a separate PR.

Dockerfile
```diff
@@ -0,0 +1,28 @@
...
+ADD . /app
+
+# Sync the project
+RUN --mount=type=cache,target=/root/.cache/uv \
+  uv sync --frozen --no-editable
+
+FROM python:3.13-slim
+
+# Copy the environment, but not the source code
+COPY --from=builder --chown=app:app /app/.venv /app/.venv
```
annanay25, line 23: curious why its better to do it this way?

annanay25 (approved):
lgtm, tested on a macbook

Dockerfile
sd2k: I followed (*cough* copied) the example [here](https://docs.astral.sh/uv/guides/integration/docker/#non-editable-installs). Because we've done a non-editable install, we don't need to include the code itself in the final image, we can just copy the venv directly

annanay25: i see, nice 👍  also helps keep the size of the image small I guess

PR #21, opened 2025-02-06
feat: add optional SSE middleware to obtain Grafana info from headers

sd2k:
This PR adds a flag to the CLI which, when combined with the
SSE transport, allows the client to include information about
the Grafana instance to use in headers.

This is a non-standard way to use the MCP server, but I'm not sure any
of the other accepted auth proposals meet our use case.

Unfortunately it involves changing some of the callsites because the
Grafana client is no longer a singleton (since it can be changed).
Instead, settings live in a contextvar and the client is recreated
for each request.

### DIFF
--- a/src/mcp_grafana/cli.py
+++ b/src/mcp_grafana/cli.py
@@ -1,4 +1,5 @@
 import enum
+from types import MethodType
 
 import typer
 
@@ -13,5 +14,59 @@ class Transport(enum.StrEnum):
 
 
 @app.command()
-def run(transport: Transport = Transport.stdio):
+def run(transport: Transport = Transport.stdio, header_auth: bool = False):
+    if transport == Transport.sse and header_auth:
+        # Monkeypatch the run_sse_async method to inject a Grafana middleware.
+        # This is a bit of a hack, but fastmcp doesn't have a way of adding
+        # middleware. It's not unreasonable to do this really, since fastmcp
+        # is just a thin wrapper around the low level mcp server.
+        mcp.run_sse_async = MethodType(run_sse_async, mcp)
+
     mcp.run(transport.value)
+
+
+async def run_sse_async(self) -> None:
+    """
+    Run the server using SSE transport, with a middleware that extracts
+    Grafana authentication information from the request headers.
+
+    The vast majority of this code is the same as the original run_sse_async
+    method (see https://github.com/modelcontextprotocol/python-sdk/blob/44c0004e6c69e336811bb6793b7176e1eda50015/src/mcp/server/fastmcp/server.py#L436-L468).
+    """
+
+    from mcp.server.sse import SseServerTransport
+    from starlette.applications import Starlette
+    from starlette.routing import Mount, Route
+    import uvicorn
+
+    from .middleware import GrafanaMiddleware
+
+    sse = SseServerTransport("/messages/")
+
+    async def handle_sse(request):
+        async with GrafanaMiddleware(request):
+            async with sse.connect_sse(
+                request.scope, request.receive, request._send
+            ) as streams:
+                await self._mcp_server.run(
+                    streams[0],
+                    streams[1],
+                    self._mcp_server.create_initialization_options(),
+                )
+
+    starlette_app = Starlette(
+        debug=self.settings.debug,
+        routes=[
+            Route("/sse", endpoint=handle_sse),
+            Mount("/messages/", app=sse.handle_post_message),
+        ],
+    )
+
+    config = uvicorn.Config(
+        starlette_app,
+        host=self.settings.host,
+        port=self.settings.port,
+        log_level=self.settings.log_level.lower(),
+    )
+    server = uvicorn.Server(config)
+    await server.serve()
--- a/src/mcp_grafana/middleware.py
+++ b/src/mcp_grafana/middleware.py
@@ -1,0 +1,54 @@
+from dataclasses import dataclass
+
+from starlette.datastructures import Headers
+
+from .settings import GrafanaSettings, grafana_settings
+
+
+@dataclass
+class GrafanaInfo:
+    """
+    Simple container for the Grafana URL and API key.
+    """
+
+    authorization: str
+    url: str
+
+    @classmethod
+    def from_headers(cls, headers: Headers) -> "GrafanaInfo | None":
+        if (url := headers.get("X-Grafana-URL")) is not None and (
+            key := headers.get("X-Grafana-API-Key")
+        ) is not None:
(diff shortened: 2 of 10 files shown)


sd2k:
This would definitely benefit from some tests. I think it'll be a bit complex though, we'll need to:
- set up a couple of mock Grafana servers on different ports
- start the MCP server using our customised `run_sse_async` method
- start one SSE session with headers pointing to the first server, then send a message to use tool A
- start a second SSE session pointing to the second server, then use a different tool
- assert that the right servers get the right calls

I would really love to have some kind of load test as well to make sure there's nothing funky going on with the settings being accidentally shared between SSE sessions...

src/mcp_grafana/cli.py
```diff
@@ -13,5 +14,59 @@ class Transport(enum.StrEnum):
...
+    )
+
+    config = uvicorn.Config(
+        starlette_app,
+        host=self.settings.host,
+        port=self.settings.port,
+        log_level=self.settings.log_level.lower(),
+    )
+    server = uvicorn.Server(config)
+    await server.serve()
```
sd2k: This should be moved elsewhere really to make it easier to test...

src/mcp_grafana/middleware.py
```diff
@@ -0,0 +1,57 @@
...
+from .settings import GrafanaSettings, grafana_settings
+
+
+@dataclass
+class GrafanaInfo:
+    """
+    Simple container for the Grafana URL and API key.
+    """
+
+    authorization: str
```
annanay25: should we just name this API key?

src/mcp_grafana/middleware.py
```diff
@@ -0,0 +1,57 @@
...
+                api_key=info.authorization,
+                tools=current_settings.tools,
+            )
+            self.settings_token = grafana_settings.set(new_settings)
+            self.client_token = grafana_client.set(
+                GrafanaClient.from_settings(new_settings)
+            )
+
+    async def __aexit__(self, exc_type, exc_val, exc_tb):
+        if self.token is not None:
```
annanay25: I don't see the `self.token` being set anywhere? Just `self.settings_token` and `self.client_token` in the 'enter' section

annanay25 (commented):
Nice, clever design of handling Grafana auth headers using ContextVars and a special middleware. +1 on the load tests, it would be nice knowing tokens aren't leaking between requests.

src/mcp_grafana/middleware.py
sd2k: Whoops good call, that was missed in a refactor when I split the tokens up!

src/mcp_grafana/middleware.py
sd2k: Yeah I think so :+1:

sd2k:
Alright, I've added a test which starts up some separate Grafana servers and makes requests with different headers to hit each one. It seems to work fine!

One thing I noticed is that after the SSE request is set up, _anyone_ can send a JSONRPC request to `/messages?<uuid>` if they can guess the UUID, without any auth, and it will cause that message to be handled by the SSE handler. All output goes to the SSE handler rather than the client calling `/messages` though so I don't think this is a big deal, especially given how it's nearly impossible to guess that UUID.

src/mcp_grafana/tools/datasources.py
```diff
@@ -7,21 +7,21 @@ async def list_datasources() -> bytes:
     """
     List datasources in the Grafana instance.
     """
-    return await grafana_client.list_datasources()
+    return await grafana_client.get().list_datasources()
```
sd2k, line 10: I would love to do away with all of these annoying `.get()` calls, tools definitely shouldn't be able to modify this contextvar...

tests/middleware_test.py
```diff
@@ -0,0 +1,222 @@
...
+    """
+
+    @pytest.mark.asyncio
+    async def test_multiple_requests(self):
+        """
+        Ensure that the contextvars do not leak across requests.
+
+        This is a bit of a tricky test, since we're not actually
+        testing the middleware itself, but rather the contextvars
+        that it uses.
```
annanay25: Found this a bit confusing since we can't peek into the contextvars (IIUC) and we're using middlewares as a proxy for the test.

tests/middleware_test.py
```diff
@@ -0,0 +1,222 @@
...
+                        },
+                    )
+                    result = await jsonrpc_result(g1_iter)
+                    # This must have come from the first Grafana server.
+                    assert json.loads(result.content[0].text) == json.dumps(  # type: ignore
+                        [{"id": 1}], indent=4
+                    )
+
+                    # Send a tool call request using the second URL.
+                    await g2_client.post(
```
annanay25, line 176: Wondering if running the two requests in parallel will make the test a bit more convincing?

annanay25 (commented):
Also, cloud tests seem to have failed with 

```
=========================== short test summary info ============================
FAILED tests/tools/incident_test.py::test_list_incidents - mcp_grafana.client.GrafanaError: {"code":"Loading","message":"Your instance is loading, and will be ready shortly."}
=================== 1 failed, 11 passed, 15 skipped in 3.38s ===================
```

🫠

tests/middleware_test.py
sd2k: Ugh that comment was added by some AI tool and isn't even true. I'll remove it!

tests/middleware_test.py
sd2k: It'd be hard to ensure they both ran in parallel anyway really :shrug: we could try but it wouldn't be deterministic. I could add a second call to the first server if that would improve things at all?

annanay25: Nah honestly I don't think its worth the effort. The current test seems good 👍

sd2k:
> Also, cloud tests seem to have failed with

Yeah this is really annoying, I can't figure out how to keep that cloud instance online rather than pausing all the time... I'll dig into it a bit more!

annanay25 (approved):
lgtm!

sd2k:
There was a good suggestion by @breitreiter that we could use a custom transport for this. In reality we're just piggy-backing on the SSE transport but it's an interesting idea to make it an explicitly different transport, not just a hack on top of SSE :sweat_smile:

sd2k:
Marking as draft while we rethink some of the details.

annanay25:
~I'd hedge towards merging this (we can always rerun tests if the cloud instance is down) and handling that in a different PR, but will leave it to you!~

Nevermind, you marked draft because of the alternate transport consideration.

PR #22, opened 2025-02-10
Update README.md

InTheCloudDan:
Fix docker instructions to move the port forwarding to before the container name and switch to short syntax. I tried with `--port-forward` locally and received `--port-forward`. I did test the updated version locally.


sd2k (approved):

sd2k:
Nice thanks!

PR #23, opened 2025-02-11
Update README.md: Use `--with-editable` from uvx for development

gitdoluquita:
I wasn't being able to see my changes being reflected when using the `--from .` command.
Using `--with-editable` you get this "Live Reload" experience.


sd2k (approved):
good spot! :+1:

PR #24, opened 2025-02-13
feat: add PoC of stateless HTTP transport

sd2k:
This is an implementation of a stateless transport over HTTP for MCP. It's not part of the official spec, but something similar will probably be added in the future.

Functionally, it just presents a single HTTP endpoint (conventionally at `/mcp`) that accepts JSONRPC messages as POST bodies, and returns JSONRPC messages as responses.

Each inbound message is validated and gets its own 'server' instance. The handler then takes care of initializing the server (which in a stateless protocol is the responsibility of the client), then forwards the inbound message to the server, and returns the server's response.

We could easily add middleware for things like authentication, similar to in #21.

This also comes with a 'client' implementation which is similar to the stdio and sse clients in the official SDK and returns a pair of read/write streams for use with a `ClientSession`. This is useful for testing or writing example MCP clients.

The idea is that this transport will be compatible with the [HTTP transport in the unofficial Go SDK][go-sdk], so we can start building examples of usage. Notably this client also supports custom headers, which will be useful for authentication.

You can test this out locally by running it like this:

```bash
uvx --with-editable . mcp-grafana --transport http
```

Then calling it like this:

```bash
curl 'localhost:8000/mcp' -d '{"jsonrpc": "2.0", "id": 1, "method": "tools/list"}'
```

[go-sdk]: https://github.com/metoro-io/mcp-golang/blob/main/examples/http_example/README.md

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -34,3 +34,7 @@ build-backend = "hatchling.build"
 
 [tool.pytest.ini_options]
 asyncio_default_fixture_loop_scope = "session"
+
+[tool.ruff]
+line-length = 88
+target-version = "py310"
--- a/src/mcp_grafana/transports/http.py
+++ b/src/mcp_grafana/transports/http.py
@@ -179,10 +179,16 @@ async def http_client(url: str, headers: dict[str, Any] | None = None):
                 logger.debug("Waiting for request body")
                 body = await write_stream_reader.receive()
 
-                logger.debug(f"Connecting to HTTP endpoint: {url}")
+                u = httpx.URL(url)
+                u_cleaned = (
+                    u
+                    if u.userinfo == b""
+                    else u.copy_with(username="<redacted>", password="<redacted>")
+                )
+                logger.debug(f"Connecting to HTTP endpoint: {u_cleaned}")
                 async with httpx.AsyncClient(headers=headers) as client:
                     response = await client.post(
-                        url, content=body.model_dump_json(by_alias=True)
+                        u, content=body.model_dump_json(by_alias=True)
                     )
                     logger.debug(f"Received response: {response.status_code}")
                     message = types.JSONRPCMessage.model_validate_json(response.content)
--- a/src/mcp_grafana/transports/http.py
+++ b/src/mcp_grafana/transports/http.py
@@ -1,0 +1,182 @@
+import logging
+from contextlib import asynccontextmanager
+from json import JSONDecodeError
+from typing import Any, Tuple
+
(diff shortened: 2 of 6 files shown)


src/mcp_grafana/transports/http.py
```diff
@@ -0,0 +1,198 @@
...
+    read_stream, read_stream_writer, write_stream, write_stream_reader = make_streams()
+
+    async with anyio.create_task_group() as tg:
+        try:
+
+            async def http_rw():
+                logger.debug("Waiting for request body")
+                body = await write_stream_reader.receive()
+
+                logger.debug(f"Connecting to HTTP endpoint: {url}")
```
annanay25: nit: This has bitten me before when the URLs contain usernames/passwords 😬  but its behind debug so might be ok!

pyproject.toml
```diff
@@ -34,3 +34,7 @@ build-backend = "hatchling.build"
 
 [tool.pytest.ini_options]
 asyncio_default_fixture_loop_scope = "session"
+
```
annanay25, line 37: (discussed offline) Need a python version bump at the top of this file.

annanay25 (commented):
> It's not part of the official spec, but something similar will probably be added in the future.

Will it be a pain to maintain the unofficial client if this doesn't get merged into the official spec? I remember reading somewhere that we'd have to maintain our own implementation anyway, but just confirming we are ok with that outcome.

Everything else LGTM!

pyproject.toml
sd2k: Yeah I think we can do that in a separate PR anyway, it's unrelated to this PR :slightly_smiling_face:

sd2k: Oh, we already bumped requires-python to 3.11! https://github.com/grafana/mcp-grafana/pull/24/commits/c78e98abf54dcbeefb60a4a33e8aa0d60e114c61 should fix this error for 3.11, whoops.

src/mcp_grafana/transports/http.py
csmarchbanks: I wouldn't say that's ok, we commonly enable debug logging in prod as needed. We should make sure to clean the url before writing it.

sd2k: Good catch, fixed.

annanay25 (approved):
lgtm!

sd2k:
We might take inspiration from this in future, but with #29 the Python codebase is no more so this can be closed.

PR #27, opened 2025-02-22
Limit the data returned in list operations

csmarchbanks:
When running list operations against large instances even the default limits were causing things like Cursor or Claude Desktop to bail as the conversation was too large. Using summary types for these calls instead allow them to return with reasonable limits.

### DIFF
--- a/src/mcp_grafana/tools/datasources.py
+++ b/src/mcp_grafana/tools/datasources.py
@@ -1,13 +1,30 @@
+from typing import List
 from mcp.server import FastMCP
+
+from mcp_grafana.grafana_types import Datasource
 
 from ..client import grafana_client
 
 
-async def list_datasources() -> bytes:
+async def list_datasources() -> List[Datasource]:
     """
     List datasources in the Grafana instance.
     """
-    return await grafana_client.list_datasources()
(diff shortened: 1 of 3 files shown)


src/mcp_grafana/tools/datasources.py
```diff
@@ -1,13 +1,30 @@
...
+            {
+                "id": ds.id,
+                "uid": ds.uid,
+                "name": ds.name,
+                "type": ds.type,
+                "isDefault": ds.is_default,
+            }
+        )
+
+    return resp
```
sd2k, line 27: I think this type annotation is incorrect since we're returning dicts not datasource objects? Probably doesn't matter for now but mypy will probably complain if/when we add it.

Perhaps we could add a more minimal `Datasource` type instead though, since the model might be expecting to see all the fields if it's passed the schema of the `Datasource` type.

csmarchbanks: Yep, that is on the list of things to fix before I mark this as ready to review.

csmarchbanks:
Supserseded by https://github.com/grafana/mcp-grafana/pull/30.

PR #29, opened 2025-03-10
feat: rewrite MCP server in Go

sd2k:
This PR removes the existing Python implementation of the MCP server and adds a Go version.

The server can be started in either stdio or SSE mode based on the arguments. We still need to add some more flags for e.g. host and port of SSE mode.

Tools are factored into a separate package which can be imported from elsewhere. Using a similar method to the `mcp-golang` SDK, we use reflection to inspect the types of the functions and generate tool definitions, used for the `list_tools` functionality.

TODO:

- [x] add flags for host/port when running SSE mode
- [x] update README
- [ ] add/update a CONTRIBUTING doc which describes how to add new tools
- [x] re-add some integration tests (using docker-compose) and possibly cloud tests (using the mcptests Grafana cloud dev instance)
- [x] update CI to stop running Python tests
- [x] document why we're using a fork of mcp-go
- [x] rename client.go to something more appropriate

Maybe later?

- [ ] add a way to disable certain categories of tools


tools/datasources.go
```diff
@@ -0,0 +1,85 @@
...
+		return nil, fmt.Errorf("list datasources: %w", err)
+	}
+	b, err := json.Marshal(datasources.Payload)
+	if err != nil {
+		return nil, fmt.Errorf("marshal datasources: %w", err)
+	}
+	return mcp.NewToolResultText(string(b)), nil
+}
+
+var ListDatasourcesTool, ListDatasourcesHandler = mcpgrafana.MustTool(
```
csmarchbanks: Do we want to have `ListDatasourcesTool` and `ListDatasourcesHandler` exposed? I would be tempted just to run the `mcpgrafana.MustTool` command inline when adding tools.

sd2k: Mmm that would be nice. The only downside is that the description is part of the `mcpgrafana.MustTool` arguments so we'd need to reproduce that anywhere we inline the `MustTool` call... that's annoying. It would be nice if we could just export a single object with `Tool` and `Handler` properties instead, perhaps that's worth doing?

sd2k: I pushed a [commit](https://github.com/grafana/mcp-grafana/pull/29/commits/b8a111c9904ed41ead067ac07948957518ae5e6c) which I think improves these APIs by defining our own `Tool` that has a `Register` method so this can be done in a single expression rather than requiring a bunch of temporaries. I also improved the type safety using generics instead of `any`, which meant I could remove a bunch of unneeded tests :partying_face:

sd2k:
To make the migration easier for users I was hoping we could continue with our Python package and shove a platform-specific Go binary inside the package wheel, and use a tiny `main.py` which forwards to the Go executable. I think we can do that in a future PR though.

tools/datasources.go
csmarchbanks: Nice, yeah I like that more.

csmarchbanks:
I think that can be a future PR, and I am not sure if it is really necessary either. If we can have easy instructions for installing the go binary then having a Python wrapper would just make things more complicated.

csmarchbanks (approved):
I fixed one small issue with incident not working, but otherwise looks good. If my most recent commit looks ok let's get this in!

PR #30, opened 2025-03-12
Limit the amount of data returned for list tools

csmarchbanks:
List incidents and list datasources can use a lot more context than necessary when returning results. For dashboards use a custom struct that only returns a few fields. For incidents use the preview incident endpoint which returns a more concise format.


sd2k:
LGTM, just removed an unused type which I guess was used before you found the `incidentSummary` type!

PR #35, opened 2025-03-12
Add datasources including dev

ioanarm:


tests/provisioning/datasources/datasources.yaml
```diff
@@ -8,3 +8,15 @@ datasources:
     access: proxy
     url: http://prometheus:9090
     isDefault: true
+  - name: DigitalOcean demo
+    id: 2
+    uid: digitalocean
+    type: prometheus
+    access: proxy
+    url: http://demo.do.prometheus.io:9090
```
csmarchbanks: This doesn't actually exist anymore and the Prometheus team isn't planning on recreating the digital ocean one. There is now a demo server at https://prometheus.demo.prometheus.io but it isn't digitalocean anymore, maybe just call it `Prometheus Demo`?

ioanarm: Removed it and also removed Prom labs and replaced with the above. Thanks

csmarchbanks (approved):

csmarchbanks (commented):
Looks like a test is only expecting one datasource. Maybe just have it look for the existence of the first datasource but not care that there is exactly one?

PR #39, opened 2025-03-13
fix: update Go installation instructions in README

deagon:
Fix the wrong step for installing go package.


sd2k (approved):
Thanks! :pray:

Issue #33, opened 2025-03-12
Add tools to query Loki


sd2k:
Useful tools could be:

- [x] run datasource query
  - should we distinguish between 'run a metric query' and 'run a log query'? the model still needs to write a LogQL
- [x] get label names
- [x] get label values
- [x] get stream stats (this lets the model get an idea of how big a query will be)

sd2k:
Done in #44.

PR #44, opened 2025-03-14
Add loki tools

ioanarm:
Related to [this issue](https://github.com/grafana/mcp-grafana/issues/33)

In this PR we are adding tools to:
- get label names
- get label values
- get stats
- run queries (either metric or log queries)
from Loki datasources. 

TODO: 
- [x] Update README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,7 +47,7 @@ This is useful if you don't use certain functionality or if you don't want to ta
 | `create_incident` | Incident | Create an incident in Grafana Incident |
 | `add_activity_to_incident` | Incident | Add an activity item to an incident in Grafana Incident |
 | `resolve_incident` | Incident | Resolve an incident in Grafana Incident |
-| `query_loki_logs` | Loki | Query and retrieve log entries using LogQL |
+| `query_loki_logs` | Loki | Query and retrieve logs using LogQL (either log or metric queries) |
 | `list_loki_label_names` | Loki | List all available label names in logs |
 | `list_loki_label_values` | Loki | List values for a specific log label |
 | `query_loki_stats` | Loki | Get statistics about log streams |
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ This provides access to your Grafana instance and the surrounding ecosystem.
   - [x] Prometheus
   - [x] Loki
     - [x] Log queries
-    - [ ] Metric queries
+    - [x] Metric queries
   - [ ] Tempo
   - [ ] Pyroscope
 - [x] Query Prometheus metadata
--- a/README.md
+++ b/README.md
@@ -10,7 +10,9 @@ This provides access to your Grafana instance and the surrounding ecosystem.
 - [x] List and fetch datasource information
 - [ ] Query datasources
   - [x] Prometheus
-  - [ ] Loki (log queries, metric queries)
+  - [x] Loki
+    - [x] Log queries
+    - [ ] Metric queries
   - [ ] Tempo
   - [ ] Pyroscope
 - [x] Query Prometheus metadata
@@ -18,6 +20,11 @@ This provides access to your Grafana instance and the surrounding ecosystem.
   - [x] Metric names
   - [x] Label names
   - [x] Label values
+- [x] Query Loki
+  - [x] Label names
+  - [x] Label values
+  - [x] Stats
+  - [x] Logs
 - [x] Search, create, update and close incidents
 - [ ] Start Sift investigations and view the results
 
@@ -41,6 +48,10 @@ This is useful if you don't use certain functionality or if you don't want to ta
 | `create_incident` | Incident | Create an incident in Grafana Incident |
 | `add_activity_to_incident` | Incident | Add an activity item to an incident in Grafana Incident |
 | `resolve_incident` | Incident | Resolve an incident in Grafana Incident |
+| `query_loki_logs` | Loki | Query and retrieve log entries using LogQL |
+| `list_loki_label_names` | Loki | List all available label names in logs |
+| `list_loki_label_values` | Loki | List values for a specific log label |
+| `query_loki_stats` | Loki | Get statistics about log streams |
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -20,11 +20,10 @@ This provides access to your Grafana instance and the surrounding ecosystem.
   - [x] Metric names
   - [x] Label names
   - [x] Label values
-- [x] Query Loki
+- [x] Query Loki metadata
   - [x] Label names
   - [x] Label values
   - [x] Stats
-  - [x] Logs
 - [x] Search, create, update and close incidents
 - [ ] Start Sift investigations and view the results
 
--- a/tools/loki.go
+++ b/tools/loki.go
@@ -15,6 +15,14 @@ import (
 	"github.com/mark3labs/mcp-go/server"
 )
 
+const (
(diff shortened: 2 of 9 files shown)


tools/loki.go
```diff
@@ -0,0 +1,484 @@
...
+		return nil, fmt.Errorf("Loki API returned unexpected response format: %s", string(bodyBytes))
+	}
+
+	return queryResponse.Data.Result, nil
+}
+
+// QueryLokiLogsParams defines the parameters for querying Loki logs
+type QueryLokiLogsParams struct {
+	DatasourceUID string `json:"datasourceUid" jsonschema:"required,description=The UID of the datasource to query"`
+	LogQL         string `json:"logql" jsonschema:"required,description=The LogQL query to execute"`
```
csmarchbanks: How well do you find the various LLMs at being able to generate LogQL? I ran into issues with valid LogQL at one point and went down a slightly different path where I instead prompted for label matchers and a string to match against, then constructed valid LogQL from that.

ioanarm: It may try to get stats for example using a complex query. When the tools is giving an error though then it will re try with a simpler approach. 

image

Maybe we can have a second iteration here to give a bit more context as @sd2k [suggested](https://raintank-corp.slack.com/archives/C087XT6A8P5/p1742221533596509).

sd2k: Yeah it might be nice to have a few ways to provide a bit more context. I can think of a couple of ways:

- a tool which just provides a LogQL primer, so the model can use it before it needs to write LogQL. We could include some brief instructions in the query logs tool to say 'make sure to call the LogQL primer tool first'. That way the context isn't automatically expanded just by having the tool included, only when the model wants to use it.
- a resource which does the same thing, but that the user or client can choose to include (e.g. by inputting `@logql-primer` or something - however the client chooses to handle that)

tools/loki.go
```diff
@@ -0,0 +1,484 @@
...
+		return nil, fmt.Errorf("unmarshalling response (content: %s): %w", string(bodyBytes), err)
+	}
+
+	return &stats, nil
+}
+
+// QueryLokiStatsParams defines the parameters for querying Loki stats
+type QueryLokiStatsParams struct {
+	DatasourceUID string `json:"datasourceUid" jsonschema:"required,description=The UID of the datasource to query"`
+	LogQL         string `json:"logql" jsonschema:"required,description=The LogQL query to execute"`
```
sd2k: It might be worth noting in the description that this can't be arbitrary LogQL, but only matchers (see https://grafana.com/docs/loki/latest/reference/loki-http-api/#query-log-statistics).

tools/loki.go
```diff
@@ -0,0 +1,484 @@
...
+		return nil, err
+	}
+
+	return stats, nil
+}
+
+// QueryLokiStats is a tool for querying stats from Loki
+var QueryLokiStats = mcpgrafana.MustTool(
+	"query_loki_stats",
+	"Query statistics about logs in a Loki datasource using LogQL",
```
sd2k:
```suggestion
	"Query statistics about log streams in a Loki datasource, using LogQL selectors to select streams",
```

sd2k (commented):
Code looks nice! A couple of suggestions, we might need to play around with the descriptions to they get decent results but this is a great start, thanks.

sd2k:
[These Anthropic docs](https://docs.anthropic.com/en/docs/build-with-claude/tool-use/overview#best-practices-for-tool-definitions) about good tool descriptions might be helpful! (we could do with improving lots of others, too)

tools/loki.go
ioanarm: Updated the descriptions a bit. Trying to test this but again ml instance is down :(

ioanarm:
> [These Anthropic docs](https://docs.anthropic.com/en/docs/build-with-claude/tool-use/overview#best-practices-for-tool-definitions) about good tool descriptions might be helpful! (we could do with improving lots of others, too)

That's really useful @sd2k thanks!

tools/loki.go
ioanarm: Keen to try this but I would probably do it on a separate PR. After changing the descriptions the stats endpoint is called with a matcher so it seems improved.

tools/loki.go
```diff
@@ -0,0 +1,484 @@
...
+		return []LogEntry{}, nil
+	}
+
+	return entries, nil
+}
+
+// QueryLokiLogs is a tool for querying logs from Loki
+var QueryLokiLogs = mcpgrafana.MustTool(
+	"query_loki_logs",
+	"Query and retrieve log entries from a Loki datasource using LogQL. Returns log lines with their timestamps and associated label metadata. Supports full LogQL syntax including filters and expressions.",
```
sd2k: It might be worth adding something like "Before using this tool, use `query_loki_stats` to ensure the log stream is not too large, `list_loki_label_names` to ensure the labels exist, and `list_loki_label_values` to ensure the label values exist.

sd2k (commented):
Could you update the README too? I think that's all that's left!

ioanarm:
> Could you update the README too? I think that's all that's left!

Done. I added one more commit which supports metric queries with Loki. 
You can ask the model to give a graph as well which it did but maybe just getting some basic stats is enough

image.

wdyt?

README.md
```diff
@@ -10,14 +10,21 @@ This provides access to your Grafana instance and the surrounding ecosystem.
...
 - [x] Query Prometheus metadata
   - [x] Metric metadata
   - [x] Metric names
   - [x] Label names
   - [x] Label values
+- [x] Query Loki
+  - [x] Label names
+  - [x] Label values
+  - [x] Stats
+  - [x] Logs
```
```suggestion
- [x] Query Loki metadata
  - [x] Label names
  - [x] Label values
  - [x] Stats
```

tools/loki.go
```diff
@@ -0,0 +1,518 @@
...
+		return []LogEntry{}, nil
+	}
+
+	return entries, nil
+}
+
+// QueryLokiLogs is a tool for querying logs from Loki
+var QueryLokiLogs = mcpgrafana.MustTool(
+	"query_loki_logs",
+	"Query and retrieve log entries or metric values from a Loki datasource using LogQL. Returns either log lines or numeric values with timestamps and labels. Use query_loki_stats first to check stream size, then list_loki_label_names/values to verify labels exist. Supports full LogQL syntax including both log queries and metric queries (e.g., rate, count_over_time).",
```
sd2k:
```suggestion
	"Query and retrieve log entries or metric values from a Loki datasource using LogQL. Returns either log lines or numeric values with timestamps and labels. Use `query_loki_stats` first to check stream size, then `list_loki_label_names` and `list_loki_label_values` to verify labels exist. Supports full LogQL syntax including both log queries and metric queries (e.g., rate, count_over_time).",
```

README.md
```diff
@@ -41,6 +48,10 @@ This is useful if you don't use certain functionality or if you don't want to ta
 | `create_incident` | Incident | Create an incident in Grafana Incident |
 | `add_activity_to_incident` | Incident | Add an activity item to an incident in Grafana Incident |
 | `resolve_incident` | Incident | Resolve an incident in Grafana Incident |
+| `query_loki_logs` | Loki | Query and retrieve log entries using LogQL |
```
```suggestion
| `query_loki_logs` | Loki | Query and retrieve logs using LogQL (either log or metric queries) |
```

sd2k (approved):
A few minor nits but lgtm!

PR #50, opened 2025-03-20
Oncall Tools

ioanarm:
The PR adds support for Grafana OnCall features to the MCP server, including:

-  `ListOnCallSchedules`: Lists schedules from Grafana OnCall, with team ID filtering
-  `GetOnCallShift`: Gets detailed information for a specific OnCall shift by ID
-  `GetCurrentOnCallUsers`: Shows who's currently on-call for a specific schedule
-  `ListOnCallTeams`: Lists all teams from Grafana OnCall with their details
-  `ListOnCallUsers`: Lists all users from Grafana OnCall with their details or Get a specific user by the UserID or Username

Also this PR adds a "cloud test" against a specific testing instance (for now `mcptests.grafana-dev.net`)

Adding support for:
* alert groups
* shift swap requests 
depend on adding support for them to the [client](https://github.com/grafana/amixr-api-go-client).  
For alert groups there is already an open PR which we are pushing forward at the moment. 

**Before merging we need to alter the github actions required to unblock merging**
At the moment we require `Test` while we are gonna need three rules specific for unit, integration and cloud.


Makefile
```diff
@@ -24,6 +24,10 @@ test: ## Run the Go unit tests.
 test-all: ## Run the Go unit and integration tests.
 	go test -v -tags integration ./...
 
+.PHONY: test-cloud
+test-cloud: ## Run only the cloud tests.
+	go test -v -tags cloud ./tools -run TestCloudOnCall
```
sd2k:
```suggestion
	go test -v -tags cloud ./...
```

tools/oncall.go
```diff
@@ -0,0 +1,231 @@
...
+	defer resp.Body.Close()
+
+	if resp.StatusCode != http.StatusOK {
+		return "", fmt.Errorf("unexpected status code from settings API: %d", resp.StatusCode)
+	}
+
+	var settings struct {
+		JsonData struct {
+			OnCallApiUrl string `json:"onCallApiUrl"`
+		} `json:"jsonData"`
```
sd2k: nit: the Go field names would conventionally be `JSONData` and `OnCallAPIURL` (the `json` tags are correct though)

tools/oncall.go
```diff
@@ -0,0 +1,231 @@
...
+func listOnCallSchedules(ctx context.Context, args ListOnCallSchedulesParams) ([]*aapi.Schedule, error) {
+	client, err := oncallClientFromContext(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("getting OnCall client: %w", err)
+	}
+
+	listOptions := &aapi.ListScheduleOptions{}
+
+	scheduleService := aapi.NewScheduleService(client)
+	response, _, err := scheduleService.ListSchedules(listOptions)
```
sd2k, line 123: It looks like this API is paginated; we should provide a `page` option to the params and pass it through to this API call.

tools/oncall.go
```diff
@@ -0,0 +1,231 @@
...
+}
+
+func listOnCallTeams(ctx context.Context, args ListOnCallTeamsParams) ([]*aapi.Team, error) {
+	client, err := oncallClientFromContext(ctx)
+	if err != nil {
+		return nil, fmt.Errorf("getting OnCall client: %w", err)
+	}
+
+	teamService := aapi.NewTeamService(client)
+	response, _, err := teamService.ListTeams(&aapi.ListTeamOptions{})
```
sd2k: This is paginated too.

tools/oncall.go
```diff
@@ -0,0 +1,231 @@
...
+	if args.UserID != "" {
+		user, _, err := userService.GetUser(args.UserID, &aapi.GetUserOptions{})
+		if err != nil {
+			return nil, fmt.Errorf("getting OnCall user %s: %w", args.UserID, err)
+		}
+		return []*aapi.User{user}, nil
+	}
+
+	// Otherwise, list all users
+	response, _, err := userService.ListUsers(&aapi.ListUserOptions{})
```
sd2k: Also paginated :slightly_smiling_face:

There's also a `Username` field here, is it worth adding that to the params or is it unlikely that the LLM would be able to figure out the username?

tools/oncall.go
```diff
@@ -0,0 +1,231 @@
...
+	"get_oncall_shift",
+	"Get details for a specific OnCall shift",
+	getOnCallShift,
+)
+
+type GetCurrentOnCallUsersParams struct {
+	ScheduleID string `json:"scheduleId" jsonschema:"required,description=The ID of the schedule to get current on-call users for"`
+}
+
+func getCurrentOnCallUsers(ctx context.Context, args GetCurrentOnCallUsersParams) (*aapi.Schedule, error) {
```
sd2k: This tool/function is named & described as if it's fetching _users_ but it's actually returning a schedule. Should we rename the tool to indicate that it gets the on call schedule? Or change the output to just the users who are currently on call (I guess the `OnCallNow` field)?

Also, how big is a `Schedule`'s JSON representation? if it's big we might need to consider returning just a summary.

tools/oncall.go
```diff
@@ -0,0 +1,231 @@
...
+	"list_oncall_schedules",
+	"List schedules from Grafana OnCall",
+	listOnCallSchedules,
+)
+
+type GetOnCallShiftParams struct {
+	ShiftID string `json:"shiftId" jsonschema:"required,description=The ID of the shift to get details for"`
+}
+
+func getOnCallShift(ctx context.Context, args GetOnCallShiftParams) (*aapi.OnCallShift, error) {
```
sd2k, line 161: `aapi.OnCallShift` looks like it might be quite big, should we just return a summary? Or is it all useful?

tools/oncall.go
```diff
@@ -0,0 +1,231 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("getting OnCall shift %s: %w", args.ShiftID, err)
+	}
+
+	return shift, nil
+}
+
+var GetOnCallShift = mcpgrafana.MustTool(
+	"get_oncall_shift",
+	"Get details for a specific OnCall shift",
```
sd2k: It might be useful to describe what a 'shift' is (and what a 'schedule' is in the other tools) - I'm not entirely sure as a human who is a member of both :grin:

sd2k (commented):
Nice, this looks like it will be super useful! I added a few comments inline.

tools/oncall.go
ioanarm: One Shift can give you info like: 
* Type of the shift, 
* Start
* Duration
* Frequency
* Weekdays
* Users that participate in it
Depends on what the user is asking can be quite useful imo. For one shift ID I think that the json response is small enough.

tools/oncall.go
ioanarm: 👍 I've modified it slightly so it returns the info asked. 
The schedule JSON representation is not that big again but I can probably remove the slack and calendar related for now.

Makefile
ioanarm: 👍 I need the path though `./tools` as we have the `mcpgrafana_test` which checks the env vars.

ioanarm:
@sd2k Thanks for reviewing this :) 
I think I covered your comments, let me know if there is anything else I can add/improve!

Makefile
ioanarm: I've separated completely unit tests, docker based integration ones and cloud so we now run three separate commands in git actions.
docker and cloud ones require env vars which are different in each case so having them both under the same command just complicated things for no reason. 
three test categories with tags maybe make a bit more sense.

ioanarm: if it works 🙄

Makefile
```diff
@@ -16,14 +16,18 @@ build-image: ## Build the Docker image.
 lint: ## Lint the Go code.
 	go tool -modfile go.tools.mod golangci-lint run
 
-.PHONY: test
-test: ## Run the Go unit tests.
-	go test ./...
+.PHONY: test-unit
+test-unit: ## Run the unit tests (no external dependencies required).
+	go test -v -tags unit ./...
```
sd2k:
```suggestion
.PHONY: test test-unit
test-unit: ## Run the unit tests (no external dependencies required).
	go test -v -tags unit ./...
test: test-unit
```

This is really minor but I already know I'm going to try running `make test` locally and it's going to fail and annoy me so this should fix it :sweat_smile:

sd2k (approved):
LGTM! one very minor suggestion. I'll update the Ruleset to refer to the new CI jobs too.

sd2k:
Ah the README will need updating to refer to the new `make` targets too.

PR #55, opened 2025-03-21
tests: start adding e2e tests

sd2k:
This PR adds end-to-end tests for Loki integration and adds test documentation.
This is iteration one as we want to add a basic structure on e2e testing for now. We need to iterate further on them.

_Note:_ prompts needs to be specific when using llm-as-a-judge. I've noticed some flakiness on the llm responses so some times tests are failing, especially the `test_loki_logs_tool`. 

When we are confident that tests are consistently passing then we can make it part of the ruleset.


tests/loki_test.py
```diff
@@ -0,0 +1,197 @@
...
+            },
+        )
+    )
+
+    # Call the LLM including the tool call result.
+    response = await acompletion(
+        model=model,
+        messages=messages,
+        tools=tools,
+    )
```
sd2k: This shouldn't be in the final test really, I just wanted to see how well stuff worked. If the model gets there without needing to call these exact tools then that's fine.

tests/pyproject.toml
```diff
@@ -0,0 +1,23 @@
...
+    { name = "Ben Sully", email = "ben.sully@grafana.com" },
+    { name = "Ioanna Armouti", email = "ioanna.armouti@grafana.com" },
+    { name = "Chris Marchbanks", email = "chris.marchbanks@grafana.com" },
+]
+requires-python = ">=3.13"
+dependencies = []
+
+[dependency-groups]
+dev = [
+    "anyio>=4.9.0",
```
sd2k, line 16: I guess we should include dotenv here since we use it explicitly.

sd2k (commented):
Nice, thanks for getting this working properly! I can't approve since I'm the OG author so we'll need someone else to take a look too.

tests/loki_test.py
```diff
@@ -0,0 +1,254 @@
...
+    ):
+        async with ClientSession(read, write) as session:
+            await session.initialize()
+            yield session
+
+
+@pytest.mark.parametrize("model", models)
+async def test_loki_logs_tool(model: str, mcp_client: ClientSession):
+    tools = await mcp_client.list_tools()
+    prompt = "Can you list the last 10 log lines from all containers using any available Loki datasource? Give me the raw log lines. Please use only the necessary tools to get this information."
```
csmarchbanks, line 58: This test is failing for me at least half the time. Generally from trying to put in some non-container label matcher, anything from `{job=~".+"}` to `{job="varlog"}`. I wonder if we could at least tweak the prompt or the tool description to get the test to work more consistently.

ioanarm: I've added a pass rate and reruns in case of failure. Pass rate alone wasn't triggering reruns so another decorator was needed. 
This doesn't solve the issue longterm though :(. We need to iterate on the tests and make sure that they are passing consistently. 
This should be fine as a first go but happy to iterate on the prompts in this PR as well.

ioanarm: Try it again and let me know!

tests/loki_test.py
```diff
@@ -0,0 +1,258 @@
...
+        write,
+    ):
+        async with ClientSession(read, write) as session:
+            await session.initialize()
+            yield session
+
+
+@pytest.mark.parametrize("model", models)
+@pytest.mark.flaky(reruns=3)
+@pytest.mark.pass_rate(0.6)
```
csmarchbanks: Flaky combined with pass_rate isn't working great for me. That said, just flaky works well for now. I think to have pass_rate work we might need more than just two models to run across? Or at least I was still get failures when one model failed and therefore 50% < 60%.

ioanarm: Yeah I agree. we can take out pass_rate for now, add the tests and iterate separately!

tests/loki_test.py
```diff
@@ -0,0 +1,256 @@
...
+        read,
+        write,
+    ):
+        async with ClientSession(read, write) as session:
+            await session.initialize()
+            yield session
+
+
+@pytest.mark.parametrize("model", models)
+@pytest.mark.flaky(reruns=3)
```
sd2k: need to add `flaky` manually and rename to `max_runs`

PR #57, opened 2025-03-24
fix: use `grafana-irm-app` as plugin ID for incident tools

### LINKED ISSUE #49: Use `grafana-irm-app` in URL for Incident tools
The Grafana Incident app has been merged with OnCall to create the IRM app and the old plugin will go away eventually. We should make sure we're using the new path instead.

sd2k:
The Incident plugin has been deprecated in favour of the IRM
plugin, so we should use the new plugin ID when constructing
the URL for the Incident client.

Fixes #49.


gerboland (approved):

sd2k:
Marking as draft until the IRM app is fully rolled out.

brojd:
@sd2k we are good to go now

PR #58, opened 2025-03-24
Alerting: Add support for getting alert rule information

alexander-akhmetov:
This PR adds new tools to interact with Grafana Alerting:

- `list_alert_rules`: List all alert rules
- `get_alert_rule_by_uid`: Get information about a specific alert rule

### DIFF
--- a/tools/alerting.go
+++ b/tools/alerting.go
@@ -3,11 +3,12 @@ package tools
 import (
 	"context"
 	"fmt"
-	"regexp"
 
 	"github.com/grafana/grafana-openapi-client-go/models"
-	mcpgrafana "github.com/grafana/mcp-grafana"
 	"github.com/mark3labs/mcp-go/server"
+	"github.com/prometheus/prometheus/model/labels"
+
+	mcpgrafana "github.com/grafana/mcp-grafana"
 )
 
 const (
@@ -48,23 +49,9 @@ func listAlertRules(ctx context.Context, args ListAlertRulesParams) ([]alertRule
 		return nil, fmt.Errorf("list alert rules: %w", err)
 	}
 
-	// Apply selectors if any
-	alertRules := response.Payload
-	if len(args.LabelSelectors) > 0 {
-		filteredResult := models.ProvisionedAlertRules{}
-		for _, rule := range alertRules {
-			if rule == nil {
-				continue
-			}
-			match, err := matchesSelectors(*rule, args.LabelSelectors)
-			if err != nil {
-				return nil, fmt.Errorf("list alert rules: %w", err)
-			}
-			if match {
-				filteredResult = append(filteredResult, rule)
-			}
-		}
-		alertRules = filteredResult
+	alertRules, err := filterAlertRules(response.Payload, args.LabelSelectors)
+	if err != nil {
+		return nil, fmt.Errorf("list alert rules: %w", err)
 	}
 
 	alertRules, err = applyPagination(alertRules, args.Limit, args.Page)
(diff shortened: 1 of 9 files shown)


tools/alerting.go
```diff
@@ -0,0 +1,51 @@
...
+		return nil, fmt.Errorf("list alert rules: %w", err)
+	}
+	return alertRules.Payload, nil
+}
+
+var ListAlertRules = mcpgrafana.MustTool(
+	"list_alert_rules",
+	"List alert rules",
+	listAlertRules,
+)
```
sd2k: We should probably add a `limit` param to this at least, and ideally some more things to enable pagination and/or filtering (e.g. filtering based on labels or annotations).

We can do the filtering in the tool if the Grafana client doesn't support it. It would be nice to support Prometheus-style selectors for label matchers - you could probably reuse the `Selector` struct from `prometheus.go` and just run the selector matching logic manually.

Otherwise we risk immediately exploding the context window of the LLM since this endpoint can return a ton of data :see_no_evil:

sd2k: It might also be worth defining a separate `alertRuleSummary` struct which just includes a subset of the fields (`uid` and `title`, maybe) and returning a slice of those from the `list_alert_rules` tool, then the LLM can use `get_alert_rule_by_uid` if it wants to view all the details. See `datasources.go` for an example of where we do this for datasources.

sd2k (commented):
Nice, thanks for the contribution! I have a couple of suggestions related to limiting the output of the tools to avoid creating too many tokens.

tools/alerting.go
alexander-akhmetov: Thanks for the review, great points :) This alerting endpoint doesn't support pagination, I'll add it in the tool.

alexander-akhmetov: I added `Labels` also to the `alertRuleSummary` just because of the new label filtering, but not sure if it should be there, maybe it's better to remove it.

sd2k (approved):
Nice, this looks great!

One comment would be that we may be able to reuse the [`Matcher`](https://pkg.go.dev/github.com/prometheus/prometheus/model/labels#Matcher) struct from the Prometheus codebase, but this has tests and works well so it's not a biggie. I'll leave this up for a little longer in case anyone else from the team wants to review but otherwise will merge soon.

Thanks again! :partying_face:

alexander-akhmetov:
Thanks for the review! I wasn't sure about pulling in Prometheus but it would definitely make things simpler. I'll take a look. If in the end you'll decide that it's good to go as is, I'll do a quick follow-up with the Prometheus matcher. Otherwise will update this one :)

sd2k:
I think we're pulling in Prometheus for the Prometheus tools anyway so dependency wise it's no different! :grin: I'll leave this up for you to update :+1:

tools/prometheus.go
```diff
@@ -195,9 +206,9 @@ var ListPrometheusMetricNames = mcpgrafana.MustTool(
 )
 
 type LabelMatcher struct {
-	Name  string `json:"name"`
-	Value string `json:"value"`
-	Type  string `json:"type"` // "=" or "!=" or "=~" or "!~"
+	Name  string `json:"name" jsonschema:"required,description=The name of the label to match against"`
+	Value string `json:"value" jsonschema:"required,description=The value to match against"`
+	Type  string `json:"type" jsonschema:"required,description=One of the '=' or '!=' or '=~' or '!~'"`
```
alexander-akhmetov, line 211: I added the descriptions here mostly for the `Type` field, without it LLM tried to use values like `exact`

alexander-akhmetov:
Updated to use `Matcher`, let me know your thoughts: 93bf87f

sd2k (approved):
LGTM :+1: thank you!

PR #62, opened 2025-03-26
Add dashboard tools

gemiit:


ioanarm (approved):
Hey @gemiit , thanks for adding this! Looks good to me 👍
One observation - dashboard JSON can be quite deeply nested and potentially large in size, which might be worth considering for any users of the tool.

PR #71, opened 2025-04-01
Add some basic logging of config on startup when using stdio

### LINKED ISSUE #70: [RFE] debug logging capability
Right now there is no debug, or any, logging capability in the MCP server. When it is queried, it outputs nothing. When it starts it doesn't even tell you what URL or API key it is going to use (although maybe logging the API key is a bad idea...)

It would be nice for debugging purposes to output some information.

sd2k:
Closes #70.

### DIFF
--- a/cmd/mcp-grafana/main.go
+++ b/cmd/mcp-grafana/main.go
@@ -4,7 +4,7 @@ import (
 	"context"
 	"flag"
 	"fmt"
-	"log"
+	"log/slog"
 	"os"
 
 	"github.com/mark3labs/mcp-go/server"
@@ -29,19 +29,21 @@ func newServer() *server.MCPServer {
 	return s
 }
 
-func run(transport, addr string) error {
+func run(transport, addr string, logLevel slog.Level) error {
+	slog.SetDefault(slog.New(slog.NewTextHandler(os.Stderr, &slog.HandlerOptions{Level: logLevel})))
 	s := newServer()
 
 	switch transport {
 	case "stdio":
 		srv := server.NewStdioServer(s)
 		srv.SetContextFunc(mcpgrafana.ComposedStdioContextFunc)
+		slog.Info("Starting Grafana MCP server using stdio transport")
 		return srv.Listen(context.Background(), os.Stdin, os.Stdout)
 	case "sse":
 		srv := server.NewSSEServer(s,
 			server.WithSSEContextFunc(mcpgrafana.ComposedSSEContextFunc),
 		)
-		log.Printf("SSE server listening on %s", addr)
+		slog.Info("Starting Grafana MCP server using SSE transport")
 		if err := srv.Start(addr); err != nil {
 			return fmt.Errorf("Server error: %v", err)
 		}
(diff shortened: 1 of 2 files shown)


cmd/mcp-grafana/main.go
```diff
@@ -29,19 +29,21 @@ func newServer() *server.MCPServer {
...
 		srv := server.NewStdioServer(s)
 		srv.SetContextFunc(mcpgrafana.ComposedStdioContextFunc)
+		slog.Info("Starting Grafana MCP server using stdio transport")
 		return srv.Listen(context.Background(), os.Stdin, os.Stdout)
 	case "sse":
 		srv := server.NewSSEServer(s,
 			server.WithSSEContextFunc(mcpgrafana.ComposedSSEContextFunc),
 		)
-		log.Printf("SSE server listening on %s", addr)
+		slog.Info("Starting Grafana MCP server using SSE transport")
```
csmarchbanks: It would be nice to still have the address here.

mcpgrafana.go
```diff
@@ -47,6 +48,7 @@ var ExtractGrafanaInfoFromEnv server.StdioContextFunc = func(ctx context.Context
 	if u == "" {
 		u = defaultGrafanaURL
 	}
+	slog.Info("Using Grafana configuration", "url", u, "api_key_set", apiKey != "")
```
csmarchbanks: URL can contain a password, if we want to log the URL we should parse it and use `u.Redacted()`.

mcpgrafana.go
```diff
@@ -111,11 +115,16 @@ var ExtractGrafanaClientFromEnv server.StdioContextFunc = func(ctx context.Conte
...
+		grafanaURL = defaultGrafanaURL
 	}
-	if apiKey := os.Getenv(grafanaAPIEnvVar); apiKey != "" {
+
+	apiKey := os.Getenv(grafanaAPIEnvVar)
+	if apiKey != "" {
 		cfg.APIKey = apiKey
 	}
 
+	slog.Debug("Creating Grafana client", "url", grafanaURL, "api_key_set", apiKey != "")
```
csmarchbanks: We already do the parsing above... should probably just save that variable and then use `Redacted()`.

mcpgrafana.go
```diff
@@ -161,7 +170,11 @@ type incidentClientKey struct{}
 
 var ExtractIncidentClientFromEnv server.StdioContextFunc = func(ctx context.Context) context.Context {
 	grafanaURL, apiKey := urlAndAPIKeyFromEnv()
+	if grafanaURL == "" {
+		grafanaURL = defaultGrafanaURL
+	}
 	incidentURL := fmt.Sprintf("%s/api/plugins/grafana-incident-app/resources/api/v1/", grafanaURL)
+	slog.Debug("Creating Incident client", "url", incidentURL, "api_key_set", apiKey != "")
```
csmarchbanks: URL also needs to be redacted (just so all of these comments need to be resolved).

csmarchbanks (changes requested):

sd2k:
Good call on the redacted URLs, updated.

csmarchbanks (approved):
A refactor on how we parse URLs might be nice in the future, but LGTM for now.

Issue #52, opened 2025-03-21
Allow 'list datasources' tool to filter by type


sd2k:
It'd be useful to avoid returning unnecessary datasources when the model knows what type it needs, to keep the context window small.

The Grafana API doesn't support it but we can just filter inside the tool.

PR #73, opened 2025-04-02
Allow filtering of datasources by type

csmarchbanks:
To reduce the amount of context passed back to the agent it is now possible to do client-side filtering of all datasources to just the type desired. This is implemented using case insensitive Contains so that the type does not have to be perfect.

Closes: #52


ioanarm (approved):
lgtm, tested with Claude as well and seems great

PR #74, opened 2025-04-03
Sift tools

ioanarm:
Adding Sift tools for 

* Create Investigation 
* List investigations with a limit parameter
* Get investigation 
* Get analyses
* Run ErrorPatternLogs Sift Check
* Run SlowRequests Sift Check

Also there is cloud test added checking the get tools for sift. 

## Implementation Approach:
I initially created separate tools for basic Sift API operations (create/get investigations) to reduce code duplication. However, for specific checks like ErrorPatternLogs and SlowRequests, I created dedicated tools that handle the entire workflow internally.

## Current Workflow Example:
When running ErrorPatternLogs in Sift, the process involves:
1. Creating an investigation with specific tags to limit checks (eg: ErrorPatternLogs)
2. Getting investigation details when complete
3. Retrieving the ErrorPatternLog analysis using the analysis ID from investigation data

This can be done through:
- Three separate tool calls when using the basic API tools
- A single tool call with our dedicated `run_error_pattern_logs` tool (which makes three API calls internally)

## Questions for Discussion:
Should we provide:
1. Basic API wrappers and let the model chain them together?
2. Higher-level tools that handle complete workflows?
3. A combination of both approaches? This maybe gives a bit more freedom to the LLM to provide more info if needed. 

**I'm leaning toward removing the createInvestigation tool and keeping just the specialized Sift workflow tools, but I'm open to suggestions.**

*Note: I've also refactored cloud tests into a utils file for better organization.*

### DIFF
--- a/README.md
+++ b/README.md
@@ -31,8 +31,8 @@ This provides access to your Grafana instance and the surrounding ecosystem.
   - [x] List Investigations with a limit parameter
   - [x] Get Investigation
   - [x] Get Analyses
-  - [x] Create an Investigation only running ErrorPatternLogs Check
-  - [x] Create an Investigation only running SlowRequests Check
+  - [x] Find error patterns in logs using Sift
+  - [x] Find slow requests using Sift
   - [ ] Add tools on the other Sift Checks
 - [ ] Alerting
   - [x] List and fetch alert rule information
@@ -77,15 +77,15 @@ the OnCall tools, use `--disable-oncall`.
 | `list_alert_rules`                | Alerting    | List alert rules                                                   |
 | `get_alert_rule_by_uid`           | Alerting    | Get alert rule by UID                                              |
 | `list_oncall_schedules`           | OnCall      | List schedules from Grafana OnCall                                 |
-| `get_oncall_shift`                | OnCall      | Get details for a specific OnCall shift                           |
+| `get_oncall_shift`                | OnCall      | Get details for a specific OnCall shift                            |
 | `get_current_oncall_users`        | OnCall      | Get users currently on-call for a specific schedule                |
 | `list_oncall_teams`               | OnCall      | List teams from Grafana OnCall                                     |
 | `list_oncall_users`               | OnCall      | List users from Grafana OnCall                                     |
-| `get_investigation`               | Sift        | Retrieve an existing Sift investigation by its UUID                     |
-| `get_analysis`                    | Sift        | Retrieve a specific analysis from a Sift investigation                 |
-| `list_investigations`             | Sift        | Retrieve a list of Sift investigations with an optional limit           |
-| `run_error_pattern_logs`          | Sift        | Create a Sift investigation with ErrorPatternLogs check and get results |
-| `run_slow_requests_check`         | Sift        | Create a Sift investigation with SlowRequests check and get results    |
+| `get_investigation`               | Sift        | Retrieve an existing Sift investigation by its UUID                |
+| `get_analysis`                    | Sift        | Retrieve a specific analysis from a Sift investigation             |
+| `list_investigations`             | Sift        | Retrieve a list of Sift investigations with an optional limit      |
+| `find_error_pattern_logs`         | Sift        | Finds elevated error patterns in Loki logs.                        |
+| `find_slow_requests`              | Sift        | Finds slow requests from the relevant tempo datasources.           |
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -81,7 +81,6 @@ the OnCall tools, use `--disable-oncall`.
 | `get_current_oncall_users`        | OnCall      | Get users currently on-call for a specific schedule                |
 | `list_oncall_teams`               | OnCall      | List teams from Grafana OnCall                                     |
 | `list_oncall_users`               | OnCall      | List users from Grafana OnCall                                     |
-| `create_investigation`            | Sift        | Create a new Sift investigation to analyze data from different datasources |
 | `get_investigation`               | Sift        | Retrieve an existing Sift investigation by its UUID                     |
 | `get_analysis`                    | Sift        | Retrieve a specific analysis from a Sift investigation                 |
 | `list_investigations`             | Sift        | Retrieve a list of Sift investigations with an optional limit           |
--- a/README.md
+++ b/README.md
@@ -26,7 +26,14 @@ This provides access to your Grafana instance and the surrounding ecosystem.
   - [x] Label values
   - [x] Stats
 - [x] Search, create, update and close incidents
-- [ ] Start Sift investigations and view the results
+- [x] Start Sift investigations and view the results
+  - [x] Create Investigations
+  - [x] List Investigations with a limit parameter
+  - [x] Get Investigation
+  - [x] Get Analyses
+  - [x] Create an Investigation only running ErrorPatternLogs Check
+  - [x] Create an Investigation only running SlowRequests Check
+  - [ ] Add tools on the other Sift Checks
 - [ ] Alerting
   - [x] List and fetch alert rule information
   - [x] Get alert rule statuses (firing/normal/error/etc.)
@@ -74,6 +81,12 @@ the OnCall tools, use `--disable-oncall`.
 | `get_current_oncall_users`        | OnCall      | Get users currently on-call for a specific schedule                |
 | `list_oncall_teams`               | OnCall      | List teams from Grafana OnCall                                     |
 | `list_oncall_users`               | OnCall      | List users from Grafana OnCall                                     |
+| `create_investigation`            | Sift        | Create a new Sift investigation to analyze data from different datasources |
+| `get_investigation`               | Sift        | Retrieve an existing Sift investigation by its UUID                     |
+| `get_analysis`                    | Sift        | Retrieve a specific analysis from a Sift investigation                 |
+| `list_investigations`             | Sift        | Retrieve a list of Sift investigations with an optional limit           |
+| `run_error_pattern_logs`          | Sift        | Create a Sift investigation with ErrorPatternLogs check and get results |
+| `run_slow_requests_check`         | Sift        | Create a Sift investigation with SlowRequests check and get results    |
 
 ## Usage
 
--- a/tools/sift.go
+++ b/tools/sift.go
@@ -35,100 +35,56 @@ const (
 )
 
 type InvestigationRequest struct {
-	AlertLabels map[string]string `json:"alertLabels,omitempty" gorm:"-"`
+	AlertLabels map[string]string `json:"alertLabels,omitempty"`
 	Labels      map[string]string `json:"labels"`
 
-	Start time.Time `json:"start" gorm:"not null"`
-	End   time.Time `json:"end" gorm:"not null"`
-
-	// TODO: Add this when we have a new investigation source field InvestigationSourceTypeMCP.
-	// InvestigationSource InvestigationSource `json:"investigationSource" gorm:"embedded;embeddedPrefix:investigation_source_"`
+	Start time.Time `json:"start"`
+	End   time.Time `json:"end"`
 
 	QueryURL string `json:"queryUrl"`
 
-	Checks []string `json:"checks" gorm:"serializer:json"`
-}
-
-// AnalysisStep represents a single step in the analysis process.
-type AnalysisStep struct {
-	CreatedAt time.Time `json:"created" validate:"isdefault"`
-	// State that the Analysis is entering.
-	State string `json:"state"`
-	// The exit message of the step. Can be empty if the step was successful.
-	ExitMessage string `json:"exitMessage"`
-	// Runtime statistics for this step
-	Stats map[string]interface{} `json:"stats,omitempty"`
-}
-
-type AnalysisEvent struct {
-	StartTime   time.Time              `json:"startTime"`
-	EndTime     time.Time              `json:"endTime"`
-	Name        string                 `json:"name"`
-	Description string                 `json:"description,omitempty"`
-	Details     map[string]interface{} `json:"details"`
+	Checks []string `json:"checks"`
 }
 
 // Interesting: The analysis complete with results that indicate a probable cause for failure.
 type AnalysisResult struct {
-	Successful      bool                   `json:"successful"`
-	Interesting     bool                   `json:"interesting"`
-	Message         string                 `json:"message"`
-	MarkdownSummary string                 `json:"-" gorm:"-"`
-	Details         map[string]interface{} `json:"details"`
-	Events          []AnalysisEvent        `json:"events,omitempty" gorm:"serializer:json"`
+	Successful  bool                   `json:"successful"`
+	Interesting bool                   `json:"interesting"`
+	Message     string                 `json:"message"`
+	Details     map[string]interface{} `json:"details"`
+}
+
+type AnalysisMeta struct {
+	Items []Analysis `json:"items"`
 }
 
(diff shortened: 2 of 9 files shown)


annanay25:
I agree with removing CreateInvestigation, in-fact should we restructure this to have a tool to `Run ErrorPatternLogs Sift Check` and wait for its results? If we don't do this we would have to add additional logic in the frontend to poll for results. Similarly for `SlowRequests`. We can just invoke and wait for results, users might be ok waiting for up to 20-30s.

tools/sift.go
```diff
@@ -0,0 +1,608 @@
...
+	investigationResponse := struct {
+		Status string        `json:"status"`
+		Data   Investigation `json:"data"`
+	}{}
+
+	if err := json.Unmarshal(buf, &investigationResponse); err != nil {
+		return nil, fmt.Errorf("failed to unmarshal response body: %w. body: %s", err, buf)
+	}
+
+	// Poll for investigation completion
```
ioanarm, line 461: @annanay25 we already poll and wait for the investigation to complete. 
ErrorPatternLogs call this createInvestigation function with specific parameters.

ioanarm:
> I agree with removing CreateInvestigation, in-fact should we restructure this to have a tool to `Run ErrorPatternLogs Sift Check` and wait for its results? If we don't do this we would have to add additional logic in the frontend to poll for results. Similarly for `SlowRequests`. We can just invoke and wait for results, users might be ok waiting for up to 20-30s.

Cool, I removed the tool. On the other comment see [here](https://github.com/grafana/mcp-grafana/pull/74#discussion_r2039032242)

README.md
```diff
@@ -72,6 +79,12 @@ This is useful if you don't use certain functionality or if you don't want to ta
 | `get_current_oncall_users`        | OnCall      | Get users currently on-call for a specific schedule                |
 | `list_oncall_teams`               | OnCall      | List teams from Grafana OnCall                                     |
 | `list_oncall_users`               | OnCall      | List users from Grafana OnCall                                     |
+| `create_investigation`            | Sift        | Create a new Sift investigation to analyze data from different datasources |
```
annanay25: nit: need a docs update.

README.md
```diff
@@ -72,6 +79,12 @@ This is useful if you don't use certain functionality or if you don't want to ta
...
+| `get_investigation`               | Sift        | Retrieve an existing Sift investigation by its UUID                     |
+| `get_analysis`                    | Sift        | Retrieve a specific analysis from a Sift investigation                 |
+| `list_investigations`             | Sift        | Retrieve a list of Sift investigations with an optional limit           |
```
annanay25: For `get_investigation`, `get_analysis` and `list_investigations` - I'm not fully convinced we need to expose that level of detail to the LLM, I'd like to see examples of where we would need this exposed to the LLM and whether we can build some more abstract tools that can help. 

For ex: right now, if I wanted to find an investigation with a given labelset, say `{cluster='prod-us-central-0', namespace='machine-learning'}`, the LLM would need to fire 3 queries: `list_investigations` -> `get_investigation` -> `get_analysis` (x3) to retrieve all the data. This will require several tokens to be used in LLM calls when we send the entire list of investigations to it. Instead, can we build a higher abstraction that can take in a labelset and directly retrieve analysis for it without burning through as many tokens?

Curious if there's a use-case I'm missing or any other actions that can be enabled by the current API.

tools/sift.go
annanay25: very nice 👍

README.md
ioanarm: Yeah that's true. Adding the tools there is mainly saving time for the user in case there is anything already running in Sift. 
There is an alert in namespace x --> Ask the assistant if there are any investigations for this namespace --> 
`List investigations` tool used --> Listing all recent investigations. 

If the user wants to grab data then they can request using the ID provided. 

image

That's the main motivation behind exposing the tools. I do agree though that we need to have maybe more targeted tools. Let's discuss in the MCP call later.

tools/sift.go
```diff
@@ -0,0 +1,608 @@
...
+
+	return errorPatternLogsAnalysis, nil
+}
+
+// RunErrorPatternLogs is a tool for running an ErrorPatternLogs check
+var RunErrorPatternLogs = mcpgrafana.MustTool(
+	"run_error_pattern_logs",
+	"Creates a Sift investigation with ErrorPatternLogs check, waits for it to complete, and returns the analysis results. This tool triggers an investigation with the ErrorPatternLogs check in the relevant Loki datasource. It investigates if the there are elevated errors rates in the logs compared to the last day's average and returns the error pattern found, if any.",
+	runErrorPatternLogs,
+)
```
sd2k: My gut feeling is that these tools should be a bit less Sift-y in their names and descriptions. For example, this one could be named 'find_error_log_patterns' and the description would be closer to just the last sentence of the existing description, since an LLM is unlikely to know what Sift is, or what a `ErrorPatternLogs check` means.

tools/sift.go
```diff
@@ -0,0 +1,608 @@
...
+	FailureReason string `json:"failureReason,omitempty"`
+
+	// Analyses contains metadata about the investigation's analyses
+	Analyses AnalysisMeta `json:"analyses"`
+}
+
+type RequestData struct {
+	Labels map[string]string `json:"labels"`
+	Checks []string          `json:"checks"`
+}
```
sd2k: I think some of these types, annotations etc are redundant here (e.g. the `gorm` tags aren't needed and some of the types are unused). Would be good to keep as many internals out of here as possible.

tools/sift.go
ioanarm: Yeah that's true. I will refactor it a bit.

ioanarm: Removed the "sift" mentions from error pattern logs and slow requests check. 
Sift mentions in general I think are good especially if we go further and add another layer to MCP explaining what Sift is and how to use it (a prompt maybe), same with OnCall etc. I haven't removed it from the rest of the tools as they are very specific, list investigations, get investigations etc. Open to discuss this though, maybe we can completely remove Sift mentions for now 🙄

sd2k: > I haven't removed it from the rest of the tools as they are very specific, list investigations, get investigations etc. Open to discuss this though, maybe we can completely remove Sift mentions for now 🙄

Yeah I think that's fine 🙂 I would even consider changing those ones to 'List Sift investigations' so it matches the docs & UI and is clearly distinct from any informal kind of 'investigation'!

README.md
```diff
@@ -26,7 +26,14 @@ This provides access to your Grafana instance and the surrounding ecosystem.
...
   - [x] Stats
 - [x] Search, create, update and close incidents
-- [ ] Start Sift investigations and view the results
+- [x] Start Sift investigations and view the results
+  - [x] Create Investigations
+  - [x] List Investigations with a limit parameter
+  - [x] Get Investigation
+  - [x] Get Analyses
+  - [x] Create an Investigation only running ErrorPatternLogs Check
+  - [x] Create an Investigation only running SlowRequests Check
```
```suggestion
  - [x] Find error patterns in logs using Sift
  - [x] Find slow requests using Sift
```

tools/sift.go
```diff
@@ -0,0 +1,565 @@
...
+		return nil, fmt.Errorf("ErrorPatternLogs analysis not found in investigation %s", completedInvestigation.ID)
+	}
+
+	return errorPatternLogsAnalysis, nil
+}
+
+// FindErrorPatternLogs is a tool for running an ErrorPatternLogs check
+var FindErrorPatternLogs = mcpgrafana.MustTool(
+	"find_error_pattern_logs",
+	"Creates an investigation to search for error patterns in logs, waits for it to complete, and returns the analysis results. This tool triggers an investigation in the relevant Loki datasource to determine if there are elevated error rates compared to the last day's average, and returns the error pattern found, if any.",
```
sd2k: Shall we get rid of mentions of an 'investigation' here too, so it looks like the tool just opaquely finds error patterns in logs? (the implementation isn't super important imo)

Same with slow requests

tools/sift.go
```diff
@@ -0,0 +1,565 @@
...
+		Transport: &authRoundTripper{
+			apiKey:     apiKey,
+			underlying: http.DefaultTransport,
+		},
+	}
+	return &SiftClient{
+		client: client,
+		url:    url,
+	}
+}
```
sd2k: These should be unexported, they shouldn't be needed outside this package 👍

tools/sift.go
```diff
@@ -0,0 +1,565 @@
...
+
+	// Status describes the state of the investigation (pending, running, failed, or finished).
+	Status InvestigationStatus `json:"status"`
+
+	// FailureReason is a short human-friendly string that explains the reason that the
+	// investigation failed.
+	FailureReason string `json:"failureReason,omitempty"`
+
+	Analyses AnalysisMeta `json:"analyses"`
+}
```
sd2k: I think these can be unexported too.

tools/sift.go
```diff
@@ -0,0 +1,565 @@
...
+	// Get the standard Grafana URL and API key
+	grafanaURL, grafanaAPIKey := mcpgrafana.GrafanaURLFromContext(ctx), mcpgrafana.GrafanaAPIKeyFromContext(ctx)
+
+	client := NewSiftClient(grafanaURL, grafanaAPIKey)
+
+	return client, nil
+}
+
+// CheckType represents the type of analysis check to perform
+type CheckType string
```
sd2k: Should also be unexported 🙂

README.md
```diff
@@ -74,6 +81,11 @@ the OnCall tools, use `--disable-oncall`.
 | `get_current_oncall_users`        | OnCall      | Get users currently on-call for a specific schedule                |
 | `list_oncall_teams`               | OnCall      | List teams from Grafana OnCall                                     |
 | `list_oncall_users`               | OnCall      | List users from Grafana OnCall                                     |
+| `get_investigation`               | Sift        | Retrieve an existing Sift investigation by its UUID                     |
+| `get_analysis`                    | Sift        | Retrieve a specific analysis from a Sift investigation                 |
+| `list_investigations`             | Sift        | Retrieve a list of Sift investigations with an optional limit           |
+| `run_error_pattern_logs`          | Sift        | Create a Sift investigation with ErrorPatternLogs check and get results |
+| `run_slow_requests_check`         | Sift        | Create a Sift investigation with SlowRequests check and get results    |
```
sd2k: I think this should be `find_error_pattern_logs` / `find_slow_requests`.

We should change the description to remove Sift there too.

sd2k (changes requested):
A few comments on naming / nits on exporting inline.

I'm trying to avoid exporting anything other than the actual tool definitions & handlers and the `AddXTools` functions where possible!

tools/sift.go
```diff
@@ -0,0 +1,556 @@
...
+
+	analysis, err := client.getSiftAnalysis(ctx, investigationID, analysisID)
+	if err != nil {
+		return nil, fmt.Errorf("getting analysis: %w", err)
+	}
+
+	return analysis, nil
+}
+
+// GetAnalysis is a tool for retrieving a specific analysis from an investigation
```
sd2k:
```suggestion
// GetSiftAnalysis is a tool for retrieving a specific analysis from an investigation
```

tools/sift.go
```diff
@@ -0,0 +1,556 @@
...
+
+	investigations, err := client.listSiftInvestigations(ctx, args.Limit)
+	if err != nil {
+		return nil, fmt.Errorf("getting investigations: %w", err)
+	}
+
+	return investigations, nil
+}
+
+// ListInvestigations is a tool for retrieving a list of investigations
```
sd2k:
```suggestion
// ListSiftInvestigations is a tool for retrieving a list of investigations
```

tools/sift.go
```diff
@@ -0,0 +1,556 @@
...
+	listSiftInvestigations,
+)
+
+// FindErrorPatternLogsParams defines the parameters for running an ErrorPatternLogs check
+type FindErrorPatternLogsParams struct {
+	Name     string            `json:"name" jsonschema:"required,description=The name of the investigation"`
+	Labels   map[string]string `json:"labels" jsonschema:"required,description=Labels to scope the analysis"`
+	Start    time.Time         `json:"start,omitempty" jsonschema:"description=Start time for the investigation. Defaults to 30 minutes ago if not specified."`
+	End      time.Time         `json:"end,omitempty" jsonschema:"description=End time for the investigation. Defaults to now if not specified."`
+	QueryURL string            `json:"queryUrl,omitempty" jsonschema:"description=Optional query URL for the investigation"`
```
sd2k: I think we can just omit this, I doubt the model will know what to put here.

tools/sift.go
```diff
@@ -0,0 +1,556 @@
...
+// ListInvestigations is a tool for retrieving a list of investigations
+var ListSiftInvestigations = mcpgrafana.MustTool(
+	"list_sift_investigations",
+	"Retrieves a list of Sift investigations with an optional limit. If no limit is specified, defaults to 10 investigations.",
+	listSiftInvestigations,
+)
+
+// FindErrorPatternLogsParams defines the parameters for running an ErrorPatternLogs check
+type FindErrorPatternLogsParams struct {
+	Name     string            `json:"name" jsonschema:"required,description=The name of the investigation"`
```
sd2k:
```suggestion
	Name     string            `json:"name" jsonschema:"required,description=The name of the investigation"`
```

I wonder if we can omit this and generate something, or if it's better to just let the model provide some random name?

tools/sift.go
```diff
@@ -0,0 +1,556 @@
...
+	findErrorPatternLogs,
+)
+
+// FindSlowRequestsParams defines the parameters for running an SlowRequests check
+type FindSlowRequestsParams struct {
+	Name     string            `json:"name" jsonschema:"required,description=The name of the investigation"`
+	Labels   map[string]string `json:"labels" jsonschema:"required,description=Labels to scope the analysis"`
+	Start    time.Time         `json:"start,omitempty" jsonschema:"description=Start time for the investigation. Defaults to 30 minutes ago if not specified."`
+	End      time.Time         `json:"end,omitempty" jsonschema:"description=End time for the investigation. Defaults to now if not specified."`
+	QueryURL string            `json:"queryUrl,omitempty" jsonschema:"description=Optional query URL for the investigation"`
```
sd2k: Same re. omission

tools/sift.go
```diff
@@ -0,0 +1,556 @@
...
+// FindErrorPatternLogs is a tool for running an ErrorPatternLogs check
+var FindErrorPatternLogs = mcpgrafana.MustTool(
+	"find_error_pattern_logs",
+	"Searches Loki logs for elevated error patterns compared to the last day's average, waits for the analysis to complete, and returns the results including any patterns found.",
+	findErrorPatternLogs,
+)
+
+// FindSlowRequestsParams defines the parameters for running an SlowRequests check
+type FindSlowRequestsParams struct {
+	Name     string            `json:"name" jsonschema:"required,description=The name of the investigation"`
```
sd2k: Same comment re. omission

tools/sift.go
ioanarm: True 🙄

tools/sift.go
ioanarm: It does an ok job generating a name (see below). Maybe generating something to show the origin is better. Though we can add an investigation type for this. 🤔 
image

sd2k: Yeah it'd definitely be nice to include some info on origin/user here eventually. This is good for now though!

sd2k (approved):
LGTM! Nice one

PR #78, opened 2025-04-04
Update incident 'limit' to required

InTheCloudDan:
Minor change to make the limit required.


csmarchbanks (approved):
Thanks!

csmarchbanks:
Actually, I probably prefer the approach in https://github.com/grafana/mcp-grafana/pull/77, setting a default limit allows the LLM to do a bit less.

PR #81, opened 2025-04-05
Add listing all contact points with pagination

sunwupark:
Added listContactPoints function to fetch and paginate contact points from Grafana. 
Supports filtering by name and returns summarized data (UID, Name, Type). 
Tests included for pagination, filtering, and invalid parameters.

### DIFF
--- a/tools/alerting.go
+++ b/tools/alerting.go
@@ -13,7 +13,7 @@ import (
 )
 
 const (
-	DefaultListAlertRulesLimit = 100
+	DefaultListAlertRulesLimit    = 100
 	DefaultListContactPointsLimit = 100
 )
 
@@ -216,7 +216,7 @@ func listContactPoints(ctx context.Context, args ListContactPointsParams) ([]con
 		return nil, fmt.Errorf("list contact points: %w", err)
 	}
 
-	pagedContactPoints, err := applyContactPointPagination(response.Payload, args.Limit)
+	pagedContactPoints, err := applyLimitToContactPoints(response.Payload, args.Limit)
 	if err != nil {
 		return nil, fmt.Errorf("list contact points: %w", err)
 	}
--- a/tools/alerting.go
+++ b/tools/alerting.go
@@ -216,12 +216,12 @@ func listContactPoints(ctx context.Context, args ListContactPointsParams) ([]con
 		return nil, fmt.Errorf("list contact points: %w", err)
 	}
 
-	pagedContactPoints, err := applyLimitToContactPoints(response.Payload, args.Limit)
+	filteredContactPoints, err := applyLimitToContactPoints(response.Payload, args.Limit)
 	if err != nil {
 		return nil, fmt.Errorf("list contact points: %w", err)
 	}
 
-	return summarizeContactPoints(pagedContactPoints), nil
+	return summarizeContactPoints(filteredContactPoints), nil
 }
 
 func summarizeContactPoints(contactPoints []*models.EmbeddedContactPoint) []contactPointSummary {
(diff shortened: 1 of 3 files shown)


sunwupark:
image

tools/alerting.go
```diff
@@ -179,7 +181,93 @@ var GetAlertRuleByUID = mcpgrafana.MustTool(
...
+		result = append(result, contactPointSummary{
+			UID:  cp.UID,
+			Name: cp.Name,
+			Type: cp.Type,
+		})
+	}
+	return result
+}
+
+func applyContactPointPagination(items []*models.EmbeddedContactPoint, limit, page int) ([]*models.EmbeddedContactPoint, error) {
```
ioanarm, line 243: I'm wondering if we might simplify by focusing solely on the limit parameter for now.
Since the external API doesn't support server-side pagination, implementing just the limit would give us the key performance benefits while reducing code complexity. We can also add some comments on adding pagination when the API supports it! What do you think?

ioanarm: I'm wondering if we might simplify by focusing solely on the limit parameter for now.
Since the external API doesn't support server-side pagination, implementing just the limit would give us the key performance benefits while reducing code complexity. We can also add some comments on adding pagination when the API supports it! What do you think?

ioanarm (commented):
Thanks so much for your contribution @sunwupark ! 
One small comment from my side! :)

tools/alerting.go
sunwupark: Thanks for the suggestion!
I’ve applied your feedback and simplified the pagination logic

sunwupark:
I have also removed the page parameter from the alert test file.

tools/alerting.go
```diff
@@ -179,7 +181,81 @@ var GetAlertRuleByUID = mcpgrafana.MustTool(
...
+		result = append(result, contactPointSummary{
+			UID:  cp.UID,
+			Name: cp.Name,
+			Type: cp.Type,
+		})
+	}
+	return result
+}
+
+func applyContactPointPagination(items []*models.EmbeddedContactPoint, limit int) ([]*models.EmbeddedContactPoint, error) {
```
ioanarm: tiny last comment, can we change the name of this function now?

sunwupark: Would you be okay with renaming the function to applyLimitToContactPoints? I think it accurately reflects the current functionality.

tools/alerting.go
```diff
@@ -179,7 +181,81 @@ var GetAlertRuleByUID = mcpgrafana.MustTool(
...
+	if args.Name != nil {
+		params.Name = args.Name
+	}
+
+	response, err := c.Provisioning.GetContactpoints(params)
+	if err != nil {
+		return nil, fmt.Errorf("list contact points: %w", err)
+	}
+
+	pagedContactPoints, err := applyLimitToContactPoints(response.Payload, args.Limit)
```
ioanarm:
```suggestion
	filteredContactPoints, err := applyLimitToContactPoints(response.Payload, args.Limit)
```
? or something similar?

tools/alerting.go
```diff
@@ -179,7 +181,81 @@ var GetAlertRuleByUID = mcpgrafana.MustTool(
...
+	if err != nil {
+		return nil, fmt.Errorf("list contact points: %w", err)
+	}
+
+	return summarizeContactPoints(pagedContactPoints), nil
```
ioanarm: same as above

tools/alerting_test.go
```diff
@@ -399,3 +399,85 @@ func TestAlertingTools_GetAlertRuleByUID(t *testing.T) {
...
+		ctx := newTestContext()
+		result, err := listContactPoints(ctx, ListContactPointsParams{})
+		require.NoError(t, err)
+		require.ElementsMatch(t, allExpectedContactPoints, result)
+	})
+
+	t.Run("list one contact point", func(t *testing.T) {
+		ctx := newTestContext()
+
+		// Get the first page with limit 1
```
ioanarm:
```suggestion
		// Get contant points with limit 1
```

ioanarm (approved):
Just to clean it up!

sunwupark:
yup! changed and commited

Issue #66, opened 2025-03-27
Handle Invalid Datasource UID in prometheus tools


ioanarm:
When attempting to use Prometheus tools with a non-existent datasource UID, the current error handling returns a generic `Client error: 404` without providing specific context about the datasource lookup failure.

Return a clear error message indicating the datasource UID is invalid

PR #84, opened 2025-04-07
Check if datasource exists before querying

ioanarm:
Related to [66](https://github.com/grafana/mcp-grafana/issues/66)


csmarchbanks (commented):
I think this is fine, but also, could we just do it based on the response code from the main call? If it is a 404 return this message, or is it some other error?

csmarchbanks (approved):

PR #85, opened 2025-04-07
update Incident List query

InTheCloudDan:
Add more description for status, remove 'and' in query.

I validate the removing 'and' by looking at the query the frontend makes. Tested this locally.

### DIFF
--- a/tools/incident.go
+++ b/tools/incident.go
@@ -12,7 +12,7 @@ import (
 type ListIncidentsParams struct {
 	Limit  int    `json:"limit" jsonschema:"description=The maximum number of incidents to return"`
 	Drill  bool   `json:"drill" jsonschema:"description=Whether to include drill incidents"`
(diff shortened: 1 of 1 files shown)


csmarchbanks (approved):
Thanks!

tools/incident.go
```diff
@@ -12,7 +12,7 @@ import (
 type ListIncidentsParams struct {
 	Limit  int    `json:"limit" jsonschema:"description=The maximum number of incidents to return"`
 	Drill  bool   `json:"drill" jsonschema:"description=Whether to include drill incidents"`
-	Status string `json:"status" jsonschema:"description=The status of the incidents to include"`
+	Status string `json:"status" jsonschema:"description=The status of the incidents to include. Valid values: 'active', 'resolved', 'mitigated'"`
```
csmarchbanks: I think `active` and `resolved` are the only standard ones. I think I am ok merging this, but I wonder if we could populate this on startup in the future by fetching the customized values.

PR #87, opened 2025-04-08
Add state retrieval for alert rules

alexander-akhmetov:
Updates the existing alert tools to retrieve state information (firing, normal, pending, etc.) for alerts.

Since the Alerting provisioning API does not expose this information, I had to add a custom alerting client that uses the same [API](https://github.com/grafana/grafana/blob/87e69e13493899cfa3083ba9327dfeb5a9a4b093/pkg/services/ngalert/api/tooling/definitions/prom.go#L15) used by the UI. This API is not part of the officially stable public APIs, but the client makes it possible to add other endpoints, for example alert state history.

This API supports pagination of the alert rule groups, but I haven't added anything for it yet. Can do a follow-up :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -29,9 +29,9 @@ This provides access to your Grafana instance and the surrounding ecosystem.
 - [ ] Start Sift investigations and view the results
 - [ ] Alerting
   - [x] List and fetch alert rule information
-  - [ ] Get alert rule statuses (firing/normal/error/etc.)
+  - [x] Get alert rule statuses (firing/normal/error/etc.)
   - [ ] Create and change alert rules
-  - [ ] List contact points
+  - [x] List contact points
   - [ ] Create and change contact points
 - [x] Access Grafana OnCall functionality
   - [x] List and manage schedules
--- a/tools/alerting_client.go
+++ b/tools/alerting_client.go
@@ -20,20 +20,20 @@ const (
 	rulesEndpointPath = "/api/prometheus/grafana/api/v1/rules"
 )
 
-type AlertingClient struct {
+type alertingClient struct {
 	baseURL    *url.URL
 	apiKey     string
 	httpClient *http.Client
 }
 
-func newAlertingClientFromContext(ctx context.Context) (*AlertingClient, error) {
+func newAlertingClientFromContext(ctx context.Context) (*alertingClient, error) {
 	baseURL := strings.TrimRight(mcpgrafana.GrafanaURLFromContext(ctx), "/")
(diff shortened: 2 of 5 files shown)


README.md
```diff
@@ -29,9 +29,9 @@ This provides access to your Grafana instance and the surrounding ecosystem.
 - [ ] Start Sift investigations and view the results
 - [ ] Alerting
   - [x] List and fetch alert rule information
-  - [ ] Get alert rule statuses (firing/normal/error/etc.)
+  - [x] Get alert rule statuses (firing/normal/error/etc.)
   - [ ] Create and change alert rules
-  - [ ] List contact points
+  - [x] List contact points
```
alexander-akhmetov, line 34: Noticed this was done in another PR: https://github.com/grafana/mcp-grafana/pull/81

tools/alerting_client.go
```diff
@@ -0,0 +1,133 @@
...
+
+	mcpgrafana "github.com/grafana/mcp-grafana"
+)
+
+const (
+	defaultTimeout    = 30 * time.Second
+	rulesEndpointPath = "/api/prometheus/grafana/api/v1/rules"
+)
+
+type AlertingClient struct {
```
sd2k: This probably doesn't need exporting - ideally we'd just have the actual tools exported from this `tools` package. (we may have missed some already but I'll try and fix that soon :sweat_smile:)

(same goes for the various structs below!)

alexander-akhmetov: Makes sense! Un-exported :)

alexander-akhmetov:
> I am curious if any of the changes in https://github.com/grafana/mcp-grafana/pull/95 have affected the use of base URL here - sorry if so!

Nope, all good!

alexander-akhmetov:
Does it make sense to have something similar to [IncidentClient](https://github.com/grafana/mcp-grafana/blob/2dbb53edf098759635c3d3570464ccb0d9a32eb5/mcpgrafana.go#L226) in mcpgrafana.go, `AlertingClientFromContext`? Or is it fine to initialize the client as it is now since it's the same as Grafana client basically?

sd2k:
Hm I think we have that separate `IncidentClient` in the context to enable connection reuse in the stdio case, but it doesn't feel like a big enough deal really (and annoyingly we need to remember to add the client to the `ContextFunc` for every transport, which is easy to forget). So I'm happy to keep it where it is (I may even stop storing the `IncidentClient` in context one day).

sd2k (approved):
LGTM, thanks again! 🥳

PR #93, opened 2025-04-14
fix: mention that stepSeconds and endRfc3339 are required for Prometheus range queries

### LINKED ISSUE #75: stepSeconds required for queryType 'range'
Should we update the doc string to specific that `stepSeconds` are required for 'range' queries? Running into a case now where that was not clear to the LLM based on the current doc: https://github.com/grafana/mcp-grafana/blob/03027288f63bfef7b3ac506d33ca32ba52eb9935/tools/prometheus.go#L84

Image

sd2k:
Fixes #75.


annanay25 (approved):
lgtm.

PR #94, opened 2025-04-15
fix: fall back to Grafana URL/API key from env vars in SSE mode

### LINKED ISSUE #91: SSE not respect Grafana ENV
from the https://github.com/grafana/mcp-grafana/blob/main/mcpgrafana.go#L143, it seems it is still reading the header 

Image

sd2k:
If there are no relevant headers, fall back to the Grafana URL and API key from the environment variables, then the defaults.

Fixes #91.


mcpgrafana_test.go
```diff
@@ -21,6 +23,52 @@ func TestExtractIncidentClientFromEnv(t *testing.T) {
...
+		req.Header.Set(grafanaURLHeader, "http://my-test-url.grafana.com")
+		require.NoError(t, err)
+		ctx := ExtractIncidentClientFromHeaders(context.Background(), req)
+
+		client := IncidentClientFromContext(ctx)
+		require.NotNil(t, client)
+		assert.Equal(t, "http://my-test-url.grafana.com/api/plugins/grafana-irm-app/resources/api/v1/", client.RemoteHost)
+	})
+
+	t.Run("no headers, with env", func(t *testing.T) {
```
ioanarm: this is with headers and env right?

sd2k: yep! fixed

ioanarm (approved):

PR #96, opened 2025-04-15
fix: allow tools to be disabled by category

### LINKED ISSUE #92: Tools can no longer be disabled
The Python version of the MCP server allowed categories of tools to be disabled using config. This is no longer possible in the Go version, but it's a useful feature because some clients (e.g. Cursor) have a hard limit on the number of tools they support.

It should be quite easy to add this back in. We currently add all tools in [main.go](https://github.com/grafana/mcp-grafana/blob/78e604e723620270cec3c63121ab478d1070f5f6/cmd/mcp-grafana/main.go#L21-L28); each category could simply be gated behind a flag or config setting (defaulting to 'on' for now).

sd2k:
Fixes #92.


ioanarm (approved):
Nice 🚀

PR #97, opened 2025-04-16
feat: add -debug flag to allow debugging of Grafana requests

### LINKED ISSUE #80: [Feature require]Support Client Debug option
Hi there,

There is an `Debug` option available in TransportConfig, [link](https://github.com/grafana/grafana-openapi-client-go/blob/8d7e1f158f65542f533d67b8fc335f4491a62a39/templates/clientFacade.gotmpl#L132). But there is no way to open it on mcp server.

It will be very helpful if this can be opened during development.

sd2k:
Should resolve #80.

### DIFF
--- a/mcpgrafana.go
+++ b/mcpgrafana.go
@@ -41,6 +41,26 @@ func urlAndAPIKeyFromHeaders(req *http.Request) (string, string) {
 type grafanaURLKey struct{}
 type grafanaAPIKeyKey struct{}
 
+// grafanaDebugKey is the context key for the Grafana transport's debug flag.
+type grafanaDebugKey struct{}
+
+// WithGrafanaDebug adds the Grafana debug flag to the context.
+func WithGrafanaDebug(ctx context.Context, debug bool) context.Context {
+	if debug {
+		slog.Info("Grafana transport debug mode enabled")
+	}
+	return context.WithValue(ctx, grafanaDebugKey{}, debug)
+}
+
+// GrafanaDebugFromContext extracts the Grafana debug flag from the context.
+// If the flag is not set, it returns false.
+func GrafanaDebugFromContext(ctx context.Context) bool {
+	if debug, ok := ctx.Value(grafanaDebugKey{}).(bool); ok {
+		return debug
+	}
+	return false
+}
+
 // ExtractGrafanaInfoFromEnv is a StdioContextFunc that extracts Grafana configuration
 // from environment variables and injects a configured client into the context.
 var ExtractGrafanaInfoFromEnv server.StdioContextFunc = func(ctx context.Context) context.Context {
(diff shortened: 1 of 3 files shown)


mcpgrafana.go
```diff
@@ -135,6 +155,8 @@ var ExtractGrafanaClientFromEnv server.StdioContextFunc = func(ctx context.Conte
 		cfg.APIKey = apiKey
 	}
 
+	cfg.Debug = GrafanaDebugFromContext(ctx)
```
annanay25, line 158: should we pass all config options through env vars? wondering if using one way or the other will make it maintainable over time

csmarchbanks (commented):
🤔 I am not seeing much different in the logs when trying with Claude Desktop. Is there something I should be looking for?

sd2k:
When I run the server with `--debug` I see this in stderr:

```
❯ GRAFANA_URL=http://localhost:3000 go run ./cmd/mcp-grafana --debug
time=2025-04-17T09:18:28.098+01:00 level=INFO msg="Starting Grafana MCP server using stdio transport"
time=2025-04-17T09:18:28.098+01:00 level=INFO msg="Grafana transport debug mode enabled"
time=2025-04-17T09:18:28.098+01:00 level=INFO msg="Using Grafana configuration" url=http://localhost:3000 api_key_set=false
{"jsonrpc": "2.0", "id": 0, "method": "tools/call", "params": {"name": "list_datasources" }}
GET /api/datasources HTTP/1.1
Host: localhost:3000
User-Agent: Go-http-client/1.1
Accept: application/json
Accept-Encoding: gzip


HTTP/1.1 401 Unauthorized
Content-Length: 123
Cache-Control: no-store
Content-Type: application/json; charset=UTF-8
Date: Thu, 17 Apr 2025 08:18:29 GMT
X-Content-Type-Options: nosniff
X-Frame-Options: deny
X-Xss-Protection: 1; mode=block

{
  "extra": null,
  "message": "Unauthorized",
  "messageId": "auth.unauthorized",
  "statusCode": 401,
  "traceID": ""
}

{"jsonrpc":"2.0","id":0,"error":{"code":-32603,"message":"list datasources: [GET /datasources][401] getDataSourcesUnauthorized {\"message\":\"Unauthorized\"}"}}
```

I haven't tried with Claude desktop to be fair, not sure where it dumps the logs...

mcpgrafana.go
sd2k: Yeah I do think we need to standardize here, and maybe revisit all of the config options to be honest. Ideally everything would be configurable via either env or CLI flag which we could get by using a more full featured CLI parsing library, but that feels like a future thing!

I chose a CLI flag because it feels similar to `--log-level` to me!

csmarchbanks (approved):
Well I see them today so maybe I just did something odd with the build yesterday... 🤷 

I'm good with the cli argument, it actually looks like grafana-runtime uses the DEBUG env var, but then overwrites it when creating the client. We could use that too in the future.

PR #98, opened 2025-04-16
Fix commas in JSONSchema descriptions and add linter

sd2k:
Commas in jsonschema `description` fields inside struct field tags need
to be escaped; otherwise, they're treated as another field and the
description is silently truncated. This commit fixes all instances of
that, and adds a linter which checks for the presence of any unescaped
commas in the `jsonschema` descriptions. The linter also includes a
`--fix` mode to automatically fix any issues, and is integrated in CI so
we should catch this in future. (Thanks Claude & Zed!)

### DIFF
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -19,14 +19,8 @@ jobs:
           go-version: "1.24"
           cache: true
 
-      # Use the golangci-lint action for improved caching etc.
-      - name: Lint with golangi-lint
-        uses: golangci/golangci-lint-action@v3
-        with:
-          version: v1.64
-
-      - name: Lint with custom linters
-        run: make lint-jsonschema
+      - name: Run linters
+        run: make lint
 
   test-unit:
     name: Test Unit
--- a/.github/workflows/go.yml
+++ b/.github/workflows/go.yml
@@ -2,7 +2,7 @@ name: Go
 
 on:
   push:
-    branches: [ main ]
+    branches: [main]
   pull_request:
 
 jobs:
@@ -16,13 +16,17 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.24'
+          go-version: "1.24"
           cache: true
 
-      - name: Lint
+      # Use the golangci-lint action for improved caching etc.
+      - name: Lint with golangi-lint
         uses: golangci/golangci-lint-action@v3
         with:
           version: v1.64
+
+      - name: Lint with custom linters
+        run: make lint-jsonschema
 
   test-unit:
     name: Test Unit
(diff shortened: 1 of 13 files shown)


sd2k:
Hmm the branch name is wrong (I started on another issue and got distracted) but I don't think that's a biggie.

annanay25 (commented):
lgtm, the only thing was i couldn't spot how this is integrated in CI?

> and is integrated in CI so we should catch this in future.

Shouldn't we have a `make lint` or something in the CI? I also don't see this in the Github actions output either: https://github.com/grafana/mcp-grafana/actions/runs/14501420096/job/40681760901?pr=98

sd2k:
Oh yeah good call! We use the golangci-lint action in CI rather than make lint, we should fix that, I'll do it tomorrow

internal/linter/jsonschema/jsonschema_lint.go
```diff
@@ -0,0 +1,230 @@
...
+	Field     string
+	Tag       string
+	FixedTag  string
+}
+
+// tagPattern matches jsonschema tags with description containing unescaped commas
+// It captures:
+// 1. The jsonschema tag
+// 2. Parts of the description containing unescaped commas
+var tagPattern = regexp.MustCompile(`jsonschema:"([^"]*)description=([^"]*[^\\],[^"]*)"`)
```
csmarchbanks: Does this fail if we use escaped quotes then have a comma inside of a jsonschema tag?

csmarchbanks (approved):
Approving, though I think the linter could fail in at least one case. Not sure how perfect we need to be...

internal/linter/jsonschema/jsonschema_lint.go
sd2k: It did not! I pushed a second commit which fixes & adds a test case for that.

.github/workflows/go.yml
```diff
@@ -16,14 +16,18 @@ jobs:
       - name: Set up Go
         uses: actions/setup-go@v5
         with:
-          go-version: '1.24'
+          go-version: "1.24"
           cache: true
 
-      - name: Lint
+      # Use the golangci-lint action for improved caching etc.
+      - name: Lint with golangi-lint
```
csmarchbanks: Why not just one step that does `make lint`? That way if we add future linters to the Makefile they will be automatically picked up.

csmarchbanks (approved):
One small comment, but :+1:

.github/workflows/go.yml
sd2k: Yeah I kept it this way because I _think_ the golangci-lint action gives us free caching but 🤷 I've switched to `make lint` instead.

Issue #69, opened 2025-03-31
Tool Classification - TBC


ioanarm:
As the MCP server continues to expand its integration with various Grafana products and services, the number of available tools exposed to the LLM has grown substantially. With this expansion across Prometheus, Loki, Grafana Dashboards, OnCall, and other services, the LLM faces an increasingly complex decision when selecting the right tool for each query. We need a smarter tool filtering system to maintain performance and accuracy as we continue to scale our support for the Grafana ecosystem.

We need to add a way to help with:
* Categorization system for all tools (Prometheus, Loki, OnCall, etc.)
* Dynamic tool filtering to only expose the relevant subset of tools to the LLM
* Description enhancements for better tool selection

This solution addresses several important benefits:
* Reduced token usage: Only sending relevant tool descriptions to the model
* Improved accuracy: The model is less likely to select inappropriate tools
* Faster responses: Less context for the model to process
* Better user experience: More relevant and precise tool selection

ioanarm:
Closing this as not planned. At the moment there is no way to do tool classification on the server side. Though there is logic to disable tools if needed from the config. We will probably revisit this soon.

PR #100, opened 2025-04-17
Add allow list for tool categories

csmarchbanks:
Right now it is only possible to disable categories of tools which is tedious for users who want to only use one or two categories, and means that new tools added in the future will automatically show up. Add the ability to also set an allow-list of tools that can then be overridden using the --disable-<category> flags.

Related to https://github.com/grafana/mcp-grafana/issues/69, but I would like that issue to do more. Specifically some sort of filtering or discovery when all tools are configured.

### DIFF
--- a/cmd/mcp-grafana/main.go
+++ b/cmd/mcp-grafana/main.go
@@ -6,6 +6,8 @@ import (
 	"fmt"
 	"log/slog"
 	"os"
+	"slices"
+	"strings"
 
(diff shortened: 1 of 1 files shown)


sd2k (approved):
:+1: I would love to overhaul config completely at some point but perhaps that can wait until the official MCP Go SDK is released somewhere down the line.

csmarchbanks:
Yeah, I agree. We will keep learning about different use cases and modes until then as well!

Issue #90, opened 2025-04-13
[Feature] Tool to update a dashboard?


OpeOnikute:
It would be useful to have an LLM modify or create dashboards. For example, I am troubleshooting some metrics and need a quick dashboard with the help of an LLM that has context. Or I ask an LLM to suggest improvements to a dashboard, and ask it to apply the suggested changes (after my approval).

I would imagine it involves updating the dashboard JSON and somehow contacting the Grafana API to update the dashboard in-place.

sd2k:
Absolutely agree these would be useful tools!

The main concern is the sheer size of the dashboard JSON; iterating on a dashboard a few times using the full JSON model would increase the context window quite a lot. I wonder if we could ensure we always return prettified JSON from any 'fetch dashboard' tools, and the editing tools allow the model to modify specific lines of the JSON rather than needing to supply the full JSON each time.

OpeOnikute:
I've been playing around locally, and getting it working with the entire dashboard JSON was pretty trivial.
I hadn't considered how that could impact context windows - that's a good call out. 

I'm planning to make a PR here, so I'll consider a solution there as well. If it doesn't work, maybe a simple disclaimer/TODO about context windows will suffice for now?

sd2k:
Sounds great, thanks. I think a disclaimer in the README and a TODO in the code would be enough for now, yep.

OpeOnikute:
I have created https://github.com/grafana/mcp-grafana/issues/101 to think of a general strategy before attempting to implement any solutions to the context-window issue. I will add the disclaimers in my Pull Request for now.

Issue #101, opened 2025-04-18
Strategy for managing context windows with Dashboard JSON


OpeOnikute:
We eventually want LLMs to understand and possibly modify dashboards. The best way to do this is to access the dashboard JSON. An alternative is to provide a screenshot of the dashboard to a model that can understand images, which doesn't scale and can miss crucial context.

Providing modification tools is fairly trivial, but we need to be careful about context windows. Dashboard JSONs can be very large and users can use up their context limits without knowing. We can use this issue to discuss possible solutions, while adding disclaimers in the code for a start.

Some early thoughts:
- We could aim to provide the dashboard JSON to LLMs (via the _get_dashboard_via_uid_ tool) only once. We then instruct the LLM to only send specific modifications. The update_dashboard tool can be built to take these specific updates instead, and validate them before applying.
- We could also provide features in _get_dashboard_via_uid_ that can fetch only specific parts of the dashboard. Maybe it could even be a new, unique tool like _get_dashboard_property_. LLMs can then use this to fetch only the context they need.

I can see possible use-cases for both, and a complete solution would probably involve more features.

ioanarm:
We could leverage [jsonpath](https://jsonpath.com/) for this too. It would let us update just specific parts of the JSON and then merge back into the original. 

@sd2k suggested this in our recent chat - definitely worth exploring that approach as well.

harry-xm:
If Grafana MCP server supports updating or creating a dashboard based on dashboard definitions with [Grafonnet](https://github.com/grafana/grafonnet), it might help with this issue because Jsonnet syntax is much more compact than JSON.

PR #102, opened 2025-04-18
Add tool to update/create a dashboard

OpeOnikute:
Adds a new tool to update/create a dashboard. Also added a warning/disclaimer about context windows, which will be solved longer-term with https://github.com/grafana/mcp-grafana/issues/101.

Closes https://github.com/grafana/mcp-grafana/issues/90.

### DIFF
--- a/tools/dashboard.go
+++ b/tools/dashboard.go
@@ -16,10 +16,37 @@ type GetDashboardByUIDParams struct {
 
 func getDashboardByUID(ctx context.Context, args GetDashboardByUIDParams) (*models.DashboardFullWithMeta, error) {
 	c := mcpgrafana.GrafanaClientFromContext(ctx)
-	
+
 	dashboard, err := c.Dashboards.GetDashboardByUID(args.UID)
 	if err != nil {
 		return nil, fmt.Errorf("get dashboard by uid %s: %w", args.UID, err)
+	}
+	return dashboard.Payload, nil
+}
+
+type UpdateDashboardParams struct {
+	Dashboard map[string]interface{} `json:"dashboard" jsonschema:"required,description=The full dashboard JSON"`
+	FolderUID string                 `json:"folderUid" jsonschema:"optional,description=The UID of the dashboard's folder"`
+	Message   string                 `json:"message" jsonschema:"optional,description=Set a commit message for the version history"`
+	Overwrite bool                   `json:"overwrite" jsonschema:"optional,description=Overwrite the dashboard if it exists. Otherwise create one"`
+	UserID    int64                  `json:"userId" jsonschema:"optional,ID of the user making the change"`
+}
+
+// updateDashboard can be used to save an existing dashboard, or create a new one.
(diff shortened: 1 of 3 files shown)


ioanarm (approved):
Looks good @OpeOnikute! Thanks for adding this tool. Works great for small dashboards, but yeah, the context window gets maxed out quickly with bigger ones like you mentioned. I tried updating a large dashboard and just got stuck in a "continue" loop 😅. Still super useful for many cases!

Tackling [this](https://github.com/grafana/mcp-grafana/issues/101) should solve the problem though.

OpeOnikute:
Thanks @ioanarm! Please can you help get this merged? 
```
This workflow requires approval from a maintainer. [Learn more about approving workflows.](https://docs.github.com/actions/managing-workflow-runs/approving-workflow-runs-from-public-forks)
```

OpeOnikute:
Fixed the lint error

tools/dashboard.go
```diff
@@ -16,20 +16,56 @@ type GetDashboardByUIDParams struct {
...
+
 var GetDashboardByUID = mcpgrafana.MustTool(
 	"get_dashboard_by_uid",
 	"Get dashboard by uid",
 	getDashboardByUID,
 )
 
+var UpdateDashboard = mcpgrafana.MustTool(
+	"update_dashboard",
+	"Update a dashboard",
```
sd2k:
```suggestion
	"Create or update a dashboard",
```

tools/dashboard.go
```diff
@@ -16,20 +16,56 @@ type GetDashboardByUIDParams struct {
...
 	if err != nil {
 		return nil, fmt.Errorf("get dashboard by uid %s: %w", args.UID, err)
 	}
 	return dashboard.Payload, nil
 }
 
+type UpdateDashboardParams struct {
+	Dashboard map[string]interface{} `json:"dashboard" jsonschema:"required,description=The full dashboard JSON"`
+	FolderUID string                 `json:"folderUid" jsonschema:"optional,description=The UID of the dashboard's folder"`
+	IsFolder  bool                   `json:"isFolder" jsonschema:"optional,description=Whether the dashboard is a folder"`
```
sd2k: I don't think we should expose this option to the model - it feels like a quirk of Grafana's dashboard API more than anything else. If we want the model to be able to create folders we should have a separate tool which just calls this API endpoint with `isFolder: true`.

sd2k (changes requested):
Hi @OpeOnikute, I just have one comment on the `IsFolder` param which I think should be omitted here, and a slightly updated name/description for the tool. Otherwise this looks good, thanks!

tools/dashboard.go
OpeOnikute: That's true. I noticed when testing that you can't even use it to save a new dashboard. The API tells you to create the folder directly instead. Removed 👍

sd2k (approved):
Nice, thank you! This should be useful for lots of people 👍

PR #105, opened 2025-04-21
First Asserts tool

xujiaxj:
Adding the first Asserts tool for an API call. It's not quite important what this tool does. It will probably go through more iterations in the future and we will add more tools. This is just a simple PR as the first step.

Tested with Claude Desktop:

image


tools/asserts.go
```diff
@@ -0,0 +1,135 @@
...
+	}
+
+	return &Client{
+		httpClient: client,
+		baseURL:    url,
+	}, nil
+}
+
+type GetAssertionsParams struct {
+	StartTime  int64  `json:"startTime" jsonschema:"description=The start time of the assertion status in RFC3339 format"`
```
ioanarm, line 33: the format here should be unix milliseconds I think.

tools/asserts.go
```diff
@@ -0,0 +1,135 @@
...
+	EndTime    int64  `json:"endTime" jsonschema:"description=The end time of the assertion status in RFC3339 format"`
```
ioanarm, line 34: same as above

tools/asserts.go
ioanarm: The format in the description here should be unix milliseconds I think.

tools/asserts.go
ioanarm: same here

ioanarm (commented):
Looks good, a couple of comments on the time format. 
I am trying to reproduce using ops and no luck so far. 

Maybe adding a cloud test along with a couple unit tests could do it. Happy to help!

tools/asserts.go
xujiaxj: The API takes unix milliseconds, so I have a `*1000` when making the call, but it probably makes sense to just tell AI to use unix milliseconds directly.

xujiaxj: There is another bug that makes it not returning assertion data though, the assertion type list needs to be lower-cased, i.e, `saturation` instead of `Saturation`. It's case-sensitive.

xujiaxj:
Not sure how to do the cloud test against `mcptests` stack as it does not have any metric.

tools/asserts.go
```diff
@@ -0,0 +1,133 @@
...
+
+	return &Client{
+		httpClient: client,
+		baseURL:    url,
+	}, nil
+}
+
+type GetAssertionsParams struct {
+	StartTime  int64  `json:"startTime" jsonschema:"description=The start time of the assertion status in unix milliseconds format"`
+	EndTime    int64  `json:"endTime" jsonschema:"description=The end time of the assertion status in unix milliseconds format"`
```
sd2k: FWIW elsewhere in the server we use `time.Time` and mention that the timestamp should be in RFC3339 in the description, which tends to work quite well (it's easier for a model to get an RFC3339 timestamp correct than a unix timestamp, I guess)

xujiaxj: Oh. If that's the case, I can revert back to use RFC3339 and do the `*1000` when calling Asserts API.

xujiaxj:
image

tools/asserts.go
sd2k, line 1: Could you un-export all of the structs (other than `GetAssertionsParams` and `GetAssertions`) so we keep the `tools` package clean when imported? :pray:

tools/asserts.go
```diff
@@ -0,0 +1,133 @@
...
+
+	return &Client{
+		httpClient: client,
+		baseURL:    url,
+	}, nil
+}
+
+type GetAssertionsParams struct {
+	StartTime  int64  `json:"startTime" jsonschema:"description=The start time of the assertion status in RFC3339 format"`
+	EndTime    int64  `json:"endTime" jsonschema:"description=The end time of the assertion status in RFC3339 format"`
```
sd2k: Hmm I don't think the jsonschema description matches the type any more; should the type be `time.Time` instead so the model can provide an RFC3339 string? (you can convert it to millis using [`t.UnixMilli`](https://pkg.go.dev/time#Time.UnixMilli))

tools/asserts.go
```diff
@@ -0,0 +1,133 @@
...
+	EntityType string `json:"entityType" jsonschema:"description=The type of the entity to list"`
+	EntityName string `json:"entityName" jsonschema:"description=The name of the entity to list"`
+	Env       string `json:"env" jsonschema:"description=The env of the entity to list"`
+	Site      string `json:"site" jsonschema:"description=The site of the entity to list"`
+	Namespace string `json:"namespace" jsonschema:"description=The namespace of the entity to list"`
```
sd2k: for each of these string types, is the model expected to know the possible options in advance? If it's a closed set (e.g. entity type) we should provide a list of options in the description really; for the others, perhaps we should tell the LLM how to obtain valid values either in the field description or tool description.

xujiaxj: We can supply a list of know types. This will help LLM.

tools/asserts.go
xujiaxj: Eventually the API expects Millis. Which way do you think will work better? Ask the LLM to come up with Millis or use RFC3399 format and convert to Millis in our code?

xujiaxj:
So I tested with the latest commits with the RFC3339 formatted string input, but keep running issues with the timestamp values.

```
Parameters
{
  "entityType": "Service",
  "entityName": "mongodb",
  "env": "asserts-demo",
  "site": "app",
  "namespace": "robot-shop",
  "startRfc3339": "2025-04-23T10:00:00Z",
  "endRfc3339": "2025-04-23T16:00:00Z"
}
```

Result
image

The result indicates the `startTime` and `endTime` are 0. I couldn't figure out what's wrong with the code. @sd2k I can use some help here.

sd2k:
@xujiaxj I saw a comment about getting zero values for the time struct which was strange so tried pushing a separate commit to your branch, hope that's OK. It might be unnecessary though? Let me know what you think.

xujiaxj:
> @xujiaxj I saw a comment about getting zero values for the time struct which was strange so tried pushing a separate commit to your branch, hope that's OK. It might be unnecessary though? Let me know what you think.

That happened when I was using Cursor with MCP. Somehow it didn't happen with Claude desktop, so it seems okay for me. But for sure, the new commit looks good to me, and I will test it out.

xujiaxj:
> > @xujiaxj I saw a comment about getting zero values for the time struct which was strange so tried pushing a separate commit to your branch, hope that's OK. It might be unnecessary though? Let me know what you think.
> 
> That happened when I was using Cursor with MCP. Somehow it didn't happen with Claude desktop, so it seems okay for me. But for sure, the new commit looks good to me, and I will test it out.

I think the reason it's flaky with Cursor is that I keep changing my MCP config to different Grafana instances. I should've restarted Cursor just like I restart Claude desktop whenever I change the MCP config.

xujiaxj:
Just fixed another bug while testing against `ops`. The scope attributes env/site/namespace is optional, so we shouldn't pass the empty string `""` to the API.

xujiaxj:
Latest test result in Cursor against `ops` seem fine.
image

sd2k (commented):
Code-wise this LGTM now. I am wondering if we could have some cloud tests running against a cloud instance which has asserts installed & running (rather than using the mcptests instance, which doesn't). Do you think that sounds doable?

xujiaxj:
Added another test and tested it locally, but need help to put up the two variables in the setting on [this page](https://github.com/grafana/mcp-grafana/settings/variables/actions) as I don't have access. cc: @csmarchbanks 

```
ASSERTS_GRAFANA_URL="https://dev.grafana-dev.net"
ASSERTS_GRAFANA_API_KEY=[the token for service account "asserts-mcp-cloud-test" on the instance]
```
Monosnap Image 2025-04-28 14-49-40

tools/asserts_cloud_test.go
```diff
@@ -0,0 +1,66 @@
...
+	"os"
+	"testing"
+	"time"
+
+	mcpgrafana "github.com/grafana/mcp-grafana"
+	"github.com/stretchr/testify/assert"
+	"github.com/stretchr/testify/require"
+)
+
+func createAssertsCloudTestContext(t *testing.T, testName string) context.Context {
```
ioanarm: This can be moved to `cloud_testing_utils.go` or you can even use the same function and parametrise the envs to grab like: 

this 
```ctx := createAssertsCloudTestContext(t, "Asserts")```
 can be replaced by
```
ctx := createCloudTestContextGeneric(t, "ASSERTS_GRAFANA_URL", "ASSERTS_GRAFANA_API_KEY", "Asserts")
```

And the `createCloudTestContext` func can get as arguments: 

```
func createCloudTestContextGeneric(t *testing.T, urlEnv, apiKeyEnv, testName string) context.Context {
	grafanaURL := os.Getenv(urlEnv)
	if grafanaURL == "" {
		t.Skipf("%s environment variable not set, skipping cloud %s integration tests", urlEnv, testName)
} ...
```

Comment (ioanarm) [reply tools/asserts_cloud_test.go]: We can do this on a separate PR though as it touches mutliple cloud test files.

File: tools/asserts_cloud_test.go [hunk 7]
```diff
@@ -0,0 +1,66 @@
+//go:build asserts
+// +build asserts
```
ioanarm:
```suggestion
//go:build cloud
// +build cloud
```

So this can be triggered by the `make test-cloud` as well

.github/workflows/go.yml
```diff
@@ -83,3 +83,22 @@ jobs:
...
+      - name: Set up Go
+        uses: actions/setup-go@v5
+        with:
+          go-version: "1.24"
+          cache: true
+
+      - name: Run asserts tests
+        env:
+          ASSERTS_GRAFANA_URL: ${{ vars.ASSERTS_GRAFANA_URL }}
+          ASSERTS_GRAFANA_API_KEY: ${{ secrets.ASSERTS_GRAFANA_API_KEY }}
```
ioanarm, line 86: You can add those two variables inside the test-cloud above so there is no specific workflow for asserts cloud testing. 
If we do this the the `make test-asserts` command is redundant

Makefile
```diff
@@ -35,6 +35,10 @@ test-integration: ## Run only the Docker-based integration tests (Requires docke
 test-cloud: ## Run only the cloud-based tests (requires cloud Grafana instance and credentials).
 	go test -v -tags cloud ./tools
 
+.PHONY: test-asserts
+test-asserts: ## Run only the asserts-based tests (requires Asserts Grafana instance and credentials).
+	go test -v -tags asserts ./tools -run TestAssertsCloudIntegration
+
```
ioanarm: Maybe redundant if we move the env variables inside the `test-cloud` git workflow.

tools/asserts_cloud_test.go
xujiaxj: Good idea! Let me refactor over.

xujiaxj:
Latest test run in local env:
Monosnap Image 2025-04-29 09-19-26

ioanarm (approved):
🚀

PR #114, opened 2025-05-05
docs: add note for vscode

FarDust:
VSCode doesn’t auto-discover custom SSE endpoints, so this config explicitly sets the MCP server URL. Prevents silent failures and confusion when the connection to /sse isn’t resolved automatically.


SandersAaronD (approved):
This looks good to me, thanks for the PR!

PR #115, opened 2025-05-06
Dashboards: Fetch panel queries

ioanarm:
Fetches panel queries along with their titles from a dashboard given a UID.


love01211:
Look forward to this function...

tools/dashboard.go
```diff
@@ -62,7 +63,74 @@ var UpdateDashboard = mcpgrafana.MustTool(
...
 )
 
+type DashboardPanelQueriesParams struct {
+	UID string `json:"uid" jsonschema:"required,description=The UID of the dashboard"`
+}
+
+type PanelQuery struct {
+	Title string `json:"title"`
+	Query string `json:"query"`
+}
```
sd2k: Should we include the datasource here too?

tools/dashboard.go
```diff
@@ -62,7 +63,74 @@ var UpdateDashboard = mcpgrafana.MustTool(
...
+			}
+		}
+	}
+
+	return result, nil
+}
+
+var GetDashboardPanelQueries = mcpgrafana.MustTool(
+	"get_dashboard_panel_queries",
+	"Returns a JSON object with a 'result' field containing an array of objects, each representing a panel in the specified dashboard. Each object includes the panel's title and its associated query string. This allows you to programmatically access all panel queries (title and query) for a dashboard.",
```
sd2k:
```suggestion
	"Get the title, datasource and query string of each panel in a dashboard. Returns a JSON object with a 'result' field containing an array of objects, each representing a panel in the specified dashboard.",
```

WDYT?

tools/dashboard.go
```diff
@@ -62,7 +63,74 @@ var UpdateDashboard = mcpgrafana.MustTool(
...
+	Title string `json:"title"`
+	Query string `json:"query"`
+}
+
+func GetDashboardPanelQueriesTool(ctx context.Context, args DashboardPanelQueriesParams) (struct {
+	Result []PanelQuery `json:"result"`
+}, error) {
+	result := struct {
+		Result []PanelQuery `json:"result"`
+	}{}
```
sd2k: Could we just return an array instead? Saves a few tokens.

sd2k (commented):
tests pass which is great :slightly_smiling_face: a few minor suggestions.

tools/dashboard.go
ioanarm: yeah I had it initially but most of the time it gives a placeholder variable ..
I can return empty if the `dashboardUID` though starts with `$`

ioanarm: I won't filter the `$`, it can easily happen on any agent side if needed.

sd2k: Ooh interesting so if it's a template variable. Perhaps we should make that clear in the tool description, so the model doesn't try to use it by accident.

tools/dashboard.go
```diff
@@ -62,7 +63,81 @@ var UpdateDashboard = mcpgrafana.MustTool(
...
 )
 
+type DashboardPanelQueriesParams struct {
+	UID string `json:"uid" jsonschema:"required,description=The UID of the dashboard"`
+}
+
+type PanelQuery struct {
+	Title         string `json:"title"`
+	Query         string `json:"query"`
+	DatasourceUID string `json:"datasourceUid"`
```
sd2k: I think we should also return the datasource type here if we know it.

ioanarm: good call, it's not available in all dashboards so I missed it 🙄

tools/dashboard.go
ioanarm: Have a look, I included some more info on the description.

tools/dashboard.go
```diff
@@ -62,7 +63,89 @@ var UpdateDashboard = mcpgrafana.MustTool(
...
+type DatasourceInfo struct {
+	UID  string `json:"uid"`
+	Type string `json:"type"`
+}
+
+type PanelQuery struct {
+	Title      string         `json:"title"`
+	Query      string         `json:"query"`
+	Datasource DatasourceInfo `json:"datasource"`
+}
```
sd2k: Cool looks good. Could you unexport these structs? :sweat_smile:

sd2k (commented):
LGTM! Could you add this tool to the table in the README? :pray:

sd2k (approved):
:shipit:

PR #122, opened 2025-05-13
Prevent panic if didn't find patterns.

gitdoluquita:
Details can be nil if, for example, no logs are found. This PR prevents this tool from panicking.


csmarchbanks (commented):
If this is an expected state when no logs are found should this just return nil instead of an error?

sd2k:
Yeah I don't think this should be returning an error, it's a successful response just with no results.

gitdoluquita:
Good call! I'm still a bit unsure on what we should error and what we should raise as an error to the LLM to pickup, but I think in this case the message is clear enough.

sd2k (approved):

csmarchbanks (approved):

PR #124, opened 2025-05-14
feat: adds admin category for listing teams

eleijonmarck:
Makes use of the openapi spec to query for teams. Eventually we would perhaps want all of the "admin" management inside of it's own `admin` functionality. 

This is a small step for adding admin functionality to the MCP server


cmd/mcp-grafana/main.go
```diff
@@ -34,7 +34,7 @@ type disabledTools struct {
 
 	search, datasource, incident,
 	prometheus, loki, alerting,
-	dashboard, oncall, asserts, sift bool
+	dashboard, oncall, asserts, sift, teams bool
```
sd2k:
```suggestion
	dashboard, oncall, asserts, sift, admin bool
```

IMO we should start with this being a larger 'admin' category for things like teams, permissions, roles. We can make it more granular later but 'teams' is quite vague.

tools/teams.go
sd2k, line 1: Similarly I would start with this being `admin.go` (and renaming `AddTeamsTools` to `AddAdminTools`, etc)

sd2k (commented):
thank you! a couple of comments on categories

cmd/mcp-grafana/main.go
eleijonmarck: renamed to admin. it just makes sense

sd2k:
Nice cheers. How hard would it be to add a test? There are several integration tests (tagged with `//go:build integration`) that could serve as inspiration but I guess we can't provision teams so may need to make some API calls to setup/teardown the team...

sd2k:
note: ideally these would even be e2e tests using Python - see https://github.com/grafana/mcp-grafana/blob/9077c3fe36d4b2d280839fc124446e77b3b078e4/tests/dashboards_test.py for example

tests/admin_test.py
```diff
@@ -0,0 +1,43 @@
...
+from utils import (
+    get_converted_tools,
+    llm_tool_call_sequence,
+)
+
+pytestmark = pytest.mark.anyio
+
+@pytest.mark.parametrize("model", models)
+@pytest.mark.flaky(max_runs=3)
+async def test_create_team_tool(model: str, mcp_client: ClientSession):
```
sd2k: Should this be `test_list_teams_tool`?

tests/admin_test.py
```diff
@@ -0,0 +1,43 @@
...
+
+    # 2. Final LLM response
+    response = await acompletion(model=model, messages=messages, tools=tools)
+    content = response.choices[0].message.content
+    panel_queries_checker = CustomLLMBooleanEvaluator(
+        settings=CustomLLMBooleanSettings(
+            prompt="Does the response contain specific information about the teams in Grafana?",
+        )
+    )
+    print("content", content)
```
sd2k:
```suggestion
```

sd2k (commented):
Looks like tests are failing because we don't have any teams atm. I'll add one and rerun...

tools/admin.go
```diff
@@ -0,0 +1,39 @@
...
+
+	"github.com/mark3labs/mcp-go/server"
+
+	"github.com/grafana/grafana-openapi-client-go/client/teams"
+	"github.com/grafana/grafana-openapi-client-go/models"
+	mcpgrafana "github.com/grafana/mcp-grafana"
+)
+
+type ListTeamsParams struct {
+	Query string `json:"query" jsonschema:"description=The query to search for teams"`
```
sd2k:
```suggestion
	Query string `json:"query" jsonschema:"description=The query to search for teams. Can be left empty to fetch all teams."`
```

sd2k: I wonder if this will help the failing test?

eleijonmarck: ill test it out

eleijonmarck: i ran the tests locally and it works

eleijonmarck: > Huh, there is already a team in our test instance so I'm not sure why this is failing, left a suggestion which could help though

which instance is it testing?

sd2k: https://mcptests.grafana-dev.net

tests/admin_test.py
```diff
@@ -0,0 +1,108 @@
...
+@pytest.mark.parametrize("model", models)
+@pytest.mark.flaky(max_runs=3)
+async def test_list_teams_tool(
+    model: str, mcp_client: ClientSession, grafana_team
+):
+    tools = await get_converted_tools(mcp_client)
+    team_name = grafana_team["name"]  # Get the name of the created team
+    prompt = (
+        f"Can you list teams in Grafana? "
+        f"There should be a team named {team_name}."
```
sd2k: I think this bit should be in the judge, not the initial prompt

sd2k (approved):
LGTM, thanks for your patience!

eleijonmarck:
> LGTM, thanks for your patience!

Thx for your humble guidance!

PR #127, opened 2025-05-16
refactor: extract Grafana Client creation from context functions

sd2k:
There was a bunch of duplicated code when creating the Grafana client from
context, and I want to be able to do this in tests too, so this just refactors
the code for later.


annanay25 (approved):
lgtm

Issue #123, opened 2025-05-14
Error while running sift error logs pattern tool


kozhuhds:
Tested on `assertschief`

Error message:
```
MCP error -32603: creating investigation: failed to unmarshal response body: json: cannot unmarshal array into Go struct field .data of type tools.Investigation. body: ...
```

annanay25:
![Image](https://github.com/user-attachments/assets/b5b62132-9e1a-4916-b76c-1dd1bcfb9ffd)

![Image](https://github.com/user-attachments/assets/5a0d504f-cb69-4b03-8807-ccaae9fd4dbf)

![Image](https://github.com/user-attachments/assets/5827083a-8198-4ce4-96a9-5fb1bb1a6722)

![Image](https://github.com/user-attachments/assets/5d281033-374b-48b0-bde7-a69c99cd4e7e)

SandersAaronD:
@kozhuhds Is this still an issue for workbench? I am not sure that the error is strictly in MCP, it doesn't seem to reproduce here. We have newly added tracing to mcp server for this if it comes up, but we can close this as stale if a fix would not currently impact ongoing work.

SandersAaronD:
Per feedback this is no longer relevant so we can close it

PR #128, opened 2025-05-16
test: add test case for finding log error patterns

sd2k:
This adds a cloud integration test for the functions underlying the error pattern tool.

The commits are pretty atomic so it might be worth checking them out independently, but it's basically:

- add some logs when creating an investigation (note: we don't log any sensitive stuff, just investigation request/response and some IDs really)
- include a Grafana client in the test context, required to get Loki datasource IDs so we can fetch some examples
- when testing analysis stuff, rather than fetching just 1 investigation, fetch 10 in case the first has no analyses (which can happen if the investigation refused to start due to lack of matching log streams etc)
- add the actual test.

I had hoped this would find the cause of #123 but so far no dice :thinking:

### DIFF
--- a/tools/cloud_testing_utils.go
+++ b/tools/cloud_testing_utils.go
@@ -11,11 +11,14 @@ import (
 	mcpgrafana "github.com/grafana/mcp-grafana"
 )
 
-// createCloudTestContext creates a context with Grafana URL and API key for cloud integration tests.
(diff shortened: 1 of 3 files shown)


csmarchbanks (approved):

annanay25 (approved):
lgtm 👍

PR #132, opened 2025-05-16
feat: add more annotations to tools

sd2k:
Add the read-only, idempotent, and destructive hint annotations to tools
where appropriate.

Co-authored-by: Luccas Quadros <luccas.quadros@grafana.com>

### DIFF
--- a/tools/sift.go
+++ b/tools/sift.go
@@ -178,6 +178,8 @@ var GetSiftInvestigation = mcpgrafana.MustTool(
 	"Retrieves an existing Sift investigation by its UUID. The ID should be provided as a string in UUID format (e.g. '02adab7c-bf5b-45f2-9459-d71a2c29e11b').",
 	getSiftInvestigation,
 	mcp.WithTitleAnnotation("Get Sift investigation"),
+	mcp.WithIdempotentHintAnnotation(true),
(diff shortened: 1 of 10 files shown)


tools/sift.go
```diff
@@ -333,6 +339,7 @@ var FindErrorPatternLogs = mcpgrafana.MustTool(
 	"Searches Loki logs for elevated error patterns compared to the last day's average, waits for the analysis to complete, and returns the results including any patterns found.",
 	findErrorPatternLogs,
 	mcp.WithTitleAnnotation("Find error patterns in logs"),
+	mcp.WithReadOnlyHintAnnotation(true),
```
annanay25, line 345: do we need to add "false" on the Idempotent hint annotation? Not sure if that really affects performance

annanay25 (approved):
lgtm overall, left a minor comment

tools/sift.go
sd2k: Yeah good point actually, I'm not sure if the default is false or just 'assume nothing'...

sd2k: The [docs](https://modelcontextprotocol.io/docs/concepts/tools#available-tool-annotations) indicate that the default is `false` so I don't think we need to add the idempotent one.

Issue #134, opened 2025-05-16
Remove Asserts tools


annanay25:
We are migrating asserts tools into the Asserts plugin as we ran into a few auth related issues, and they're also easier to maintain there. This issue is to track the removal of `getAssertions` tool from this server.

annanay25:
Ben fixed the auth issue :)

xujiaxj:
All right. Let's keep it for now.

PR #135, opened 2025-05-19
fix: pass OBO auth to asserts client, if available

sd2k:
This was just missed off the original OBO PR.

Should fix the auth issues mentioned in #134.


annanay25 (approved):
lgtm!

Issue #121, opened 2025-05-13
Empty list returned on prometheus queries


leandregagnonlewis:
Hi! I am experimenting with the mcp and I observed that the `prometheus_query` tool is not working. Requests always return empty arrays even if the same queries return lot of values when I do them by myself.

Here is the output in `grafana-llm-app` plugin:
```
Tool Calls
list_datasources
{"type":"prometheus"}

Response
query_prometheus
{"datasourceUid":"eelrn8py2skcgb","expr":"up{container=\"alertmanager\"}","startRfc3339":"2023-10-01T00:00:00Z","queryType":"instant"}

Response
{
  "content": [
    {
      "type": "text",
      "text": "[]"
    }
  ]
}
```
and here is the output in grafana explore:

<img width="1485" alt="Image" src="https://github.com/user-attachments/assets/3f66ca54-d8ec-4be3-8347-20adb1ac9820" />

I have also experimented with the `mcp` with an other agent and I had the same problem.

sd2k:
Huh, this is odd. I wonder if it's to do with the very old timestamp used in `startRfc3339`? Is your Grafana Explore time range set to the same thing (October 2023)?

If that is it, I wonder if we should make the `startRfc3339` parameter optional and default it to `now` in the case that the agent doesn't actually know the current time 🤔

leandregagnonlewis:
I still get values when setting the time range start to october 2023. I don't think this parameter should have an impact on instant queries.

![Image](https://github.com/user-attachments/assets/00a40996-1dcf-4451-be9d-f35048c9b9e3)

sd2k:
I think Grafana's explore view uses the _end_ of the range as the query timestamp in instant queries. If you set it to 2023-10-10 00:00:00 to 2023-10-10 01:00:00 you will probably get no results.

My hunch is that your agent didn't have access to the current time and simply used its best guess (2023-10-10 00:00:00) in the `startRfc3339` parameter, since I can't see where else that value would come from. You could test by asking the agent to run the query at a more recent timestamp (like 2025-05-15 00:00:00).

One solution could be to give your agent access to the current time, maybe using another MCP server that adds a tool for that. In our server, we _could_ make that tool default to `now`, but I'm not sure it's the best idea since it could return undesirable results in some cases. I'll think about it some more.

leandregagnonlewis:
You are right, it works when setting the time manually . I need to find how to share my system time now

failedanswer:
[...]

PR #142, opened 2025-05-22
Add support for relative time specifications in Prometheus queries

leandregagnonlewis:
# Add support for relative time specifications in Prometheus queries

## Description
This PR adds support for relative time specifications in Prometheus queries by introducing `StartRelative` and `EndRelative` fields to the `QueryPrometheusParams` struct. This enhancement allows users to specify time ranges using human-readable formats like "now" and "now-1h" instead of having to manually format RFC3339 timestamps.

## Changes
- Added `StartRelative` and `EndRelative` string fields to `QueryPrometheusParams` struct
- Implemented `parseRelativeTime` helper function to convert relative time expressions to absolute timestamps
- Updated `queryPrometheus` function to handle both RFC3339 and relative time formats
- Added comprehensive unit tests for the new functionality
- Updated integration tests with better timestamp comparison logic
- Updated documentation to reflect the new functionality

## Implementation Details
The implementation supports two formats:
1. `now` - representing the current time
2. `now-<duration>` - representing a time in the past relative to now (e.g., "now-1h" for one hour ago)

The duration format follows Go's standard duration format (e.g., "1h", "30m", "1h30m", etc.).

## Testing
- Added unit tests for the `parseRelativeTime` function
- Updated integration tests for both instant and range queries with relative timestamps
- Improved test assertions to handle timestamp comparison with appropriate tolerances


tools/prometheus.go
```diff
@@ -103,12 +103,42 @@ var ListPrometheusMetricMetadata = mcpgrafana.MustTool(
...
-	EndRFC3339    string `json:"endRfc3339,omitempty" jsonschema:"description=The end time in RFC3339 format. Required if queryType is 'range'\\, ignored if queryType is 'instant'"`
-	StepSeconds   int    `json:"stepSeconds,omitempty" jsonschema:"description=The time series step size in seconds. Required if queryType is 'range'\\, ignored if queryType is 'instant'"`
-	QueryType     string `json:"queryType,omitempty" jsonschema:"description=The type of query to use. Either 'range' or 'instant'"`
+	DatasourceUID  string `json:"datasourceUid" jsonschema:"required,description=The UID of the datasource to query"`
+	Expr           string `json:"expr" jsonschema:"required,description=The PromQL expression to query"`
+	StartRFC3339   string `json:"startRfc3339" jsonschema:"description=The start time in RFC3339 format"`
+	EndRFC3339     string `json:"endRfc3339,omitempty" jsonschema:"description=The end time in RFC3339 format. Required if queryType is 'range'\\, ignored if queryType is 'instant'"`
+	StartRelative  string `json:"startRelative,omitempty" jsonschema:"description=The start time relative to now (e.g. 'now-1h'\\, 'now-1d'). Alternative to startRfc3339"`
```
ioanarm: I think that go’s time.ParseDuration does not support "d" or "w" as units—only hours ("h"), minutes ("m") etc. I think that some processing is required if we want to support those as well. Otherwise we can add this limitation to the tool description so "1d" can be converted to "24h" for example.

ioanarm (commented):
Hello @leandregagnonlewis! Thanks so much for adding this, really useful :) 
I started playing with it and I got an error, it seems that `ParseDuration` doesn't support days as relative time 😞

ioanarm:
Oh actually according to @sd2k comment [here](https://github.com/grafana/mcp-grafana/issues/121#issuecomment-2890801882) `gtime` package will take care of it :). Let me know if there is anything I can help with!

tools/prometheus.go
ioanarm: Oh actually according to @sd2k comment [here](https://github.com/grafana/mcp-grafana/issues/121#issuecomment-2890801882) `gtime` package will take care of it :). Let me know if there is anything I can help with!

leandregagnonlewis: Thanks for pointing that out. I went a bit too fast 😅 .  I pushed a commit to make use of `gtime` and support "d" units. It still does not support weeks though. This probably would require a custom parsing function.

tools/prometheus.go
```diff
@@ -122,19 +153,42 @@ func queryPrometheus(ctx context.Context, args QueryPrometheusParams) (model.Val
...
+			if err != nil {
+				return nil, fmt.Errorf("parsing relative end time: %w", err)
+			}
+		} else if args.EndRFC3339 != "" {
+			endTime, err = time.Parse(time.RFC3339, args.EndRFC3339)
+			if err != nil {
+				return nil, fmt.Errorf("parsing RFC3339 end time: %w", err)
+			}
+		} else {
+			return nil, fmt.Errorf("either endRfc3339 or endRelative must be provided when queryType is 'range'")
```
ioanarm: There is some code duplication here and in lines 156–169 on time parsing. Should we extract this to a small helper function? `getTime` or something?

ioanarm (commented):
Nice, one small comment!

tools/prometheus.go
leandregagnonlewis: Sure. i will also try to fix the tests!

leandregagnonlewis: Refactoring is done! But for the failing tests, it looks like an API key is missing so I can't fix it by myself

ioanarm: The workflow has been updated to skip cloud and e2e tests on fork PRs! If you try and rebase now this should be fixed 🤞 !

leandregagnonlewis: ✅

sd2k (commented):
Hey @leandregagnonlewis, thanks for the PR!

My hunch would be that we should switch to a single of each `start` and `end` parameter and accept either RFC3339 _or_ relative times (`now` etc) in that single param, rather than having multiple. That would match the way Grafana's start/end times work in the UI and should be fairly easy to handle. Do you think that would work, or can you see any downsides to that approach?

leandregagnonlewis:
> Hey @leandregagnonlewis, thanks for the PR!
> 
> My hunch would be that we should switch to a single of each `start` and `end` parameter and accept either RFC3339 _or_ relative times (`now` etc) in that single param, rather than having multiple. That would match the way Grafana's start/end times work in the UI and should be fairly easy to handle. Do you think that would work, or can you see any downsides to that approach?

I think you are right. I initially did not want to break the existing tool interface, but I guess it should not matter with an MCP server. I will work on this

tools/prometheus.go
```diff
@@ -7,6 +7,7 @@ import (
 	"strings"
 	"time"
 
+	"github.com/gogf/gf/v2/os/gtime"
```
sd2k: This isn't the correct gtime package :thinking: I was expecting the [Grafana plugin SDK's gtime package](https://pkg.go.dev/github.com/grafana/grafana-plugin-sdk-go@v0.277.1/backend/gtime#NewTimeRange). We should be able to use the `NewTimeRange` function and avoid any need for custom parsing ourselves.

sd2k (commented):
Thanks for adding the tests :+1: added a comment inline on the gtime package though.

sd2k (approved):
This is great! Thanks so much @leandregagnonlewis :muscle:

sd2k:
We should be able to reuse this in other queries (e.g. Loki) but we can cross that bridge later.

PR #152, opened 2025-05-28
docs: improve usage examples and docs on stdio vs SSE defaults

### LINKED ISSUE #149: ambiguous docker example in readme
Here is the docker mcp example in the README.md - https://github.com/grafana/mcp-grafana/blob/v0.4.1/README.md?plain=1#L212-L237

```json
{
  "mcpServers": {
    "grafana": {
      "command": "docker",
      "args": [
        "run",
        "--rm",
        "-p",
        "8000:8000",
        "-e",
        "GRAFANA_URL",
        "-e",
        "GRAFANA_API_KEY",
        "mcp/grafana",
        "-debug"
      ],
      "env": {
        "GRAFANA_URL": "http://localhost:3000",
        "GRAFANA_API_KEY": "<your service account token>"
      }
    }
  }
}
```

It appears that it launches an stdio server, with the command and args, but in reality it initiates a SSE server.

The DX of the `mcp/grafana` container image is also a little bit confusing:

When I run `docker run -it --rm mcp/grafana --help` It shows

```bash
  -t string
        Transport type (stdio or sse) (default "stdio")
  -transport string
        Transport type (stdio or sse) (default "stdio")
```

which leaves me under the impression that it by default starts a stdio server, however in reality it starts a SSE server by default via its entrypoint - https://github.com/grafana/mcp-grafana/blob/v0.4.1/Dockerfile#L41

sd2k:
Our Docker image runs the server using SSE by default and it would be
a backwards incompatible change to switch it to STDIO (which would
cause confusion amongst users). This commit instead improves usage
examples by explaining this, and by updating the Docker examples to
use STDIO instead, since it's what's used by most other MCP Docker
images that I could find.

Fixes #149.

### DIFF
--- a/README.md
+++ b/README.md
@@ -107,11 +107,22 @@ the OnCall tools, use `--disable-oncall`.
 
 2. You have several options to install `mcp-grafana`:
 
-   - **Docker image**: Use the pre-built Docker image from Docker Hub:
+   - **Docker image**: Use the pre-built Docker image from Docker Hub.
+
+     **Important**: The Docker image's entrypoint is configured to run the MCP server in SSE mode by default, but most users will want to use STDIO mode for direct integration with AI assistants like Claude Desktop:
+
+     1. **STDIO Mode**: For stdio mode you must explicitly override the default with `-t stdio` and include the `-i` flag to keep stdin open:
 
      ```bash
      docker pull mcp/grafana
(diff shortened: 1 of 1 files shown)


sd2k:
I did wonder this too, but it would be a really confusing experience for users, especially since it would just happen invisibly overnight (or whenever they ended up pulling a new version of the image)... I figured we can punt that decision for now and potentially change our minds later.

PR #154, opened 2025-06-02
ci(e2e): build/run MCP server separately, and print logs on failure

sd2k:
Before this it was hard to tell what was going wrong when the e2e tests failed because the MCP server logs are piped to a file which is never printed.

This commit builds the MCP server before running it with nohup, then prints the logs if the tests fail.


ioanarm (approved):
❤️

PR #165, opened 2025-06-07
Add read/write mode definition to tools

edwardcqian:
Currently there's no good way to avoid using tools that can persist changes.

Adding the notion of mode to separate the "read" tools from the "write" tools.

Notes for reviewers:
- Not sure if read or write is the best terms here, open to suggestions
- Sift tools are a bit tricky, since error pattern and slow request technically creates an investigation


edwardcqian:
Ah, just realized there's the readOnly annotation 🤦

PR #171, opened 2025-06-11
added streamable-http support

ambhojgoyal:
### Summary
Introduces streamable HTTP mode for the MCP server, enabling real-time streaming of responses over HTTP connections.

### Compatibility
Fully backward compatible - existing non-streaming clients continue to work unchanged.

### DIFF
--- a/cmd/mcp-grafana/main.go
+++ b/cmd/mcp-grafana/main.go
@@ -106,6 +106,15 @@ func run(transport, addr, basePath string, logLevel slog.Level, dt disabledTools
 		if err := srv.Start(addr); err != nil {
 			return fmt.Errorf("Server error: %v", err)
 		}
+	case "streamable-http":
+		srv := server.NewStreamableHTTPServer(s, server.WithHTTPContextFunc(mcpgrafana.ComposedHTTPContextFunc(gc.debug)),
+			server.WithStateLess(true),
+			server.WithEndpointPath(basePath),
+		)
+		slog.Info("Starting Grafana MCP server using StreamableHTTP transport", "address", addr, "basePath", basePath)
+		if err := srv.Start(addr); err != nil {
+			return fmt.Errorf("Server error: %v", err)
+		}
 	default:
 		return fmt.Errorf(
 			"Invalid transport type: %s. Must be 'stdio' or 'sse'",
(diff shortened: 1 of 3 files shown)


cmd/mcp-grafana/main.go
```diff
@@ -124,15 +133,22 @@ func main() {
 		"stdio",
 		"Transport type (stdio or sse)",
 	)
-	addr := flag.String("sse-address", "localhost:8000", "The host and port to start the sse server on")
-	basePath := flag.String("base-path", "", "Base path for the sse server")
+	addr := flag.String("address", "localhost:8000", "The host and port to start the sse server on")
+	basePath := flag.String("base-path", "", "Base path for the sse or streamable-http server")
```
sd2k: The base path of the SSE server is a bit different from the endpoint path of the streamable HTTP server so I think we should split this flag up into `base-path` (for SSE) and `endpoint-path` (for streamable HTTP). We can provide better defaults for each, then.

ambhojgoyal: Hey @sd2k  ,
Thanks for the the input,
Have added the changes

cmd/mcp-grafana/main.go
```diff
@@ -124,16 +133,17 @@ func main() {
 		"stdio",
 		"Transport type (stdio or sse)",
 	)
-	addr := flag.String("sse-address", "localhost:8000", "The host and port to start the sse server on")
-	basePath := flag.String("base-path", "", "Base path for the sse server")
+	addr := flag.String("address", "localhost:8000", "The host and port to start the sse server on")
+	basePath := flag.String("base-path", "", "Base path for the sse")
```
sd2k:
```suggestion
	basePath := flag.String("base-path", "", "Base path for the sse server")
```

sd2k (commented):
Thanks!

ambhojgoyal:
> Hi @ambhojgoyal, thanks for the contribution, it looks decent. I added an inline comment on the `base-path` flag, but otherwise the only thing I'd like are some end-to-end tests for this (see the 'tests' directory) and it might be tricky for you to iterate (since they don't run in CI on forks 😱). Let me know if you're up to adding them, if not I can take a look.

Hey @sd2k 
Was not able to get much hang of things here, will appreciate if you can help me out here

cmd/mcp-grafana/main.go
ambhojgoyal: @sd2k Done, updated

sd2k (approved):
Nice! I'll add some e2e tests in a followup PR to avoid polluting this one. Thanks again :+1:

PR #190, opened 2025-06-24
Local config implementation

jpinsonneau:
This is a WIP implementing extra configs, allowing more deployments possibilities such as local or k8s.

It rely on extra environment variables to configure endpoints and TLS + forward Authorization header when provided.

- [X] Prometheus implementation
- [X] Loki implementation

Stdio run using env variables:
mcp
image

Prometheus query forwarding auth token:
image

Loki query:
image

### DIFF
--- a/examples/k8s/k8s-mcp.yaml
+++ b/examples/k8s/k8s-mcp.yaml
@@ -1,0 +1,42 @@
+kind: Pod
+apiVersion: v1
+metadata:
+  name: k8s-mcp
+  namespace: my-app
+  labels:
+    app: k8s-mcp
+spec:
+  restartPolicy: Always
+  containers:
+    - name: mcp-grafana
+      image: 'quay.io/grafana/mcp-grafana:dev'
+      command: ["/app/mcp-grafana"]
+      args:
+        - '-transport=streamable-http'
+        - '-address=localhost:8000'
+        - '-endpoint-path=/mcp'
+        - '-use-grafana-config=false'
+        - '-enabled-tools=prometheus,loki'
+        - '-log-level=debug'
+      env:
+        - name: LOKI_URL
+          value: http://loki.my-app.svc:3100
+        - name: PROM_URL
+          value: http://prom.my-app.svc:9091
+---
+kind: Service
+apiVersion: v1
+metadata:
+  name: k8s-mcp
+  namespace: my-app
+  labels:
+    app: k8s-mcp
+spec:
+  ports:
+    - protocol: TCP
+      port: 8000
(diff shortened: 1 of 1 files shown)


jpinsonneau:
> [CLA assistant check](https://cla-assistant.io/grafana/mcp-grafana?pullRequest=190) Thank you for your submission! We really appreciate it. Like many open source projects, we ask that you sign our [Contributor License Agreement](https://cla-assistant.io/grafana/mcp-grafana?pullRequest=190) before we can accept your contribution.You have signed the CLA already but the status is still pending? Let us [recheck](https://cla-assistant.io/check/grafana/mcp-grafana?pullRequest=190) it.

Getting a blank box on that :angel: 

Edit: here we go, it loaded after some minutes

sd2k:
Hi @jpinsonneau, thanks for the contribution.

I'm a little confused about this PR - it seems like it results in bypassing Grafana entirely, which makes me think that it doesn't really belong in this MCP server, but instead in separate Prometheus/Loki MCP servers. I would really rather avoid adding more configuration complexity into this MCP server (which already has quite a few config options).

jpinsonneau:
> Hi @jpinsonneau, thanks for the contribution.
> 
> I'm a little confused about this PR - it seems like it results in bypassing Grafana entirely, which makes me think that it doesn't really belong in this MCP server, but instead in separate Prometheus/Loki MCP servers. I would really rather avoid adding more configuration complexity into this MCP server (which already has quite a few config options).

Hi @sd2k, thanks for your feedback.

Indeed that could be in a smaller MCP server covering only Prom & Loki which could be even imported here as a dependency to avoid code duplication.
This work is part of investigations around AI connectors. Red Hat started a workgroup cross orgs about that: 
https://docs.google.com/document/d/13OJvWKGKZL0V4nPpv9anitL5je92FCdQyXePcrY2N84
https://groups.google.com/g/kubernetes-sig-architecture/c/j_K2VvUhBWY/m/8KCimY5zCwAJ
https://github.com/cncf/tag-runtime/issues/212

Feel free to join or ping people who could be interested :smiley: 
Thanks

SandersAaronD:
MCP servers don't seem to have a set pattern now, and we're still somewhat grappling with whether the datasource should have the mcp server on it, or whether we would want to forward mcp servers from elsewhere via grafana. In this specific case it's not clear to me that this is specific to grafana, and it looks like it would make more sense to either run it side-by-side or have it forwarded to grafana

jpinsonneau:
Closing this in favor of https://github.com/jpinsonneau/mcp-shield

Issue #192, opened 2025-06-27
Enable ADO authentication


KnicKnic:
I use Azure Managed Grafana and Azure Monitor workspace (promql endpoint). 
I would like to use the MCP and have authentication be transparent in a system that has access to AAD auth.

What this would entail. 
1. Establishing at startup the aad credential object
2. grabbing a token any time a call is about to be made
    1. recreating the http client associated with the token if the token is different from the previous time this was done
        1.  Not sure the cleanest way to add this in the existing code
    1. This needs to done for both promql & grafana
4. updating the code to not proxy promql calls through grafana and send them direct
5. Adding some new type of config that would specify to do this

Willing to make a PR if it will be accepted.

sigurdfalk:
Probably need this one https://github.com/grafana/mcp-grafana/issues/192 or similar before we can do this

PR #193, opened 2025-06-27
Add ability to authenticate to grafana using AAD auth.

KnicKnic:
Addresses https://github.com/grafana/mcp-grafana/issues/192

1. Adds USE_AAD_AUTH to use aad for grafana.
1. Adds to GrafanaConfig a AAD default credential if USE_AAD_AUTH is set.
2. Changed storage of GrafanaClient to a functor to allow computation of the grafana client
    1. This seemed like the cleanest way to update the code  
    1. Updated GrafanaClient functor to allow returning errors and fixed up all call sites to deal with errors.
1. updated the promClientFromContext to assume it is goign to a Azure Monitor workspace and login again with AAD auth
    1. I didn't add logic to cache this client as it isn't being cached today, but it is a potential source of improvement.


tools/prometheus.go
```diff
@@ -31,14 +36,26 @@ var (
...
+	if cfg.AADCredential != nil {
+		url = datasource.URL
+		cred, err := cfg.AADCredential.GetToken(ctx, policy.TokenRequestOptions{
+			Scopes: []string{defaultPrometheusAADResource},
+		})
+		if err != nil {
+			return nil, fmt.Errorf("getting AAD token for Prometheus: %w", err)
+		}
+		cfg.APIKey = cred.Token
+	}
```
sd2k: I'm not sure it makes sense to do this for _all_ Prometheus data sources - there could be multiple configured, some which are pointing to Azure Monitor instances and some which aren't. Is there a way to tell from the datasource settings whether it is or not?

KnicKnic:
@sd2k  @csmarchbanks Can you approve running the pending checks? It would be useful to know if more work is required for the PR. I do not have planned work other than responding to comments.

tools/prometheus.go
sd2k: E.g. perhaps if the URL hostname ends with `prometheus.monitor.azure.com`?

sd2k (commented):
Hi! Thanks for the contribution. The code mostly LGTM, I'm just a little unsure about merging this right now for a few reasons:

1. Neither I nor any of the other maintainers use Azure so it's tricky for us to test this or fix any future issues
2. We're hoping to (eventually) have support for OAuth inside Grafana, although we have no timelines here so it could be a while

It also seems slightly infeasible to add different auth mechanisms for every different way of running Grafana (something we're hoping to circumvent with OAuth support).

With that said, it does look like a fairly straightforward addition to the code, so we could perhaps get this in with the caveat that it's experimental.

I'll look into whether we can get some simple end to end smoke tests using an Azure Managed Grafana instance so we have a bit more confidence.

tools/prometheus.go
KnicKnic: I added a check for the azureAuth and then go directly.
https://github.com/grafana/mcp-grafana/blob/539047c3fd6c9e4e16f2aba6125dd2d3ab8a26e9/tools/prometheus.go#L49-L60

KnicKnic:
> Hi! Thanks for the contribution. The code mostly LGTM, I'm just a little unsure about merging this right now for a few reasons:
> 
> 1. Neither I nor any of the other maintainers use Azure so it's tricky for us to test this or fix any future issues
> 2. We're hoping to (eventually) have support for OAuth inside Grafana, although we have no timelines here so it could be a while
> 
> It also seems slightly infeasible to add different auth mechanisms for every different way of running Grafana (something we're hoping to circumvent with OAuth support).
> 
> With that said, it does look like a fairly straightforward addition to the code, so we could perhaps get this in with the caveat that it's experimental.
> 
> I'll look into whether we can get some simple end to end smoke tests using an Azure Managed Grafana instance so we have a bit more confidence.

I understand about adding smoke tests. This feature currently is explicitly opt in. But I understand not breaking it.
Also the oauth flow may want some similar mechanisms that I added around returning failures.

 I will state that using AAD allows various auth flows that start with the logged in context of the user on the machine or service. This has been really nice when using to just have people auto logged in.

david-wagih:
@KnicKnic 

does this PR impact the issue when trying to run the MCP server on my graffana instance which has active directory as a sign in way, as it for now returns errors from the MCP server, but when tested on local graffana instance without the active directory setup it worked properly

ioanarm:
Hello @KnicKnic ! I am closing this one as it's stale for more than 6 months now.

KnicKnic:
BTW for those that find this thread, I now use azure's built in one - see https://learn.microsoft.com/en-us/azure/managed-grafana/grafana-mcp-server

PR #198, opened 2025-07-02
Hackathon-13: add git sync capabilities

itsmylife:
Complimentary PR https://github.com/grafana/grafana/pull/107447

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,7 +9,12 @@ wheels/
 .vscode/
 .env
 .cursor/
+.idea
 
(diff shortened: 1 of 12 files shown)


SandersAaronD:
I kind of love this as an idea but am not extremely sure if we'd want to put it into the self-same mcp server as a matter of design, do we have further docs on this one?

Regardless: For now I'm going to close it and open an issue linking to it since I _think_ the intention is mostly to use this as reference.

PR #206, opened 2025-07-17
feat: Add otel spans to tool calls

SandersAaronD:
- Adds trace spans for o11y into the call chain
- Tool arguments are not logged by default to prevent accidental PII exposure, although this seems like it may be excessively cautious and might hinder debugging especially
- We should be able to simply set this to log tool calls when we want to
- Also fixes test environment isolation issues in `TestExtractGrafanaInfoFromHeaders` that was causing failures when `GRAFANA_API_KEY` environment variables were present.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,10 +20,9 @@ require (
 	github.com/prometheus/common v0.65.0
 	github.com/prometheus/prometheus v0.305.0
 	github.com/stretchr/testify v1.10.0
-	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.60.0
-	go.opentelemetry.io/otel v1.35.0
-	go.opentelemetry.io/otel/sdk v1.35.0
-	go.opentelemetry.io/otel/trace v1.35.0
+	go.opentelemetry.io/contrib/instrumentation/net/http/otelhttp v0.61.0
+	go.opentelemetry.io/otel v1.36.0
+	go.opentelemetry.io/otel/sdk v1.36.0
 )
 
 require (
@@ -122,11 +121,9 @@ require (
 	go.opentelemetry.io/contrib/instrumentation/net/http/httptrace/otelhttptrace v0.61.0 // indirect
 	go.opentelemetry.io/contrib/propagators/jaeger v1.36.0 // indirect
 	go.opentelemetry.io/contrib/samplers/jaegerremote v0.30.0 // indirect
-	go.opentelemetry.io/otel v1.36.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace v1.36.0 // indirect
 	go.opentelemetry.io/otel/exporters/otlp/otlptrace/otlptracegrpc v1.36.0 // indirect
 	go.opentelemetry.io/otel/metric v1.36.0 // indirect
-	go.opentelemetry.io/otel/sdk v1.36.0 // indirect
 	go.opentelemetry.io/otel/trace v1.36.0 // indirect
 	go.opentelemetry.io/proto/otlp v1.6.0 // indirect
 	golang.org/x/exp v0.0.0-20250106191152-7588d65b2ba8 // indirect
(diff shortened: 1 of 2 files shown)


mcpgrafana.go
```diff
@@ -56,6 +58,17 @@ type GrafanaConfig struct {
...
+	// EnableTracing enables OpenTelemetry instrumentation for both tool execution
+	// and HTTP requests, providing complete observability with perfect correlation
+	// between tool spans and HTTP request spans. Defaults to false for minimal overhead.
+	EnableTracing bool
+
+	// LogTraceArguments enables logging of tool arguments in OpenTelemetry spans
+	// when tracing is enabled. This should only be enabled in non-production 
+	// environments or when you're certain the arguments don't contain PII. 
+	// Defaults to false for safety. Requires EnableTracing to be true.
+	LogTraceArguments bool
```
annanay25: is this really helpful somewhere? can we debug log arguments passed to the tools before calling mcp instead?

annanay25 (commented):
agree overall to more instrumentation, just concerned about the `LogTraceArguments` option.

mcpgrafana.go
SandersAaronD: We can either log for that separately in the LLM app (which we don't right now) and in anything else we build with the mcp server, or we can log it here. This seems tidier

tools.go
```diff
@@ -84,31 +88,90 @@ func ConvertTool[T any, R any](name, description string, toolHandler ToolHandler
 	}
 
 	handler := func(ctx context.Context, request mcp.CallToolRequest) (*mcp.CallToolResult, error) {
+		// Create OpenTelemetry span for tool execution (safely, with fallback)
+		var span trace.Span
+		
+		// Only create spans if tracing is explicitly enabled
```
sd2k: I wonder if these checks are actually needed 🤔 we should be able to create spans anyway, and they just won't be propagated anywhere if tracing is disabled? it would make this code a lot cleaner.

mcpgrafana.go
sd2k: Is 'Log' the correct verb here? At first I thought this referred to regular logging, not including the args in span attributes, so it's a bit confusing.

README.md
```diff
@@ -211,6 +211,128 @@ If you're using VSCode and running the MCP server in SSE mode (which is the defa
...
+
+```
+mcp.tool.query_prometheus
+├── attributes: {
+│   ├── mcp.tool.name: "query_prometheus"
+│   ├── mcp.tool.description: "Query Prometheus using PromQL..."
+│   └── mcp.tool.arguments: "{\"expr\":\"up\",\"datasourceUID\":\"abc123\"}"
+│   }
+```
+
```
sd2k: This shouldn't be in the main README IMO since it only applies to people using the MCP server programmatically, which is surely a small minority of users.

mcpgrafana.go
```diff
@@ -56,6 +58,17 @@ type GrafanaConfig struct {
 	// Debug enables debug mode for the Grafana client.
 	Debug bool
 
+	// EnableTracing enables OpenTelemetry instrumentation for both tool execution
+	// and HTTP requests, providing complete observability with perfect correlation
+	// between tool spans and HTTP request spans. Defaults to false for minimal overhead.
+	EnableTracing bool
```
sd2k: I wonder if we should just get rid of this flag and always add the trace context propagation by modifying the Grafana client transport. Is there a reason we don't want to do that? 🤔

sd2k (commented):
This does look helpful, I think we should unconditionally create spans and propagate the context though?

Perhaps we should use the flag to determine whether to instantiate an OTel exporter in our CLI?

mcpgrafana.go
SandersAaronD: Solid point, pushed something changing this name.

README.md
SandersAaronD: I think I agree, pulled it out

SandersAaronD:
I think we've got this addressed and we have span creation by default now with a cleaner contract, if there's anything still outstanding I can throw those changes on too.

tools.go
```diff
@@ -84,31 +88,88 @@ func ConvertTool[T any, R any](name, description string, toolHandler ToolHandler
...
+		// Set up span cleanup if we successfully created one
+		if span != nil {
+			defer span.End()
+			
+			// Add tool metadata as span attributes
+			span.SetAttributes(
+				attribute.String("mcp.tool.name", name),
+				attribute.String("mcp.tool.description", description),
+			)
+		}
```
sd2k: I'm pretty sure this can't panic, we'll just end up with a `NoopTracerProvider` if there's none configured, so we can safely get rid of this and simplify the code quite a bit.

(we can also get rid of all the `if span != nil` checks below)

```suggestion
		// Create OpenTelemetry span for tool execution (no-op when no exporter configured)
		config := GrafanaConfigFromContext(ctx)
		ctx, span = otel.Tracer("mcp-grafana").Start(ctx, fmt.Sprintf("mcp.tool.%s", name))
		defer span.End()
		
		// Add tool metadata as span attributes
		span.SetAttributes(
			attribute.String("mcp.tool.name", name),
			attribute.String("mcp.tool.description", description),
		)
```

sd2k (commented):
The reflection is a bit odd but I guess it does save an import!

SandersAaronD:
@sd2k commit + some tidy up for testing and the extra nil checks are done, this should be good to go

PR #207, opened 2025-07-18
fix: remove main.version from goreleaser ldflags

sd2k:
Our `main.version` variable is no longer just a string, it's a
`sync.Once`, which means that it cannot be overridden by an
ldflags value. This commit updates the goreleaser config to
specify a custom value for 'ldflags', which removes the
`main.version` variable from the defaults.

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -14,6 +14,7 @@ builds:
       - linux
       - windows
       - darwin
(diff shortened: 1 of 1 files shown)


sd2k:
Tested locally by running `goreleaser build --clean --skip validate`, which fails on main but succeeds here.

gitdoluquita (approved):

PR #217, opened 2025-07-23
chore: split out golangci-lint in CI

sd2k:
The dedicated GitHub action does a much better job of caching and
will be considerably faster.

This also removes the go.tools.mod way of running things because it's discouraged by the golangci-lint docs.


examples/tls_example.go
```diff
@@ -181,6 +183,7 @@ func customClientExample() {
 		TLSConfig: tlsConfig,
 	}
 	ctx = mcpgrafana.WithGrafanaConfig(ctx, config)
+	_ = ctx // Use ctx to avoid ineffectual assignment warning
```
dimitarvdimitrov, line 186: Nit: the example below says that what the transport may be used for. Maybe yiu can say the same about the ctx here

tools/pyroscope.go
```diff
@@ -370,14 +370,16 @@ func (c *pyroscopeClient) get(ctx context.Context, path string, params url.Value
 	if err != nil {
 		return nil, fmt.Errorf("failed to send request: %w", err)
 	}
-	defer res.Body.Close()
+	defer func() {
+		_ = res.Body.Close() //nolint:errcheck // Ignore close error in defer
```
dimitarvdimitrov, line 374: Nit but the comment doesn't really say anything useful 😄

Issue #166, opened 2025-06-09
Feature request: tool to generate URLs for navigation


sd2k:
It would be cool if the MCP server had a tool that could generate deeplinks into certain resources. LLMs can already take a stab at this but they get things wrong a lot (e.g. when resources don't follow conventions).

iamrajiv:
so i will put here what i am thinking @sd2k obviously this is great addition so what i think how we can do this is

structure can be like this:

```go
type GenerateDeeplinkParams struct {
	ResourceType ...
	UID ...
	// Optional fields for more complex links
	PanelID ...
	DashboardUID ...
}
```

so the llm would first use other tools like list_datasources etc to find the uid of a resource and then call generate deeplink with the resourcetype and the uid to get the deeplink if this is fine one thing is what resource type to support at start?

sd2k:
Yeah that approach sounds sensible to me 👍 

> what resource type to support at start

Let's start with just dashboards, panels and Explore queries, we can always add more later.

PR #218, opened 2025-07-23
feat: Add generate deeplink tool for dashboards, panels, and explore queries

iamrajiv:
Fixes: #166 

Screenshot 2025-07-23 at 11 59 31 PM


iamrajiv:
hey @sd2k can you take a look when you get a chance this builds on what we discussed in #166 i added the generate deeplink tool for dashboards panels and explore also added time support since i personally wanted that for sharing specific views let me know if anything needs changing

sd2k (commented):
Nice, @iamrajiv, this looks awesome.

I'd love to see an end-to-end test for this (in the `tests` directory), it'd be a great way to demo how it looks. Do you think you could have a go at that?

Could you also update the README with the new tool?

mcp-grafana
sd2k, line 1: this shouldn't be committed

run-mcp-grafana.sh
sd2k, line 1: This script shouldn't be needed, you can use `make run`, `make run-sse` etc instead?

sd2k (changes requested):
Thanks for adding docs & tests, I'm trying to run the tests locally but they're not all passing, perhaps some of the requirements are a bit strict. Honestly they're mostly smoke tests since it's impossible to control the underlying models so you can simplify them to just include the first one or two, that should be less flaky.

iamrajiv:
thanks for the review @sd2k. i’ve removed the unnecessary file i had created those to test my stuff and yes we can use make run i just noticed the makefile now

also i’ve updated the readme to include this feature and we have e2e tests too which are working fine

do you think we should have a demo video as well like how to connect usage all mcp features and all that not in this pr but maybe in a separate issue and improvement?

tools/navigation.go
```diff
@@ -0,0 +1,102 @@
...
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+
+	mcpgrafana "github.com/grafana/mcp-grafana"
+)
+
+type GenerateDeeplinkParams struct {
+	ResourceType string            `json:"resourceType" jsonschema:"required,description=Type of resource: dashboard\\, panel\\, or explore"`
+	UID          string            `json:"uid" jsonschema:"required,description=UID of the resource"`
```
sd2k: I think we should get rid of this parameter and use more specific `dashboardUID` / `datasourceUID` parameters instead, since it's a bit ambiguous what it should refer to (especially in the 'explore' case where it should be a datasource UID).

iamrajiv: sure @sd2k updated the field to `DashboardUID, DatasourceUID, PanelID`

iamrajiv:
also @sd2k the e2e tests pass locally but they need paid api keys and sometimes run into rate limits while testing since we already use string-based checks that cover the same logic should we make the llm checks optional for contributors without api credits?

sd2k:
Hey @iamrajiv, I'm happy to run the e2e tests and sort them out, it looks like just the `test_generate_deeplink_with_custom_params` ones are failing for me. I'll merge and fix any issues in a separate PR, thanks again!

PR #226, opened 2025-07-31
feat: Add proxied tools functionality, specifically make sure it proxies tempo correctly

SandersAaronD:
This seems good to my testing but it's hard to be sure; specifically the edge case around having two conflicting tempo MCP servers available from datasources may be buggy (since we only have one version on it currently!). Anyway, this sets up a flow for proxying tools via datasources and specifically carves out logic for handling tempo.

### DIFF
--- a/testdata/provisioning/datasources/datasources.yaml
+++ b/testdata/provisioning/datasources/datasources.yaml
@@ -27,3 +27,10 @@ datasources:
     access: proxy
     url: http://pyroscope:4040
     isDefault: false
+  - name: Tempo
+    id: 4
+    uid: tempo
+    type: tempo
+    access: proxy
+    url: http://tempo:3200
+    isDefault: false
--- a/tests/tempo_test.py
+++ b/tests/tempo_test.py
@@ -1,0 +1,138 @@
+import pytest
+
+
+class TestTempoMCPProxy:
+    """Test Tempo MCP proxy functionality.
+    
+    Note: These tests run with GRAFANA_URL environment variable set, which triggers
+    automatic discovery of Tempo datasources and registration of their tools at
+    server startup. If no Tempo datasources exist or GRAFANA_URL is not set,
+    no Tempo tools will be registered.
+    """
+
+    @pytest.mark.anyio
+    async def test_tempo_tools_discovery_and_proxying(self, mcp_client):
+        """Test that Tempo tools are discovered, wrapped, and properly proxied through our MCP server."""
+        
+        # Step 1: Verify all expected Tempo tools are discovered and registered
+        list_response = await mcp_client.list_tools()
+        all_tool_names = [tool.name for tool in list_response.tools]
+        
+        # Look for tempo-prefixed tools
+        tempo_tools = [name for name in all_tool_names if name.startswith("tempo_")]
+        expected_tempo_tools = [
+            "tempo_traceql_search",
+            "tempo_traceql_metrics_instant", 
+            "tempo_traceql_metrics_range",
+            "tempo_get_trace",
+            "tempo_get_attribute_names",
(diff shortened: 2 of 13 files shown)


SandersAaronD:
Remaining test failures are (probably) because the tempo version that implements the mcp server isn't deployed to hosted grafana yet, so this may have to be left to sit for a while.

testdata/provisioning/datasources/datasources.yaml
```diff
@@ -27,3 +27,10 @@ datasources:
     access: proxy
     url: http://pyroscope:4040
     isDefault: false
+  - name: Tempo
+    id: 4
+    uid: tempo
+    type: tempo
+    access: proxy
+    url: http://tempo:3200
+    isDefault: false
```
sd2k, line 36: perhaps worth adding two of these, so that we can test deduplication of tools in integration tests?

tests/tempo_test.py
```diff
@@ -0,0 +1,138 @@
...
+                "tool": "tempo_get_attribute_names",
+                "args": {"datasource_uid": "tempo"},
+                "expected_response_indicators": ["tempo", "datasource"]
+            },
+            {
+                "tool": "tempo_docs_traceql", 
+                "args": {"datasource_uid": "tempo"},
+                "expected_response_indicators": ["tempo", "traceql"]
+            }
+        ]
```
sd2k: nit: could use `pytest.mark.parametrize`

tools/proxied_tools.go
```diff
@@ -0,0 +1,651 @@
...
+		gc := mcpgrafana.GrafanaConfig{
+			URL:    grafanaURL,
+			APIKey: grafanaAPIKey,
+		}
+		ctx = mcpgrafana.WithGrafanaConfig(ctx, gc)
+		
+		// Create Grafana client
+		client := mcpgrafana.NewGrafanaClient(ctx, grafanaURL, grafanaAPIKey)
+		ctx = mcpgrafana.WithGrafanaClient(ctx, client)
+	}
```
sd2k: hm this won't work in some cases where we don't set these settings using env vars, e.g. in the LLM app integration we extract them from context. in those cases we would need to do the proxying at call-time, since that's the only time we have access to a context with the URL/API key 😞

tools/proxied_tools.go
```diff
@@ -0,0 +1,651 @@
...
+}
+
+type MCPCallToolResult struct {
+	Content []MCPContent `json:"content"`
+}
+
+type MCPContent struct {
+	Type string `json:"type"`
+	Text string `json:"text"`
+}
```
sd2k: can we get these from mcp-go instead?

SandersAaronD: Got all of these I could, I left the jsonrpcresponse because the mcp-go struct is much heavier in a way we don't need

tools/proxied_tools.go
SandersAaronD: I'm chewing on this, but I think I can probably do a small refactor to make it so this is importable into the llm app nicely

SandersAaronD: I decoupled these variables from the proxied tools thing, so it should be possible to do this without worrying about that now. I poked around the plugin and i think we can actually get credentials on startup though, it looks like backend.GrafanaConfigFromContext works for this

SandersAaronD:
Had a conversation with @sd2k about this, I think an extra concern to note is that when/if we have this set up to run as a hosted service that proxies to more than one grafana instance, this pattern is awkward. I don't think that should be blocking here, although it may mean that when we have this set up to work as a hosted service we should port the proxied tools up separately since there's a very different pattern surrounding them.

sd2k:
Superseded by #314

PR #238, opened 2025-08-08
Add mentions of Grafana Cloud intances in setup

federicotdn:
Extend the README a bit so that it also mentions that Grafana Cloud instances can be used as well!


sd2k (approved):
🙏

PR #251, opened 2025-08-15
Add HTTP header passthrough

callumj:
Background: we have Grafana behind a reverse proxy that authenticates the user to a given LDAP based on an authentication ticket (set in the header as `Authorization: Bearer <token>`). As such there is no need for a service account if we can authenticate them with their `Authorization: Bearer <token>` credentials - it also saves on a step where users need to setup a service account just to use the MCP.

This adds support for two ways to share credentials from the client:

* For STDIO, accept custom headers via the envs `GRAFANA_PASS_THROUGH_HEADER_FOO` e.g. `GRAFANA_PASS_THROUGH_HEADER_AUTHORIZATION="Bearer sdfdsadsdas"` 
* For SSE/HTTP support clients sending custom headers (that are allowlisted on the server side) that would then be passed along to the Grafana instance.


miked98:
Hello @callumj, could you explain why this one was closed? 
I think it's a very good idea!

callumj:
Hey @miked98, so it turned out that our reverse proxy/IAP system takes the same header `Authorization` that the Grafana API takes (makes sense, it is standardized) so I was able to set our IAP credentials as the API key when configuring MCP Grafana. So I figured it made little sense to complicate your code - but happy to add it if you think it would still be valuable.

We're also now bundling mcp-grafana in our binary so we can automatically fetch the proxy token (using a SSO flow) and set it automatically, here is an example:

```go
	token, err := getToken(grafanaUrl)
	if err != nil {
		zap.L().Fatal("Error getting IAP token", zap.Error(err))
	}

	// Grafana doesn't seem to use this config everywhere for STDIO
	// so we also need to set it as an ENV
	_ = os.Setenv("GRAFANA_URL", grafanaUrl)
	_ = os.Setenv("GRAFANA_API_KEY", token)
	grafanaConfig := mcpgrafana.GrafanaConfig{
		URL:    grafanaUrl,
		APIKey: token,
	}
	contextFunc := mcpgrafana.ComposedStdioContextFunc(grafanaConfig)
```

I do need to open some PRs though to fix the token & URL not being propagated everywhere (hence the use of setting ENVs).

miked98:
Thank you @callumj for sharing this. 
I think our use-case is a little bit different, we plan to host this MCP somewhere centrally to give people an option to use it. 
They should than authenticate as easy as possible via header so something like:

```json
{
	"servers": {
		"grafana-mcp": {
			"transport": "streamable_http",
			"url": "https://grafana-mcp.com/mcp",
			"headers": {
				"Authorization": "Bearer <GRAFANA_API_KEY>"
			},
			"type": "http"
	}	
}	
```

PR #254, opened 2025-08-19
feat(tools): Add ClickHouse support with querying and metadata retrieval

shalevshalit:
- Updated README to include ClickHouse as a supported datasource.
- Implemented ClickHouse querying capabilities, including SQL execution, database and table listing, and table description.
- Added integration and unit tests for ClickHouse functionalities.
- Enhanced the disabled tools configuration to include ClickHouse options.

### DIFF
--- a/README.md
+++ b/README.md
@@ -39,7 +39,7 @@ The dashboard tools now include several strategies to manage context window usag
 ### Datasources
 
 - **List and fetch datasource information:** View all configured datasources and retrieve detailed information about each.
-  - _Supported datasource types: Prometheus, Loki._
+  - _Supported datasource types: Prometheus, Loki, ClickHouse._
 
 ### Prometheus Querying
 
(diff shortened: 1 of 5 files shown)


talshalom:
@shalevshalit - why embedding clickhouse capabilities in Grafana mcp, instead of using separate clickhouse [mcp](https://github.com/ClickHouse/mcp-clickhouse)?

shalevshalit:
@talshalom same reason to use Prometheus or Loki.
The ability to query any grafana dashboard no matter what datasource you have.

talshalom:
Loki & Prometheus are integral to Grafana, for example querying Loki can help an agent to iterate on errors when creating panels. With your approach "_no matter what datasource_", the Grafana mcp becomes "all datasources" mcp. I am not sure that is the intention, unless I missed the use case here. Why do you think using an embedded capabilities is better than using a standalone clickhouse mcp?

picone:
> Loki & Prometheus are integral to Grafana, for example querying Loki can help an agent to iterate on errors when creating panels. With your approach "_no matter what datasource_", the Grafana mcp becomes "all datasources" mcp. I am not sure that is the intention, unless I missed the use case here. Why do you think using an embedded capabilities is better than using a standalone clickhouse mcp?

@talshalom  Not everyone has access to the ClickHouse credentials.
For example, as a developer I only work on editing panels, but the DataSource is not managed by me. So the only way I can get ClickHouse data is through the Grafana.

sd2k:
@shalevshalit Clickhouse support was added in #535 so I don't think this is needed any more.

Issue #257, opened 2025-08-22
JSONPath-based issue with pointer syntax extension /- for appending to arrays


talshalom:
@ioanarm - great addition supporting jsonpath - thanks. I found an issue where the agent can't append to existing panels array, require support for pointer syntax extension `/-` for appending to arrays using `op`: `add`. with current implementation there is a need to use `op`:`replace` and provide all existing panels plus the new one/s (better than the entire dashboard, yet would be better to append new panels only). 
The path resolution needs to handle these cases:
`$.panels/-` → append to panels array
`$.panels[2]/-` → append to nested array at index 2

PR #260, opened 2025-08-25
feet(tools): Add append functionality to update_dashboard tool

ioanarm:
Related to #257


sd2k (approved):
Slightly wary of mixing JSONPath and JSON Pointer but let's give it a go, it's not like we're married to any API 😅

ioanarm:
Thanks for taking a look @sd2k! I explored doing it only using a json pointer library but they don't cover fully our usecase - at least for now. This mixed approach isn't pure standards compliance 😬 , but adds the feature with the least imo complexity added possible!

talshalom:
@ioanarm @sd2k thanks for your support adding this feature. [rfc6901 section 3] (https://datatracker.ietf.org/doc/html/rfc6901#section-3):
"exactly the single character "-", making the new referenced value the (nonexistent) member after the last array element."

PR #270, opened 2025-09-02
feat: add support for https when using streamable-http transport

### LINKED ISSUE #244: Feature: TLS/HTTPS Endpoint Support for MCP Server
Currently, the server only serves on HTTP.
Can we add tls-support for the mcp server so that the server is hosted with tls?

sd2k:
mcp-go v0.39.0 adds support for https when using the streamable-http
transport. This commit adds support for configuring the server to
serve https using custom certificates via the CLI.

Fixes #244.

### DIFF
--- a/cmd/mcp-grafana/main.go
+++ b/cmd/mcp-grafana/main.go
@@ -115,7 +115,16 @@ func newServer(dt disabledTools) *server.MCPServer {
 	return s
 }
 
-func run(transport, addr, basePath, endpointPath string, logLevel slog.Level, dt disabledTools, gc mcpgrafana.GrafanaConfig) error {
+type tlsConfig struct {
+	certFile, keyFile string
+}
+
+func (tc *tlsConfig) addFlags() {
+	flag.StringVar(&tc.certFile, "server.tls-cert-file", "", "Path to TLS certificate file for server HTTPS (required for TLS)")
+	flag.StringVar(&tc.keyFile, "server.tls-key-file", "", "Path to TLS private key file for server HTTPS (required for TLS)")
+}
+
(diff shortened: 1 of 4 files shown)


cmd/mcp-grafana/main.go
```diff
@@ -135,10 +144,15 @@ func run(transport, addr, basePath, endpointPath string, logLevel slog.Level, dt
...
 		}
 	case "streamable-http":
-		srv := server.NewStreamableHTTPServer(s, server.WithHTTPContextFunc(mcpgrafana.ComposedHTTPContextFunc(gc)),
+		opts := []server.StreamableHTTPOption{
+			server.WithHTTPContextFunc(mcpgrafana.ComposedHTTPContextFunc(gc)),
 			server.WithStateLess(true),
 			server.WithEndpointPath(endpointPath),
-		)
+		}
+		if tls.certFile != "" || tls.keyFile != "" {
```
svennergr, line 152:
```suggestion
		if tls.certFile != "" && tls.keyFile != "" {
```

Just a nit, but this seems to be a condition in mcp-go. You can also leave it, to actually return mcp-go's error message.

https://github.com/mark3labs/mcp-go/blob/35fc389d56010b24ef49f966868b8b948d2af48e/server/streamable_http.go#L204-L215

svennergr (approved):
LGTM, left a nit, but feel free to ignore.

cmd/mcp-grafana/main.go
sd2k: Huh yeah, I thought I'd used your logic already but evidently not. I think I actually prefer just using mcp-go's error message though. Thanks for spotting!

PR #273, opened 2025-09-04
feat: add title/annotations to admin tools

sd2k:
These were missed off before.


kolesnikovae (approved):
LGTM. Thanks for fixing this!

PR #274, opened 2025-09-04
feat: add support for GRAFANA_SERVICE_ACCOUNT_TOKEN env var

### LINKED ISSUE #239: Consider renaming `GRAFANA_API_TOKEN` to something else
The environment variable `GRAFANA_API_TOKEN` could potentially create some confusion given that googling for "grafana api token" yields the following docs page: https://grafana.com/docs/grafana/latest/administration/service-accounts/migrate-api-keys/, saying that API tokens are deprecated.

Since the variable always expects a Service Account Token, maybe the name could be changed to something more fitting, e.g. `GRAFANA_SERVICE_ACCOUNT_TOKEN`.

sd2k:
Closes #239.


gerboland (approved):
Yep is good now

Issue #290, opened 2025-09-17
Claude desktop extension submission


joan-anthropic:
Hi from Anthropic! Thanks for your submission to Claude desktop extensions - we don't see a manifest.json in your [repo](https://github.com/grafana/mcp-grafana) nor a packaged .mcpb file per [spec](https://github.com/anthropics/mcpb); would you mind re-submitting via the [form](https://github.com/anthropics/mcpb)?

joan-anthropic:
One last request - can you ensure the manifest.json includes links to any relevant privacy_policies, and mcp tools are correctly annotated as read-only (readOnlyHint) or destructive (destructiveHint)? many thanks!

sd2k:
Hey @joan-anthropic, I know this issue is extremely old now but I think the [existing extension](https://claude.ai/directory/ant.dir.gh.grafana.grafana-mcp) in the catalog is very outdated and perhaps buggy. #1077 will begin building a new mcpb bundle on releases, is there an automated process we can follow to get new versions into the catalog?

PR #291, opened 2025-09-18
chore: add MCPB manifest

sd2k:
Addresses #290.


okianus:
Some notes from a conversation with @sd2k:

Two separate requirements:

1. the MCPB manifest, which defines how the server should be packaged up for installation and consumption (independent of Docker)
2. the server.json manifest, which is required for adding it to the registry (could either add our server as a Docker image or an MCPB package)

If we want to add it as a Docker image we need to build and tag our image, but the problem is that our [Docker image](https://hub.docker.com/r/mcp/grafana) isn't built by our repo but by [Docker's mcp-registry repo](https://github.com/docker/mcp-registry/blob/main/servers/grafana/server.yaml). 

We can update our Dockerfile to include the required label (`LABEL io.modelcontextprotocol.server.name="io.github.username/server-name"`) but the issue is that those Docker builds controlled by Docker's mcp-registry repo don't include versioned tags. A better approach is to do the MCPB manifest, then use that manifest in our server.json rather than using a Docker image, so the steps would be:

1. remove server.json and the docker build GitHub Actions workflow from that PR (for now)
2. add a separate GitHub action to build the MCPB and add it to the GitHub releases on each tag
3. make sure we can build the MCPB (and hopefully validate it)

Once that is done & merged, we can add a server.json which uses the MCPB as mentioned in [these requirements](https://github.com/modelcontextprotocol/registry/blob/main/docs/guides/publishing/publish-server.md#requirements-4).

But this approach has an issue that our server is a Go binary so needs the platform specified but that isn't supported by MCPB yet (see [this](https://github.com/anthropics/mcpb/issues/10) and [this](https://github.com/anthropics/mcpb/issues/63)). Using Node.js approach sounds like a bad idea, the alternative is to wait for MCPB to support platform-specific builds. Using our own Docker image published to the Docker hub is another option.

ioanarm:
@sd2k and @okianus we now have our own docker image with the label required to publish to the mcp registry. 
https://hub.docker.com/repository/docker/grafana/mcp-grafana

This should be good to go! Have a look and let me know!

sd2k:
Hm I think we can use the custom Docker image for the server.json but not for the MCPB manifest? Let's chat on Monday

sd2k:
I've removed the server.json and Docker changes from this PR since those were done in #323; this now just contains the manifest.json. Unfortunately it's still not workable due to https://github.com/anthropics/mcpb/issues/10 and https://github.com/anthropics/mcpb/issues/63, so I'll mark as draft for now.

sd2k:
Closing, not actionable until https://github.com/modelcontextprotocol/mcpb/issues/10 is fixed.

PR #292, opened 2025-09-18
feat: add more alerting tools

sd2k:
Add tools for creating, updating, and deleting Grafana alert rules.

Also update the list alert tools to include provisioning data as well as
runtime data.


sd2k:
@failedanswer hm if we do it that way then the nested fields will be undocumented, no? i.e. they don't have `jsonschema` tags so won't have `description`s for the model to read.

failedanswer:
> @failedanswer hm if we do it that way then the nested fields will be undocumented, no? i.e. they don't have `jsonschema` tags so won't have `description`s for the model to read.

Makes sense.
Although i still feel we can use the validate function from the client to double check before making a request:
 ```params := provisioning.NewPostAlertRuleParams().WithContext(ctx).WithBody(rule)```
 
 if you feel that might be unnecessary/redundant it should be fine.

tools/alerting.go
```diff
@@ -264,8 +382,267 @@ var ListContactPoints = mcpgrafana.MustTool(
...
+	mcp.WithTitleAnnotation("Create alert rule"),
+)
+
+type UpdateAlertRuleParams struct {
+	UID          string            `json:"uid" jsonschema:"required,description=The UID of the alert rule to update"`
+	Title        string            `json:"title" jsonschema:"required,description=The title of the alert rule"`
+	RuleGroup    string            `json:"ruleGroup" jsonschema:"required,description=The rule group name"`
+	FolderUID    string            `json:"folderUID" jsonschema:"required,description=The folder UID where the rule will be created"`
+	Condition    string            `json:"condition" jsonschema:"required,description=The query condition identifier (e.g. 'A'\\, 'B')"`
+	Data         interface{}       `json:"data" jsonschema:"required,description=Array of query data objects"`
```
ioanarm: In CreateAlertRuleParams you use `any`. Should we use the same?

tools/alerting.go
```diff
@@ -264,8 +382,267 @@ var ListContactPoints = mcpgrafana.MustTool(
...
+	NoDataState  string            `json:"noDataState" jsonschema:"required,description=State when no data (NoData\\, Alerting\\, OK)"`
+	ExecErrState string            `json:"execErrState" jsonschema:"required,description=State on execution error (NoData\\, Alerting\\, OK)"`
+	For          string            `json:"for" jsonschema:"required,description=Duration before alert fires (e.g. '5m')"`
```
ioanarm, line 505: is there any validation for this value on the api side? 🤔 
Does it worth adding one?

tools/alerting_test.go
```diff
@@ -495,3 +501,309 @@ func TestAlertingTools_ListContactPoints(t *testing.T) {
...
+		require.Contains(t, err.Error(), "title is required")
+	})
+}
+
+func TestAlertingTools_UpdateAlertRule(t *testing.T) {
+	t.Run("update existing alert rule", func(t *testing.T) {
+		ctx := newTestContext()
+
+		// First create a rule to update
+		sampleData := []any{
```
ioanarm, line 633: we can extract this, it's used in multiple places in the test file

tools/alerting.go
sd2k: I imagine it will be validated on the API side yeah, it might be painful to maintain the validation on our side too. We do use the OpenAPI validation [here](https://github.com/grafana/mcp-grafana/blob/bb6226343d9137fa82ae482bd03b0fa1ed85e85d/tools/alerting.go#L587) which is something.

tools/alerting.go
sd2k: Yep good call, done.

PR #309, opened 2025-10-03
feet(tools+context): Dynamic tool reg

ioanarm:
### Summary
Adds dynamic toolset functionality that allows clients to discover and selectively enable tool categories at runtime, significantly reducing context window usage by only loading tools when needed.

### Motivation
MCP tool descriptions can consume substantial context window space. By enabling dynamic discovery, clients only load the tools they actually need, preserving context for more important data.

### Implementation

New flag: `--dynamic-toolsets` enables dynamic toolset mode
Discovery tools: Added grafana_list_toolsets and grafana_enable_toolset meta-tools
Notification support: Sends tools/list_changed notification when toolsets are enabled, triggering client refresh

#### Integration with --enabled-tools:
- No flag --> all toolsets discoverable
- `--enabled-tools=""` --> no toolsets discoverable
- `--enabled-tools="datasources,dashboards"` --> only specified toolsets discoverable

#### Client Compatibility
✅ Supported: Cursor, VSCode (support notifications)
❌ Not yet: Claude Desktop, Claude Code (no notification support) there are open issues though for it

#### Known Behavior
Slight delay expected: There may be a few seconds of delay between calling grafana_enable_toolset and the tools becoming available in the client, as the client needs to receive and process the `tools/list_changed` notification.

Note: This is opt-in via the `--dynamic-toolsets` flag. Existing static tool registration remains the default behavior.

#### How it works with Cursor. 
Currently Sonnet 4.5 model selected - 1m tokens available for this conversation

Normal behavior all tools available: 
- At the beginning 13.9% of context used
   image

- How many tools are available 14.1%
   image

- List my prometheus datasources 14.8%
   image

Dynamic tool discovery enabled with all tools discoverable. 
- At the beginning 5.5% of context used
    image

- List available toolsets - 5.9%
   image

- Enable Datasource toolset 
   image

- List Datasources 7%
    image

### DIFF
--- a/cmd/mcp-grafana/main.go
+++ b/cmd/mcp-grafana/main.go
@@ -14,7 +14,7 @@ import (
 	"syscall"
 	"time"
 
-	"github.com/mark3labs/mcp-go/server"
+		"github.com/mark3labs/mcp-go/server"
 
 	mcpgrafana "github.com/grafana/mcp-grafana"
 	"github.com/grafana/mcp-grafana/tools"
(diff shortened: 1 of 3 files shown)


ioanarm:
Added a comment about supported protocols. 
Thanks @gerboland for trying it.

gerboland:
My Cursor is still struggling with this, while it can enable the datasource toolset, it fails to recognise the tools as callable.
image

I have verified the `notifications/tools/list_changed` event is being sent by `mcp-grafana` on toolset change:
image

So I think Cursor is the problem for me. Should it update the list of tools in its MCP settings UI on the `notifications/tools/list_changed` event?

dynamic_tools.go
```diff
@@ -0,0 +1,160 @@
...
+	}
+	return toolsets
+}
+
+// ToolsetInfo provides information about a toolset
+type ToolsetInfo struct {
+	Name        string `json:"name" jsonschema:"required,description=The name of the toolset"`
+	Description string `json:"description" jsonschema:"description=Description of what the toolset provides"`
+	Enabled     bool   `json:"enabled" jsonschema:"description=Whether the toolset is currently enabled"`
+}
```
sd2k, line 113: Do you think it would be worth including `ToolNames []string` here, as a more concrete hint of what tools this contains?

README.md
```diff
@@ -240,6 +243,60 @@ The `mcp-grafana` binary supports various command-line flags for configuration:
...
+  }
+}
+```
+
+**Limitations and Client Compatibility:**
+
+Protocol Support:
+- ✅ **stdio protocol** - Fully supported
+- ❌ **SSE (Server-Sent Events)** - Not yet supported
+- ❌ **Streamable HTTP** - Not yet supported
```
sd2k: Huh, is this because notifications aren't as well supported over streamable HTTP?

sd2k (commented):
Love the idea!

README.md
ioanarm: This is now fixed. After rebasing on top of your PR. https://github.com/grafana/mcp-grafana/pull/314/files 🥳

dynamic_tools.go
ioanarm: Done!

ioanarm:
> So I think Cursor is the problem for me. Should it update the list of tools in its MCP settings UI on the `notifications/tools/list_changed` event?

Yeah it should update on the UI as well. The moment you ask more tools to be enabled and the tool is called, the new tools should appear on the UI. 
Can you try with SSE? This works now.

README.md
```diff
@@ -168,6 +168,8 @@ Scopes define the specific resources that permissions apply to. Each action requ
 
 | Tool                              | Category    | Description                                                         | Required RBAC Permissions               | Required Scopes                                     |
 | --------------------------------- | ----------- | ------------------------------------------------------------------- | --------------------------------------- | --------------------------------------------------- |
+| `grafana_list_toolsets`           | Meta        | List available toolsets for [dynamic discovery](#dynamic-toolset-discovery) | None (meta-tool)                        | N/A                                                 |
+| `grafana_enable_toolset`          | Meta        | Enable a specific toolset [dynamically](#dynamic-toolset-discovery) | None (meta-tool)                        | N/A                                                 |
```
sd2k: nit: let's keep this table aligned

README.md
```diff
@@ -298,6 +301,60 @@ All read operations remain available, allowing you to query dashboards, run Prom
...
+
+**How it works:**
+1. Start the server with `--dynamic-toolsets` flag
+2. Use `grafana_list_toolsets` to discover available toolset categories
+3. Use `grafana_enable_toolset` to load specific toolsets (e.g., "datasource", "dashboard")
+4. The client receives a `tools/list_changed` notification and refreshes its tool list
+
+**Integration with `--enabled-tools`:**
+- No flag → all toolsets are discoverable
+- `--enabled-tools=""` → no toolsets are discoverable
```
sd2k: Is this ever likely to be useful? I wonder if we should just error in this case 🤷

README.md
```diff
@@ -298,6 +301,60 @@ All read operations remain available, allowing you to query dashboards, run Prom
...
+For even more efficient context window usage, you can enable **dynamic toolset discovery** mode with the `--dynamic-toolsets` flag. In this mode, tools are not loaded at startup. Instead, clients can discover available toolsets and selectively enable only the ones they need at runtime.
+
+**Benefits:**
+- Significantly reduces initial context window usage by not loading tool descriptions upfront
+- Tools are loaded on-demand only when needed
+- Preserves context space for more important data
+
+**How it works:**
+1. Start the server with `--dynamic-toolsets` flag
+2. Use `grafana_list_toolsets` to discover available toolset categories
```
sd2k: This list conflates things that the user does (start the server) with things that the LLM does (use those tools); we should make it clear that the user doesn't have to actually do those things, and it's the LLM doing it instead.

README.md
sd2k: [GitHub's dynamic tool discovery](https://github.com/github/github-mcp-server/blob/main/docs/server-configuration.md#dynamic-discovery-local-only) has 3 tools (`enable_toolset`, `list_available_toolsets`, `get_toolset_tools`) - did you experiment with this instead? I guess `get_toolset_tools` returns some more detailed info about the tools vs just the name (which we provide in the toolset list).

Either way we can iterate on that, not a blocker!

README.md
```diff
@@ -298,6 +301,60 @@ All read operations remain available, allowing you to query dashboards, run Prom
 - `--server.tls-cert-file`: Path to TLS certificate file for server HTTPS
 - `--server.tls-key-file`: Path to TLS private key file for server HTTPS
 
+### Dynamic Toolset Discovery
+
+For even more efficient context window usage, you can enable **dynamic toolset discovery** mode with the `--dynamic-toolsets` flag. In this mode, tools are not loaded at startup. Instead, clients can discover available toolsets and selectively enable only the ones they need at runtime.
```
```suggestion
For more efficient context window usage, you can enable **dynamic toolset discovery** mode with the `--dynamic-toolsets` flag. In this mode, tools are not loaded at startup. Instead, clients can discover available toolsets and selectively enable only the ones they need at runtime.
```

let's not pretend we're efficient in the first place 😅 

I think we could even be more clear here about our high default token usage, and instead say something like "Grafana supports a lot of functionality, so by default the server exposes a lot of tools which can eat into your context window. For more efficient context window usage, ...".

sd2k (commented):
Looks cool! Would love to see it in, think we need to fix conflicts and lints/tests but otherwise LGTM

README.md
ioanarm: I haven't no. I don't remember this being a thing when I raised this PR. 
I can have a look if we push this but I would do it separately mainly because I need to put time again into that! 😏

cmd/mcp-grafana/main.go
```diff
@@ -109,6 +111,68 @@ func (dt *disabledTools) addTools(s *server.MCPServer) {
...
+		{"admin", "Tools for administrative tasks (list teams, manage users)", []string{"list_teams", "list_users_by_org"}, tools.AddAdminTools},
+		{"pyroscope", "Tools for profiling applications with Pyroscope", []string{"list_pyroscope_label_names", "list_pyroscope_label_values", "list_pyroscope_profile_types", "fetch_pyroscope_profile"}, tools.AddPyroscopeTools},
+		{"navigation", "Tools for generating deeplink URLs to Grafana resources", []string{"generate_deeplink"}, tools.AddNavigationTools},
+		{"annotations", "Tools for managing annotations", []string{"get_annotations", "create_annotation", "create_graphite_annotation", "update_annotation", "patch_annotation", "get_annotation_tags"}, func(mcp *server.MCPServer) { tools.AddAnnotationTools(mcp, enableWriteTools) }},
+		{"rendering", "Tools for rendering dashboard panels as images", []string{"get_panel_image"}, tools.AddRenderingTools},
+	}
```
sd2k, line 156: This is gonna get out of date very quickly 😅 can we think of a better way of constructing this slice automatically?

cmd/mcp-grafana/main.go
```diff
@@ -147,8 +211,39 @@ func newServer(transport string, dt disabledTools) (*server.MCPServer, *mcpgrafa
...
+- folder: Manage folders
+- incident: Manage incidents
+- alerting: Manage alerts
+- oncall: Manage OnCall schedules
+- asserts: Grafana Asserts functionality
+- sift: Sift investigations
+- admin: Administrative tasks
+- pyroscope: Application profiling
+- navigation: Generate deeplinks
+- proxied: Access tools from external MCP servers (like Tempo) through dynamic discovery
```
sd2k, line 241: likewise I think this will get out of date quickly.

README.md
```diff
@@ -325,6 +328,59 @@ All read operations remain available, allowing you to query dashboards, run Prom
...
+Protocol Support:
+- ✅ **stdio protocol** - Fully supported
+- ✅ **SSE (Server-Sent Events)** - Fully supported
+- ✅ **Streamable HTTP** - Fully supported
+
+Client Support:
+- ✅ **Cursor** - Fully supported (supports notifications via stdio, SSE, and streamable-http)
+- ✅ **VS Code** - Fully supported (supports notifications via stdio, SSE, and streamable-http)
+- ❌ **Claude Desktop** - Not yet supported (no notification support, but open issues exist)
+- ❌ **Claude Code** - Not yet supported (no notification support, but open issues exist)
```
sd2k, line 377: Any chance these have been updated since we last checked?

sd2k (commented):
Very much like this!

ioanarm:
Closing for now as it seems that Cursor doesn't listen to notifications atm.

PR #314, opened 2025-10-10
feat: add support for proxying to Tempo datasources' MCP servers

### LINKED ISSUE #53: Add tools for Tempo
Useful tools (To add more):
- [ ] Search for a specific trace
- [ ] Search for traces based on tags
- [ ] Tag names 
- [ ] Tag values
- [ ] etc

Goals of this:
- [ ] Analysis of individual traces
- [ ] Comparison between traces
- [ ] Pattern analysis between traces 
- [ ] etc

### LINKED ISSUE #222: feat: Support MCP servers defined by datasources
We have #53, still open, for adding Tempo support. Towards the end of that issue you can find the link to the PR that adds an MCP server to Tempo.

It is almost certainly better to have Tempo (or any other team) supporting the Tempo datasource than to support it over in the grafana mcp server. We do, however, want to be able to query our core datasources after only installing this mcp server. It's difficult to see how to do this.

Currently, it looks like the way to do this would be to have the mcp server iterate over available datasources and check them for available mcp servers by quering a known endpoint or configuration for them. We could then expose an mcp tool from mcp-grafana that simply proxies to the datasource mcp server. This introduces a few difficulties:

1. In cases where different datasources have mcp servers with the same name but different arguments, we would have to give them different names before exposing them and keep track of which mcp tools could validly be called on which datasources.
2. It's not obvious how to proxy an mcp server; the protocol is underdocumented and this use case has not seen a lot of use generally.
3. To keep the listing of MCP tools current, the mcp server would have to poll the grafana instance periodically, which could be thorny.

I did look briefly at importing the Tempo MCP server into this one and exposing it, but this seems like it creates more problems than it solves.

If we don't have a better idea for how to expose datasource mcp tools to the user, it would seem like a good idea to proceed with implementing the proxying approach described above.

sd2k:
This commit adds support for 'proxied' datasource tools. These are tools
that are hosted on a remote MCP server (e.g., a Tempo datasource).
Before the first MCP method is called, provided a session is available
(i.e. not during an 'initialize' request), the MCP server will iterate
over all suitable available datasources in the configured Grafana
instance and attempt to discover and register tools from them, by
calling the remote MCP server's 'list_tools' method then using dynamic
session-based tools to register the discovered tools.

Tools are named with the format: <datasource_type>_<original_tool_name>.
For example, a Tempo traceql-search tool would be named:
tempo_traceql-search. Each added tool also has a 'datasourceUid'
parameter added to its input schema, which is used to identify the
datasource to query.

The `--disable-proxied` flag can be used to disable this feature.

Fixes #53.
Fixes #222.
Supersedes #226.

### DIFF
--- a/proxied_tools.go
+++ b/proxied_tools.go
@@ -250,19 +250,8 @@ func (tm *ToolManager) InitializeAndRegisterProxiedTools(ctx context.Context, se
 		}
 	}
 
-	state.mutex.Lock()
-	// Check if already initialized and registered
-	if state.proxiedToolsInitialized && len(state.proxiedTools) > 0 {
-		state.mutex.Unlock()
-		return
-	}
-
-	// If already initialized but not registered, skip discovery
-	alreadyDiscovered := state.proxiedToolsInitialized
-	state.mutex.Unlock()
-
-	// Step 1: Discover and connect (if not already done)
-	if !alreadyDiscovered {
+	// Step 1: Discover and connect (guaranteed to run exactly once per session)
+	state.initOnce.Do(func() {
 		// Discover datasources with MCP support
 		discovered, err := discoverMCPDatasources(ctx)
 		if err != nil {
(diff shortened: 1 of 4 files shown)


ioanarm (approved):
🚀 Perfect Ben!
Tested it locally and works as expected!

KyriosGN0:
hey @sd2k is there a doc on how to configure it ?

sd2k:
@KyriosGN0 no configuration should be required really, as long as you have recent enough Tempo datasources correctly configured in your Grafana instance the MCP server should find them and add corresponding tools.

PR #323, opened 2025-10-14
chore: add server.json for MCP registry

### LINKED ISSUE #287: Publish to the registry
How to https://github.com/modelcontextprotocol/registry/blob/main/docs/guides/publishing/publish-server.md

sd2k:
Also update the 'docker' CI workflow to publish to the registry after
building and pushing the image.

This is split out of #291 since it's actually a different thing; this PR
adds our MCP server to the registry, using the Docker image. The other
will add an MCPB manifest which we _could_ use for the registry instead,
and which can be used for Claude Desktop.

Following the docs for [publishing](https://github.com/modelcontextprotocol/registry/blob/main/docs/guides/publishing/publish-server.md) and the [server.json format](https://github.com/modelcontextprotocol/registry/blob/main/docs/reference/server-json/generic-server-json.md).

Closes #287.

### DIFF
--- a/server.json
+++ b/server.json
@@ -6,13 +6,13 @@
     "url": "https://github.com/grafana/mcp-grafana",
     "source": "github"
   },
-  "version": "0.6.5",
+  "version": "$VERSION",
   "packages": [
     {
       "registryType": "oci",
       "registryBaseUrl": "https://docker.io",
       "identifier": "grafana/mcp-grafana",
-      "version": "0.6.5",
+      "version": "$VERSION",
       "transport": {
         "type": "stdio"
       },


server.json
svennergr, line 1: Is this only needed for the `publish` workflow, or does it also need to be as static content in the repo? Reason why I ask is that we would need to remember to change versions `"version": "0.6.5"` in server.json now. And if this is really only a file needed for the Github workflow, we might maybe just sed/replace the version in the JSON before running `mcp-publisher`

sd2k: I think it's only needed for `publish` (at least I hope so, it would indeed be annoying to have to change it, although version bumps are done via CI so it might not be too bad). Our release CI is already doing the sed/replacing (using `jq`) so yeah maybe we just make it abundantly clear by setting it to `$VERSION`... good spot

PR #325, opened 2025-10-15
fix: don't use registryBaseUrl in server.json

sd2k:
[These docs][docs] are misleading, apparently oci images shouldn't
use registryBaseUrl but should just have the full Docker image tag
in the identifier.

[docs]: https://github.com/modelcontextprotocol/registry/blob/main/docs/guides/publishing/publish-server.md#example-serverjson-docker-hub


sd2k:
needs more changes, @ioanarm is working on it

Issue #264, opened 2025-08-27
Support multiple organisations


ire4ever1190:
Feature request to support multiple organisations. There is a number of organisations that I need to switch between when using grafana but have noticed that the MCP only uses the default organisation? (Possible bad prompting/user error)

Would be nice if the [switch context](https://grafana.com/docs/grafana/latest/developers/http_api/user/#switch-user-context-for-signed-in-user) and [list organisations](https://grafana.com/docs/grafana/latest/developers/http_api/user/#switch-user-context-for-signed-in-user) functions were exposed so that the client could use different organisations

Amoolaa:
I'll pick this one up

PR #331, opened 2025-10-15
feat: add user and org tools for multi-org support

Amoolaa:
issue ref: https://github.com/grafana/mcp-grafana/issues/264

This adds tools to support:
- fetching current org of current session
- fetching orgs of current session
- switching org of current session

### DIFF
--- a/.goreleaser.yaml
+++ b/.goreleaser.yaml
@@ -6,6 +6,9 @@ before:
   hooks:
     - go mod tidy
 
+git:
+  prerelease_suffix: "-rc"
+
 builds:
   - id: default
     env:


Amoolaa:
fwiw i tried these changes today on the grafana instance at work (running LDAP auth) and the change organization calls don't persist like they do when hitting the test instance running a normal basic auth setup. forgot about session cookies in this context (could be wrong but current org is tracked there). 

i think we should use the X-Grafana-Org-Id header for multi-org support instead. the clear solution is to make it a passthrough header from the mcp client, but this restricts the ability to switch orgs without modifying the mcp client config (which is kinda the user-oriented flow I was after). 

Another option is to call the available orgs tool to fetch org ids, then include an org id header param for every tool that requires it, and wrap the grafana client with the header if specified, but seems excessive. Open to discussions here

sd2k:
Yeah I was a little worried this would use session cookies to track it, I wasn't sure how else it would work in a load-balanced scenario.

I would like to avoid the LLM having to provide org ID to everything. My feeling is it would be nice if the user could just say 'use this org', then the LLM could call some sort of 'set org' tool which stores the org on the session and sends the `X-Grafana-Org-Id` header with every subsequent request. Do you think that flow would work OK?

Two issues I can see:

- needs stateful sessions (also required for #314 which I hope to get merged soon), not the end of the world
- sessions don't work/exist in `mcp-go` when using stdio (we work around that in #314 too)

It should be quite easy to test though!

In terms of implementing, it might be worth waiting until #314 is in so there's already a `SessionManager` you can build off, sorry about that. I'll ping folks for some reviews today

Amoolaa:
> My feeling is it would be nice if the user could just say 'use this org', then the LLM could call some sort of 'set org' tool which stores the org on the session and sends the X-Grafana-Org-Id header with every subsequent request. Do you think that flow would work OK?

yep. we'd still probably need a tool to fetch the current user's orgs to map the id to some human readable value. only issue with adding state is that it no longer becomes trivial to horizontally scale, but im not sure many people are doing that/need to in the case of remote deployments, and then i guess the next logical step would be to store it in a distributed cache

Amoolaa:
@sd2k what do you think about making the org header pass through anyways? and override if the org switch tool is called?

sd2k:
Hm by pass through do you mean use a setting from env/config (`GRAFANA_ORG_ID`) and set the header on the client if it's set? That SGTM.

PR #335, opened 2025-10-17
docs: restructure README and separate into separate files

sd2k:
Needs a good old read through before merging, but the README was pretty
unmanageable before.


sd2k:
Superseded by #441 and #442

PR #355, opened 2025-10-22
docs: suggest Editor role as simple RBAC alternative (issue #240)

### LINKED ISSUE #240: Consider suggesting Account Service roles suggestions
When creating a new Service Account for use with the MCP server, one must configure the correct RBAC permissions in order for the server to be able to access different parts of Grafana.

A quick way of doing this (but less granular) is to use one of the predefined roles such as "Editor" when creating the account. This makes the process easier, specially for users not very familiar with RBAC.

Maybe the README could suggest a specific role to use, in case the user does not want to manually configure RBAC.

Pika-Gopher:
Fixes #240 
Adds two short tips to the README recommending the built-in Editor role as a simple alternative to fine‑grained Grafana RBAC for service accounts to help users get started quickly. The tips include a brief caveat that Editor grants broader privileges than minimal scopes. Prefer least‑privilege for production environments.


sd2k:
LGTM, thanks @Pika-Gopher !

PR #369, opened 2025-10-29
ci(docker): add multi-platform build support

hyperb1iss:
Add multi-platform build support for Docker images

Configure Docker workflow to build images for both linux/amd64 and linux/arm64 platforms. This enables the container to run on ARM-based systems such as Apple Silicon and ARM servers.


sd2k (approved):
Thanks!

PR #373, opened 2025-10-31
fix(tests): update Claude model, switch from langevals to deepeval

ioanarm:
* Replaced langevals with deepeval for e2e python tests
* Updated Claude model

tests are running again 🥳 ! They fail but this is separate.


tests/conftest.py
```diff
@@ -15,11 +15,77 @@
...
+
+    # After test: clean up worker state
+    try:
+        from litellm.litellm_core_utils.logging_worker import logging_worker_instance
+        if hasattr(logging_worker_instance, "_queue"):
+            logging_worker_instance._queue = None
+        if hasattr(logging_worker_instance, "_task"):
+            logging_worker_instance._task = None
+    except Exception:
+        pass
```
sd2k: I don't think we need this any more?

ioanarm: it seems that tests don't give loop related errors so I will remove!

PR #375, opened 2025-11-03
docs: update enabled-tools description

inhun:
Updated the README to avoid confusion by changing the phrase "list of enabled tools" to "list of enabled categories" for the `--enabled-tools` option.


sd2k:
Thanks!

PR #380, opened 2025-11-04
Added matchers in listLokiLabelValues

rahulvijil:
## 🧩 Summary

This PR adds support for optional **`matchers`** in the `list_loki_label_values()` function to allow filtered label value queries in Loki.

Previously, the MCP API did not accept `matchers`, causing the function to always return unfiltered results (all label values across namespaces and apps).  
With this change, users can now pass Loki stream selectors (e.g., `{app="...", namespace="..."}`) to filter results as needed.

---

## 🔍 Why this change was needed

My agent was calling the Grafana MCP API as follows:

```json
list_loki_label_values({
  "datasourceUid": "<datasource>",
  "labelName": "<label-name>",
  "startRfc3339": "<start-timestamp format., YYYY-MM-DDTHH:MM:SS.ssssssZ>",
  "endRfc3339": "<end-timestamp format., YYYY-MM-DDTHH:MM:SS.ssssssZ>",
  "matchers": "{app=\"<app>\", namespace=\"<namespace>\"}"
})
```

Since the backend did not support the `matchers` field, the query ignored these filters and returned all values.
This PR fixes that gap by forwarding the `matchers` parameter as a query argument to Loki’s `/loki/api/v1/label/{name}/values` endpoint.

## Key Changes
- Updated ListLokiLabelValuesParams to include an optional Matchers string field.
- Modified listLokiLabelValues() to pass args.Matchers into fetchData().
- Updated fetchData() to handle and append matchers (as query param) when provided.
- Added debug prints for final URL and Loki response (for verification).

## Notes
- Tested against Loki v3.5.2 (/loki/api/v1/status/buildinfo confirmed).
- No breaking changes introduced — matchers remains optional.

### DIFF
--- a/tools/loki.go
+++ b/tools/loki.go
@@ -190,7 +190,7 @@ func (c *Client) fetchLabels(ctx context.Context, urlPath string, startRFC3339, 
 	if matcher != "" {
 		params.Add("query", matcher)
 	}
-	return c.handleAPICall(ctx, urlPath, params)
+	return c.handleAPICall(ctx, params, urlPath)
 }
 
 // fetchData is a generic method to fetch data from Loki API
@@ -202,7 +202,7 @@ func (c *Client) fetchData(ctx context.Context, urlPath string, startRFC3339, en
 	if endRFC3339 != "" {
 		params.Add("end", endRFC3339)
 	}
-	return c.handleAPICall(ctx, urlPath, params)
+	return c.handleAPICall(ctx, params, urlPath)
 }
 
 func NewAuthRoundTripper(rt http.RoundTripper, accessToken, idToken, apiKey string, basicAuth *url.Userinfo) *authRoundTripper {
@@ -247,6 +247,7 @@ type ListLokiLabelNamesParams struct {
 	DatasourceUID string `json:"datasourceUid" jsonschema:"required,description=The UID of the datasource to query"`
 	StartRFC3339  string `json:"startRfc3339,omitempty" jsonschema:"description=Optionally\\, the start time of the query in RFC3339 format (defaults to 1 hour ago)"`
 	EndRFC3339    string `json:"endRfc3339,omitempty" jsonschema:"description=Optionally\\, the end time of the query in RFC3339 format (defaults to now)"`
+	Matchers      string `json:"matchers,omitempty" jsonschema:"description=Optionally\\, the optional filters such as app, namespace wrapped up with curly braces {"app"="<app-name>", "namespace"="<namespace>"} (defaults to {})"`
 }
 
 // listLokiLabelNames lists all label names in a Loki datasource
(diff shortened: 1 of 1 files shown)


tools/loki.go
```diff
@@ -256,6 +256,7 @@ type ListLokiLabelValuesParams struct {
 	LabelName     string `json:"labelName" jsonschema:"required,description=The name of the label to retrieve values for (e.g. 'app'\\, 'env'\\, 'pod')"`
 	StartRFC3339  string `json:"startRfc3339,omitempty" jsonschema:"description=Optionally\\, the start time of the query in RFC3339 format (defaults to 1 hour ago)"`
 	EndRFC3339    string `json:"endRfc3339,omitempty" jsonschema:"description=Optionally\\, the end time of the query in RFC3339 format (defaults to now)"`
+	Matchers      string `json:"matchers"`
```
svennergr: Hi @rahulvijil,

thanks for your contribution. Would you mind adding two things to this `matchers` prop:
1. `omitempty`
2. A description that would help the LLM understand what this parameter is used for.

tools/loki.go
```diff
@@ -265,16 +266,21 @@ func listLokiLabelValues(ctx context.Context, args ListLokiLabelValuesParams) ([
...
+	}
+
+	// Call the original fetchData with start/end as before
 	result, err := client.fetchData(ctx, urlPath, args.StartRFC3339, args.EndRFC3339)
 	if err != nil {
-		return nil, err
+		return nil, fmt.Errorf("fetching label values: %w", err)
 	}
 
 	if len(result) == 0 {
```
svennergr, line 306: Are these changes needed?

rahulvijil: Not actually, I will revert `fmt.Errorf("fetching label values: %w", err)` to `err`

tools/loki.go
```diff
@@ -140,14 +140,23 @@ func (c *Client) makeRequest(ctx context.Context, method, urlPath string, params
...
 		params.Add("start", startRFC3339)
 	}
 	if endRFC3339 != "" {
 		params.Add("end", endRFC3339)
 	}
+	if optional != "" {
+		params.Add("query", optional)
+	}
+	fullURL := fmt.Sprintf("%s?%s", urlPath, params.Encode())
+	fmt.Println("Final request URL:", fullURL)
```
svennergr: Hm, I think instead adding this here as optional, we might be better creating a new `fetchLabels` method.

tools/loki.go
rahulvijil: Added the description

tools/loki.go
```diff
@@ -139,15 +139,9 @@ func (c *Client) makeRequest(ctx context.Context, method, urlPath string, params
...
-	params := url.Values{}
-	if startRFC3339 != "" {
-		params.Add("start", startRFC3339)
-	}
-	if endRFC3339 != "" {
-		params.Add("end", endRFC3339)
-	}
+func (c *Client) handleAPICall(ctx context.Context, params url.Values, urlPath string) ([]string, error) {
+	fullURL := fmt.Sprintf("%s?%s", urlPath, params.Encode())
+	fmt.Println("Final request URL:", fullURL)
```
sd2k, line 144: Could you remove this Println?

tools/loki.go
```diff
@@ -174,9 +168,43 @@ func (c *Client) fetchData(ctx context.Context, urlPath string, startRFC3339, en
 		return []string{}, nil
 	}
 
+	pretty, _ := json.MarshalIndent(labelResponse, "", "  ")
+	fmt.Println("Loki Full Response:\n", string(pretty))
```
sd2k, line 172: This Println also needs removing

ioanarm:
Hello @rahulvijil ! I am closing the PR as it's stale for a long time now. 
Feel free to pick this up again though, we'd be happy to review!

PR #381, opened 2025-11-04
fix: loki metric type query

### LINKED ISSUE #368: Loki metric queries return null labels for aggregate groupings
## Problem
When executing Loki metric queries with aggregate functions like `sum by()` using the `query_loki_logs` tool, the results include metric values but all label groupings return as `null`.

## Expected Behavior
Metric queries with grouping should return the labels that were grouped by. For example:
```logql
sum by(app, namespace) (count_over_time({namespace="dev2"} | json | Level="Error" [3h])) > 0
```
Should return results with `app` and `namespace` labels populated.

## Actual Behavior
The query executes successfully but all results have `"labels": null`:
```json
[
  {"timestamp":"1761755580","line":"5","labels":null},
  {"timestamp":"1761755623","line":"5","labels":null},
  ...
]
```

## Steps to Reproduce
1. Configure Grafana MCP with a Loki datasource
2. Execute a metric query with grouping:
```
query_loki_logs(
  datasourceUid="<loki-uid>",
  logql='sum by(app, namespace) (count_over_time({namespace="dev2"} [3h])) > 0',
  ...
)
```
3. Observe that `labels` field is `null` in all results

## Workaround
Instead of using metric queries, fetch raw log samples and extract labels from individual log entries.

## Environment
- Grafana MCP version: latest (Docker image `mcp/grafana`)
- Transport: stdio via Docker

zhiyu0729:
Summary:
- Fix Loki metric queries by routing instant requests through /loki/api/v1/query, range requests through /loki/api/v1/query_range, and correctly handling time, direction, and step parameters.
- Update LogEntry parsing so metric vectors and matrices populate labels and value fields correctly without mixing log semantics.

Issue
- Fix #368

### DIFF
--- a/tools/loki.go
+++ b/tools/loki.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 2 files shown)


zhiyu0729:
Thanks a lot for the detailed feedback!
I agree with your points, I’m going to close this PR for now.

igor-raits:
@sd2k any chance there is going to be another solution in place anytime soon? Without this feature, Loki returns simply too much data making the MCP server unusable for some of the cases in our environment.

PR #408, opened 2025-12-01
feat: add support to specify public URL via GRAFANA_PUBLIC_URL

greedy52:
Add support to specify public URL via `GRAFANA_PUBLIC_URL` environment variable or `X-Grafana-Public-URL` header.

The public URL is used to generate links to Grafana. Useful when the MCP server access the Grafana instance via a **local** URL, while users access the Grafana instance with this **public** URL (where Grafana/MCP server behind load balancers or reverse proxies).

### DIFF
--- a/README.md
+++ b/README.md
@@ -410,11 +410,14 @@ When an organization ID is provided, the MCP server will set the `X-Grafana-Org-
          "command": "mcp-grafana",
          "args": [],
          "env": {
-           "GRAFANA_URL": "http://localhost:3000",  // Or "https://myinstance.grafana.net" for Grafana Cloud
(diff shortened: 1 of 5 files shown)


sd2k (commented):
Thanks @greedy52. I'm curious whether, for your use case, it would work if we fetched Grafana's public URL from its settings instead (`/api/frontend/settings` in the `appUrl` field of the JSON). If we can avoid adding another config var that would be good.

Trollgeir:
@greedy52 is this work discontinued? We would also like this feature.

ioanarm:
Closing this one as it's stale for some months now. 
There is another [PR](https://github.com/grafana/mcp-grafana/pull/664) about this.

Issue #406, opened 2025-11-27
CVEs in container image


nissessenap:
Hi

There are currently a number of CVEs in the container and it contains a number of critical ones 

- CVE-2023-45853	
- CVE-2019-8457

I'm assuming these CVEs isn't an issue due to the MCP server don't use SQLlite for exmaple.

To see exactly which CVEs that exist, you can use [grype](https://github.com/anchore/grype) for example
`grype grafana/mcp-grafana:0.7.9`

The problem is that the image used is very broad.
https://github.com/grafana/mcp-grafana/blob/5ff432c30928e5cf93d419f194be73058e3a83c3/Dockerfile#L20

My suggestion is to either start using [KO](https://ko.build/), just like we do in the [grafana-operator](https://github.com/grafana/grafana-operator/blob/95ffa394e08c28179dc80d38d3c9ae31544047cb/.github/workflows/release.yaml#L206).
But I don't know how the general [MCP](https://hub.docker.com/mcp/server/grafana/overview) container gets built the same way or not compared to https://hub.docker.com/r/grafana/mcp-grafana 
It looks like the mcp-publisher only gets an image https://github.com/grafana/mcp-grafana/blob/5ff432c30928e5cf93d419f194be73058e3a83c3/.github/workflows/docker.yml#L75 so we could probably use ko. 

Or we just change it to using a small image, like distroless or chainguard (this is what go does nowadays).

I'm willing to help out with a PR around this, but I don't want to waste my time unless you think it's a good idea.
So please provide some feedback on which way to go.

sd2k:
Hi @nissessenap , agreed this is a good idea, happy to try out ko as well. Contributions would be very welcome, if not I'll try and get round to this in the next few days.

PR #410, opened 2025-12-02
Use KO to build container image

nissessenap:
Solves #406 

@sd2k realized this became harder than I thought. Since you are using a number of internal github actions with access requirements to Vault for example.

I did some educated guesses based on how it's done in the grafana-operator, but you would definitely need todo things on your side.

The main issue is that KO don't support setting a custom entrypoint the same way as is currently done in the Dockerfile.

- Dockerfile image: docker run grafana/mcp-grafana → starts SSE server on 0.0.0.0:8000
- Ko image: docker run grafana/mcp-grafana → uses binary defaults

So it would be a breaking change.
Claude updated the README for me, I have barely read through it.

I can take a real stab at updating the README, but before doing, so I wanted to point out the breaking change and get your thoughts on it. + handing over the GHA stuff to you since I don't have internal access, and I have just guessed allot. 

I fully understand if we just close this PR, but hopefully it was some use for you when thinking about using KO.

### DIFF
--- a/README.md
+++ b/README.md
@@ -352,43 +352,43 @@ When an organization ID is provided, the MCP server will set the `X-Grafana-Org-
 
    - **Docker image**: Use the pre-built Docker image from Docker Hub.
 
-     **Important**: The Docker image's entrypoint is configured to run the MCP server in SSE mode by default, but most users will want to use STDIO mode for direct integration with AI assistants like Claude Desktop:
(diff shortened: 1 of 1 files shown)


sd2k:
Sorry for the very long delay with looking at this. I think the breaking change to the Docker image's entrypoint is likely to be a dealbreaker, and unfortunately I don't have time to learn & maintain a `ko` based build process right now, but I'll keep it in mind in case we have further issues with our Docker builds. Thank you anyway!

PR #455, opened 2026-01-06
fix: update mcp-tokens-action to valid version

jedrzejdocs-code:
The mcp-tokens-action commit SHA was returning 404. Updated to @main.

### DIFF
--- a/.github/workflows/token-check.yml
+++ b/.github/workflows/token-check.yml
@@ -45,7 +45,8 @@ jobs:
             ANTHROPIC_API_KEY=anthropic:api-key
 
       - name: Analyze tokens
-        uses: sd2k/mcp-tokens-action@main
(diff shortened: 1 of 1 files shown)


sd2k:
Hm I think the action version is actually correct? (https://github.com/sd2k/mcp-tokens-action/tree/2c121e1bb2ef8bf6c2fc9fe4a342c205b6d37c85 works).

But the install.sh script seems to be failing for some reason, I'll look into it.

sd2k:
Looks like our API requests to GitHub were being rate limited. I've fixed this in the latest version of the action and updated this PR to point to it, thanks for raising this. If you can sign the CLA then we can get this merged :slightly_smiling_face:

sd2k:
@jedrzejdocs-code we enforce pinning actions to specific SHAs. Can you use `642a929d4372959ddaa87ad5843c38e9accab25a` instead of `main` please?

PR #462, opened 2026-01-11
feat: add disableProvenance parameter for alert rule UI editability

### LINKED ISSUE #461: Alerts created by MCP cannot be edited in the UI
Using `provenance: "api"` to protect IaC created resources is logical but when creating with the MCP seems to be a bit wrong. The MPC is just an extension of an engineer.

It would be preferable to be able to easily manually edit any resources created by the MCP, as there could be errors in it.

Would it be possible to change how this works?

lreisner:
## Summary

Add `disableProvenance` parameter to `create_alert_rule` and `update_alert_rule` tools.

When `true` (default), sets the `X-Disable-Provenance` header so alerts remain editable in the Grafana UI. When `false`, alerts are marked with provenance 'api' and locked from UI editing.

## Changes

- Added `disableProvenance` parameter to `CreateAlertRuleParams` and `UpdateAlertRuleParams`
- Default behavior is `true` (UI-editable) to prevent users from accidentally locking themselves out
- Updated README with documentation for the new parameter
- Added integration tests for both create and update with various `disableProvenance` values
- Added unit tests for parameter validation

## Testing

- All existing tests pass
- New integration tests verify the header is set correctly
- New unit tests verify parameter validation

Closes #461


README.md
```diff
@@ -66,6 +66,8 @@ The dashboard tools now include several strategies to manage context window usag
 ### Alerting
 
 - **List and fetch alert rule information:** View alert rules and their statuses (firing/normal/error/etc.) in Grafana. Supports both Grafana-managed rules and datasource-managed rules from Prometheus or Loki datasources.
+- **Create and update alert rules:** Create new alert rules or modify existing ones. By default, alerts created via MCP remain editable in the Grafana UI. Set `disableProvenance: false` to mark alerts as API-provisioned (locked from UI editing).
```
sd2k: The `disableProvenance` thing isn't something a user can do here, so we can just omit it from the README.

sd2k (approved):
LGTM except one comment on the README change. Thanks!

README.md
sd2k:
```suggestion
- **Create and update alert rules:** Create new alert rules or modify existing ones.
```

lreisner: Can you clarify this for me?  I made `disableProvenance` a configurable parameter on the tools, so a user **can** instruct their agent to enable or disable provenance.  I did this for backwards compatibility in case anyone was relying on the old behavior.

So I can either:
1. Remove just the comment (perhaps because you think it is too much detail for the README here?) but leave the configurable parameter in the code.
2. Remove the configurable parameter so it always disables provenance.
3. Keep everything as-is.

Let me know what you'd prefer.

sd2k: Ah sorry, just 1. is fine, I'm concerned about the README growing too large. Keeping the parameter around is good with me.

lreisner: Okie doke, it's done.

sd2k:
Thanks! Unfortunately there's now a merge conflict, happy to merge once fixed.

lreisner:
> Thanks! Unfortunately there's now a merge conflict, happy to merge once fixed.

Okay, the conflicts are fixed.  I also had to update the tests to accommodate type changes.  Everything should be good now.

PR #477, opened 2026-01-14
[WIP] Add tools for library elements

valery1707:
Tests will be added soon


ioanarm:
Hello @valery1707! I will close this PR as there is no activity for the last 2 months. 
In general we are trying to avoid adding new tools atm as the server already adds about 15k tokens, though if you are interested to push this in the future we can have a discussion about it :)

PR #491, opened 2026-01-16
feat(loki): Add regex-based log masking for sensitive data protection

Marie673:
# feat(loki): Add server-side log masking for sensitive data protection

## Summary

Add regex-based log masking capability to the Loki log query tool. This feature automatically masks sensitive information (PII, authentication tokens, etc.) in log data before returning results to AI assistants, enhancing security and compliance.

**Key Design Decision**: Masking configuration is set at server startup by system administrators. AI/LLM cannot modify or override masking settings via tool calls. This ensures organizational security policies are enforced consistently and prevents unintended data exposure.

### New Files

| File | Description |
|------|-------------|
| `tools/loki_masking.go` | Core masking logic, builtin pattern definitions, and consistent hashing |
| `tools/loki_masking_unit_test.go` | Comprehensive unit tests, benchmarks, SLO compliance tests |

### Modified Files

| File | Description |
|------|-------------|
| `tools/loki.go` | Integration of masking processing via Context-based masker injection |
| `tools/loki_test.go` | Integration tests for masking functionality |
| `mcpgrafana.go` | `LokiMaskingConfig` struct, `WithMasker`/`MaskerFromContext` context functions |
| `cmd/mcp-grafana/main.go` | CLI flags, environment variable support, masker initialization |
| `mcpgrafana_test.go` | Context management tests |
| `README.md` | Documentation for masking feature and configuration |
| `examples/tls_example.go` | Context function signature updates |

### Server-Level Configuration

Masking is configured at server startup via environment variables:

```bash
# Enable masking
export LOKI_MASKING_ENABLED=true

# Specify patterns (comma-separated)
export LOKI_MASKING_PATTERNS=email,credit_card,ip_address,jwt_token
```

Or via CLI flags:

```bash
./mcp-grafana \
  --loki-masking-enabled \
  --loki-masking-patterns=email,credit_card,ip_address,jwt_token
```

Or via Docker Compose:

```yaml
environment:
  - LOKI_MASKING_ENABLED=true
  - LOKI_MASKING_PATTERNS=email,credit_card,ip_address,jwt_token
```

### Builtin Patterns

All patterns use SHA256-based consistent hashing. The same sensitive value always produces the same hash, enabling correlation analysis while protecting the actual data.

| Pattern ID | Description | Example |
|------------|-------------|---------|
| `email` | Email addresses | `user@example.com` → `[MASKED:email:a1b2c3d4]` |
| `phone` | E.164 international phone numbers | `+819012345678` → `[MASKED:phone:e5f6a7b8]` |
| `credit_card` | Credit card numbers | `4111-1111-1111-1111` → `[MASKED:credit_card:c9d0e1f2]` |
| `ip_address` | IPv4/IPv6 addresses | `192.168.1.1` → `[MASKED:ip_address:3a4b5c6d]` |
| `mac_address` | MAC addresses (EUI-48/EUI-64, all formats) | `00:1A:2B:3C:4D:5E` → `[MASKED:mac_address:7e8f9a0b]` |
| `api_key` | API keys/tokens | `api_key=abc123...` → `[MASKED:api_key:1c2d3e4f]` |
| `jwt_token` | JWT tokens | `eyJ...` → `[MASKED:jwt_token:5a6b7c8d]` |

### Consistent Hashing

All masked values include a deterministic hash suffix derived from the original value:

- **Same value → Same hash**: Identical sensitive values always produce the same masked output, enabling correlation (e.g., tracking the same email or IP across log entries)
- **Different values → Different hashes**: Different values produce different hashes, preserving distinctness
- **One-way**: SHA256 hashing is irreversible — the original value cannot be recovered from the hash

#### MAC Address Format Normalization

MAC addresses additionally support format-agnostic normalization. All representations of the same address produce the same hash:

- `00:1A:2B:3C:4D:5E` (colon-separated)
- `00-1A-2B-3C-4D-5E` (dash-separated)
- `001A2B3C4D5E` (no separators)
- `001A.2B3C.4D5E` (Cisco dot format)
- Case-insensitive (uppercase/lowercase produce same hash)

### Supported MAC Address Formats

| Format | EUI-48 Example | EUI-64 Example |
|--------|---------------|----------------|
| Colon-separated | `00:1A:2B:3C:4D:5E` | `00:1A:2B:3C:4D:5E:6F:70` |
| Dash-separated | `00-1A-2B-3C-4D-5E` | `00-1A-2B-3C-4D-5E-6F-70` |
| No separators | `001A2B3C4D5E` | `001A2B3C4D5E6F70` |
| Cisco dot format | `001A.2B3C.4D5E` | `001A.2B3C.4D5E.6F70` |

### Security Model

| Aspect | Detail |
|--------|--------|
| Configuration by | System administrators only |
| Configuration timing | Server startup only |
| Configuration method | Environment variables / CLI flags |
| AI/LLM access to settings | **Not possible** |
| Modification via tool calls | **Not possible** (no masking parameter in API) |
| Hash reversibility | **Not possible** (SHA256 is one-way; only first 4 bytes used for display) |

### Context-Based Masker Injection

```
Server Startup → Load Config → Create LogMasker → Inject into Context
                                                         ↓
MCP Client → query_loki_logs → Fetch Logs → Apply Masker from Context → Return Masked Logs
```

The masking is applied automatically when enabled — AI assistants receive masked logs without any ability to bypass or modify the masking configuration.

### Consistent Hashing Flow

```
Match sensitive value → [Normalize if applicable] → SHA256 hash → Take first 4 bytes → [MASKED:<id>:<hex>]
```

### Performance

- Maximum pattern count: 20
- SLO: 100 entries × 20 patterns completed within 100ms
- Regex patterns compiled once at server startup (RE2 engine with linear time guarantee)
- Consistent hashing uses SHA256 computed per match (negligible overhead)

### Error Handling

- Invalid builtin pattern identifier at startup → Warning log, skip invalid patterns
- Pattern count exceeds limit → Warning log, apply first 20 patterns only
- Masking failure at runtime → Return error, never return unmasked data

### Backward Compatibility

- Masking is disabled by default (`LOKI_MASKING_ENABLED=false`)
- When disabled, behavior is identical to previous versions
- No changes to `query_loki_logs` tool API (no new parameters added)

## Breaking Changes

- `ComposedStdioContextFunc`, `ComposedSSEContextFunc`, `ComposedHTTPContextFunc` signatures changed (added `masker any` parameter)
- If using these functions directly from external code, pass `nil` as the second argument to maintain existing behavior

### Unit Tests

- [x] Builtin pattern matching tests (positive/negative cases for each pattern)
- [x] `LogMasker` masking behavior tests
- [x] `MaskingConfig` validation tests
- [x] Pattern compilation and caching verification
- [x] Performance benchmarks (SLO compliance)
- [x] Consistent hashing: same value → same hash (all patterns)
- [x] Consistent hashing: different values → different hashes
- [x] Consistent hashing: multiple same values in one line → same hash
- [x] MAC address format normalization (different formats → same hash)

### Integration Tests

- [x] Query execution with masking enabled
- [x] Backward compatibility with masking disabled
- [x] Context-based masker injection verification

## Non-Goals (Intentionally Not Supported)

- Custom regex patterns (for security and simplicity)
- Tool-level masking parameter (AI cannot modify settings)
- Local phone number formats (E.164 international format only)
- Label value masking (log body only)

### DIFF
--- a/tools/loki_masking.go
+++ b/tools/loki_masking.go
@@ -93,10 +93,9 @@ func ValidateMaskingConfig(config *MaskingConfig) error {
 }
 
 type compiledPattern struct {
-	regex       *regexp.Regexp
-	replacement string
-	id          string
-	normalizer  func(string) string // if non-nil, enables consistent hashing for same-value identification
+	regex      *regexp.Regexp
+	id         string
+	normalizer func(string) string // if non-nil, normalizes matched value before hashing (e.g., MAC address format normalization)
 }
 
 // normalizeMACAddress strips all separators and lowercases the MAC address
@@ -137,14 +136,10 @@ func NewLogMasker(config *MaskingConfig) (*LogMasker, error) {
 			return nil, err
 		}
 
-		// Fixed replacement format: [MASKED:<pattern_id>]
-		replacement := fmt.Sprintf("[MASKED:%s]", id)
-
 		masker.patterns = append(masker.patterns, &compiledPattern{
-			regex:       regex,
-			replacement: replacement,
-			id:          id,
-			normalizer:  patternNormalizers[BuiltinPatternID(id)],
+			regex:      regex,
+			id:         id,
+			normalizer: patternNormalizers[BuiltinPatternID(id)],
 		})
 	}
 
@@ -176,15 +171,14 @@ func (m *LogMasker) maskLine(line string) string {
 	}
 
 	for _, pattern := range m.patterns {
-		if pattern.normalizer != nil {
-			line = pattern.regex.ReplaceAllStringFunc(line, func(match string) string {
-				normalized := pattern.normalizer(match)
-				h := sha256.Sum256([]byte(normalized))
-				return fmt.Sprintf("[MASKED:%s:%s]", pattern.id, hex.EncodeToString(h[:4]))
-			})
-		} else {
-			line = pattern.regex.ReplaceAllLiteralString(line, pattern.replacement)
-		}
+		line = pattern.regex.ReplaceAllStringFunc(line, func(match string) string {
+			input := match
+			if pattern.normalizer != nil {
+				input = pattern.normalizer(input)
+			}
+			h := sha256.Sum256([]byte(input))
+			return fmt.Sprintf("[MASKED:%s:%s]", pattern.id, hex.EncodeToString(h[:4]))
+		})
 	}
 
 	return line
--- a/tools/loki_masking_unit_test.go
+++ b/tools/loki_masking_unit_test.go
@@ -668,55 +668,41 @@ func TestLogMasker_MaskEntries_BuiltinPatterns(t *testing.T) {
 		name     string
 		patterns []string
 		input    string
-		expected string
 	}{
 		{
 			name:     "email pattern",
 			patterns: []string{"email"},
 			input:    "User login: user@example.com",
-			expected: "User login: [MASKED:email]",
 		},
 		{
 			name:     "phone pattern (E.164)",
 			patterns: []string{"phone"},
 			input:    "Contact: +819012345678",
-			expected: "Contact: [MASKED:phone]",
 		},
 		{
 			name:     "credit card pattern",
 			patterns: []string{"credit_card"},
 			input:    "Payment with 4111-1111-1111-1111",
-			expected: "Payment with [MASKED:credit_card]",
 		},
 		{
 			name:     "ip address pattern (IPv4)",
 			patterns: []string{"ip_address"},
 			input:    "Server IP: 192.168.1.100",
-			expected: "Server IP: [MASKED:ip_address]",
 		},
 		{
 			name:     "mac address pattern",
 			patterns: []string{"mac_address"},
 			input:    "Device MAC: 00:1A:2B:3C:4D:5E",
-			expected: "", // MAC uses consistent hashing, checked separately
 		},
 		{
 			name:     "api key pattern",
 			patterns: []string{"api_key"},
 			input:    "api_key=abc123def456ghi789jkl",
-			expected: "[MASKED:api_key]",
 		},
 		{
 			name:     "jwt token pattern",
 			patterns: []string{"jwt_token"},
 			input:    "Token: eyJhbGciOiJIUzI1NiJ9.eyJzdWIiOiIxIn0.signature",
-			expected: "Token: [MASKED:jwt_token]",
-		},
-		{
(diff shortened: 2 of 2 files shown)


gitdoluquita:
Hey @Marie673, first of all, sorry for the very late reply on this, and thank you for the work here.

After discussing internally, we don't think the MCP server is the right layer for this kind of masking. The main concerns are:

1. This shouldn't be an MCP-only feature. Log masking isn't specific to MCP — it's a general concern for anyone querying Loki, whether through Grafana Explore, the API, or any other client. Building it into the MCP server means it only protects one access path, which doesn't really solve the problem.

2. Ingestion-time is the right place for this. Loki's ecosystem already has good primitives for redacting sensitive data before it ever hits storage. Alloy's [loki.secretfilter](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.secretfilter/) component and the [replace pipeline stage](https://grafana.com/docs/alloy/latest/reference/components/loki/loki.process/#stagereplace-block) are both examples that handle this at the collection layer.

3. Regex-based PII detection is a moving target. Maintaining patterns for every possible form of sensitive data (locale-specific phone formats, evolving token formats, etc.) is an ongoing burden that's hard to get right and easy to get wrong, giving a false sense of security when patterns inevitably miss something and adding more maintenance work to this repo.

Again, really appreciate you putting this together. It's clear you thought through the design carefully. We just think this is better solved at a different layer.

Marie673:
@gitdoluquita Thank you for the review and thoughtful feedback. No worries about the delay at all.

I understand that the solutions you mentioned (`loki.secretfilter`, `replace` pipeline stage) are designed to prevent sensitive data from being stored in Loki in the first place. However, the intent of this PR is different from that.

**The goal of this PR is to prevent sensitive information from being passed to LLMs/Agents.**

Specifically:

- In practice, there are cases where human operators need raw data in Loki for debugging purposes, making it impractical to mask all logs at ingestion time.
- Even if ingestion-time masking is adopted, it has no effect on historical logs already stored in Loki.
- The MCP server is the entry point for passing data to LLMs. Masking at this layer enables selective control — data remains visible to humans but is hidden from LLMs.

In other words, ingestion-time masking and MCP-layer masking **protect against different things**. The former hides data from all clients; the latter hides data specifically from LLMs.

As LLM usage grows, I believe controlling what information is passed to Agents is a reasonable responsibility for the MCP server layer. I would appreciate it if you could reconsider with this distinction in mind.

Issue #498, opened 2026-01-21
Feature Request: Add `query_loki_patterns` Tool


loafoe:
## Summary

Add a new MCP tool `query_loki_patterns` to expose Loki's pattern detection endpoint (`GET /loki/api/v1/patterns`) through the Grafana MCP server.

## Motivation

Loki's patterns API automatically detects common log structures and aggregates occurrence counts over time. This is valuable for:

- **Log analysis**: Quickly understand the structure of logs without manual inspection
- **Anomaly detection**: Identify unusual patterns or spikes in specific log types
- **Troubleshooting**: Find recurring error patterns across services
- **Capacity planning**: Understand log volume distribution by pattern

Currently, users can query logs and stats via the MCP server, but cannot leverage Loki's built-in pattern detection capabilities.

## Proposed Solution

Add `query_loki_patterns` tool with the following parameters:

| Parameter | Required | Description |
|-----------|----------|-------------|
| `datasourceUid` | Yes | The UID of the Loki datasource |
| `logql` | Yes | A LogQL stream selector (e.g., `{job="nginx"}`) |
| `startRfc3339` | No | Start time in RFC3339 format (defaults to 1 hour ago) |
| `endRfc3339` | No | End time in RFC3339 format (defaults to now) |
| `step` | No | Query resolution step (e.g., `5m`) |

### Example Usage

```
query_loki_patterns(
  datasourceUid: "loki",
  logql: "{service_name=\"mimir\"}",
  startRfc3339: "2026-01-21T21:00:00Z",
  endRfc3339: "2026-01-21T22:00:00Z",
  step: "5m"
)
```

### Example Response

```json
[
  {
    "pattern": "ts=<_> caller=grpc_logging.go:66 level=info method=/cortex.Ingester/Push duration=<_> msg=gRPC",
    "totalCount": 426,
  }
]
```

## Design Fit

This tool fits naturally with the existing Loki tools:

- Uses the same `newLokiClient` → same datasource proxy → same `datasources:query` RBAC model
- Reuses existing time range helpers (`addTimeRangeParams`, `getDefaultTimeRange`)
- Follows the same parameter naming conventions (`datasourceUid`, `startRfc3339`, `endRfc3339`)

## Requirements

- Loki must have `pattern_ingester.enabled: true` configured
- User must have `datasources:query` permission on the target Loki datasource

PR #499, opened 2026-01-21
feat: add Loki patterns query_loki_patterns tool

loafoe:
Add new tool to query detected log patterns from Loki datasources.

This allows us to leverage the pattern matching done on the Loki server which can drastically reduce tool calls and token count for LLMs pattern matching.

Implements #498

### DIFF
--- a/README.md
+++ b/README.md
@@ -50,6 +50,7 @@ The dashboard tools now include several strategies to manage context window usag
 
 - **Query Loki logs and metrics:** Run both log queries and metric queries using LogQL against Loki datasources.
 - **Query Loki metadata:** Retrieve label names, label values, and stream statistics from Loki datasources.
+- **Query Loki patterns:** Retrieve log patterns detected by Loki to identify common log structures and anomalies.
 
 ### Incidents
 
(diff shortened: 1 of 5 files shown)


sd2k:
Nice, thanks for this PR. I think it is worth updating the Loki config in the docker-compose setup so we can see this working in tests (or at least returning an empty result). Other than that, could you update the README with the new tool near the other Loki tools?

loafoe:
> Nice, thanks for this PR. I think it is worth updating the Loki config in the docker-compose setup so we can see this working in tests (or at least returning an empty result). Other than that, could you update the README with the new tool near the other Loki tools?

Yes, already working on enabling the integration test and will update the README 👍🏻

loafoe:
@sd2k updated the integration test to include the `query_loki_patterns` test, also updated the README accordingly.

README.md
sd2k, line 1: There's also a table of Tools further down this doc that needs to be updated. Sorry, the README needs a rework!

loafoe: Ah, missed that one, updated!

PR #505, opened 2026-01-23
fix: centralize tool error handling

### LINKED ISSUE #504: Prometheus query errors returned as JSON-RPC error (-32603) instead of tools/call isError=true result
I think this is generic behavior of the grafana-mcp, not specific to query_prometheus.

When calling Grafana MCP Prometheus instant query tool with an invalid PromQL query, the server returns a JSON-RPC top-level error response (code: -32603), which causes MCP clients to throw McpError and prevents the caller/LLM from recovering.

## Expected behavior
For tool execution errors (invalid PromQL / Prometheus “bad_data”), MCP server should return a successful JSON-RPC response with:

* `result.isError: true`
* error details in `result.content`

This follows MCP guidance that tool failures should be returned as normal tool results, not protocol errors.

## Actual behavior
It returns:

* JSON-RPC `error.code = -32603`
* error message contains Prometheus parse error

## Environment

MCP client: n8n + @modelcontextprotocol/sdk@1.24.0

Transport: StreamableHTTPClientTransport

grafana-mcp version / docker image tag: `docker.io/mcp/grafana:latest@sha256:bd85b2644ea409ce8222088a20fd01b132047f3dcb85240bbce2b2030872a4f7`

## Why this matters
This makes tool errors unrecoverable in many MCP clients because JSON-RPC errors abort the request instead of letting the LLM/workflow handle the failure and retry with corrected input.

igor-raits:
Previously, tool execution errors were returned as JSON-RPC error -32603 (Internal Error), making it difficult for LLM clients to distinguish between user input errors (retryable with corrections) and actual system failures.

This change modifies the centralized ConvertTool function to automatically convert all tool execution errors to CallToolResult with isError=true, following MCP specification guidance. This allows LLM clients to see error details in their context and retry with corrected input.

Changes:
- Add HardError escape hatch for rare protocol-level failures (missing auth, etc.)
- Modify ConvertTool in tools.go to convert errors to CallToolResult{IsError: true}
- Update tests to expect IsError=true instead of Go errors

This applies to all 40+ tools automatically without requiring individual tool modifications.

Fixes #504


sd2k:
Hi @igor-raits, thanks for the PR and issue. I wonder if it might be better to change our [conversion](https://github.com/grafana/mcp-grafana/blob/1c773ebb3f795ac3fa4bed42dc977cf80cbc641a/tools.go#L141-L159) so that all errors are converted to `CallToolResult` with `IsError = true`, instead; otherwise we will have to go through all our tools and do this manually. I think that is the expected behaviour. We can consider adding a separate error type for a hard failure which tools can choose to use if they like, but I don't think it's a common case.

igor-raits:
@sd2k did you think about something like … https://github.com/grafana/mcp-grafana/pull/505/commits/c72eb0c73e45052985ff9810325d655bc379f850 ?

tools/loki.go
sd2k, line 1: I don't think we need the changes to this or the Prometheus files any more.

igor-raits: Let me fix those

igor-raits:
@sd2k done!

igor-raits:
Looks like I forgot one test - fixed now.

sd2k:
Thanks! Could you rename the PR & update the description to match so the final commit message is accurate? then I think we can merge.

igor-raits:
@sd2k done.

Issue #334, opened 2025-10-16
update alert rule


einatmeidan:
I would like the ability to edit an existing alert rule directly from MCP.
This would make it easier to update conditions, thresholds, or other rule settings without needing to recreate the rule or make changes manually.
For example:
- add, remove, or update labels
- change the alert rule name
- adjust the pending time before an alert is triggered

sierikov:
I had a quick look on a codebase. I can try to implement all or at least some of handlers one-by-one. 

My vision is adding to `tools/alerting.go` following handlers in order:

- `rename_alert_rule` with params `uid string` and `newTitle string` - to change the alert rule name. And must be the easiest one. 
- `set_alert_rule_pending_duration` with params: `uid string`, `duration strfmt.Duration` - this one can be tricky, but maybe I can get it done only updating `ProvisionedAlertRule.For` field. (If it is tricky - I will do next 2 instead, and move it to the end of the list)
- `add_alert_rule_labels` with params `uid string` and `labelsToAdd map[string]string` - to add labels to the rule. 
- `remove_alert_rule_labels` with params: `uid string`, `labelKeys []string` - to remove labels from the rule.

For the last two, we should probably think about validation. If we don’t, we could just pass the input to the instance, and it’ll handle the validation for us, which would save us from duplicating the same logic.  But if we want to fail on MCP level we need define rules for labels.
What do you think?

sd2k:
@sierikov nice, thanks for looking into that. My hunch is that we should avoid adding too many different tools - we already have quite a lot - so perhaps just an 'update alert rule' tool would be sufficient.

The [docs for the alerting HTTP API](https://grafana.com/docs/grafana/latest/alerting/set-up/provision-alerting-resources/http-api-provisioning/) will certainly come in handy, probably the `RoutePutAlertRule` route. We can probably just rely on the endpoint itself to do validation as long as it returns suitable error messages for us to return to the model, at least to start.

pkalsi97:
@sd2k will adding tools for 
1. rule group endpoints (GET and PUT)
2. contact point tools: create (POST), update (PUT), delete (DELETE)

would satisfy the scope of this issue ? If so I'll work on this promptly.

sd2k:
Done in https://github.com/grafana/mcp-grafana/pull/292.

PR #509, opened 2026-01-27
feat: add notification_settings support to create/update alert rule tools

meisner:
## Summary

Adds `notification_settings` parameter support to the `create_alert_rule` and `update_alert_rule` MCP tools, allowing clients to configure which contact point (receiver) an alert rule routes to directly.

## Problem

Currently there's no way to set the notification receiver on an alert rule via MCP. Users must manually configure this in the Grafana UI after creating/updating a rule.

## Solution

Added `NotificationSettings *models.AlertRuleNotificationSettings` to both `CreateAlertRuleParams` and `UpdateAlertRuleParams`, and wire it through to the `models.ProvisionedAlertRule` passed to the Grafana provisioning API.

The `grafana-openapi-client-go` model already has this field — this change just exposes it through the MCP tools.

## Usage

```json
{
  "notification_settings": {
    "receiver": "My Slack Channel"
  }
}
```

## Testing

Tested end-to-end: built patched binary locally, created an alert rule with `notification_settings.receiver` set, verified via `get_alert_rule_by_uid` that the setting persisted correctly.

Relates to #334

### DIFF
--- a/tools/alerting.go
+++ b/tools/alerting.go
@@ -581,6 +581,7 @@ type CreateAlertRuleParams struct {
 	Labels            map[string]string    `json:"labels,omitempty" jsonschema:"description=Optional labels"`
 	UID               *string              `json:"uid,omitempty" jsonschema:"description=Optional UID for the alert rule"`
 	OrgID             int64                `json:"orgID" jsonschema:"required,description=The organization ID"`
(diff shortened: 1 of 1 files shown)


sd2k:
Closing, outdated since #619.

PR #510, opened 2026-01-27
fix(loki): support metric queries with proper label parsing

### LINKED ISSUE #368: Loki metric queries return null labels for aggregate groupings

igor-raits:
Fix Loki metric queries returning null labels by properly handling
the different response formats from Loki's API:

- Add queryType parameter ("instant" or "range") to route queries
  to the appropriate Loki endpoint (/query vs /query_range)
- Add stepSeconds parameter for range metric queries
- Parse results based on resultType field (streams/vector/matrix)
- Use "metric" field for labels in metric queries (not "stream")
- Add MetricValue type and Values field for range metric results
- Add integration tests for instant and range metric queries

Fixes #368

### DIFF
--- a/tools/loki.go
+++ b/tools/loki.go
@@ -306,19 +306,34 @@ var ListLokiLabelValues = mcpgrafana.MustTool(
 	mcp.WithReadOnlyHintAnnotation(true),
 )
 
-// LogStream represents a stream of log entries from Loki
-type LogStream struct {
+// LokiLogStream represents a stream of log entries from Loki (resultType: "streams")
+// Labels are in the "stream" field, timestamps are nanosecond strings
+type LokiLogStream struct {
 	Stream map[string]string   `json:"stream"`
-	Values [][]json.RawMessage `json:"values"` // [timestamp, value] where value can be string or number
+	Values [][]json.RawMessage `json:"values"` // [[ts_nanos_string, log_line], ...]
 }
 
-// QueryRangeResponse represents the response from Loki's query_range API
-type QueryRangeResponse struct {
+// LokiMetricSample represents a metric sample from Loki (resultType: "vector" or "matrix")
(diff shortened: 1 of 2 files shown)


igor-raits:
I have tested this on

```json
{
  "datasourceUid": "LOKI_PRODUCTION",
  "logql": "sum by(cluster_name, host) (count_over_time({sourcetype=\"audit\"} [1h]))",
  "queryType": "instant",
  "startRfc3339": "2026-01-27T09:33:45+01:00",
  "endRfc3339": "2026-01-27T10:33:44+01:00"
}
```

and works just fine :)

igor-raits:
@sd2k I have hopefully addressed your concerns in linked PR.

PR #516, opened 2026-01-28
fix(tests): DeepEval MCP evaluation for e2e tests

ioanarm:
E2E tests now use DeepEval MCP evaluation (MCPUseMetric + GEval).

Before we were testing a fixed tool path: the LLM had to call specific tools in a specific order with specific args (e.g. list_datasources then query_loki_logs with that UID). Tests failed if the model used a different path.

Now we are testing that the right tools were used (and key args checked where it matters) and that the final answer is good. DeepEval (MCPUseMetric + GEval) scores tool use and output quality. We don’t require one exact sequence; the model can discover things via different tools. 

This approach is a bit more relaxed: Different models take different valid paths, so enforcing a single sequence caused flakiness even when the outcome was correct. We care more about “did the user get a good result?” than “did the model follow our preferred path?” So we keep outcome and important-arg checks and accept varied strategies, with lower MCPUseMetric thresholds where models often do a bit of exploration first.

Loki tests are still flickering.

### DIFF
--- a/tests/loki_test.py
+++ b/tests/loki_test.py
@@ -44,9 +44,7 @@ async def test_loki_container_labels(
     mcp_transport: str,
 ):
     prompt = (
-        "List the values for the label 'container' for the last 10 minutes using Loki. "
-        "If you don't have a Loki datasource UID, use list_datasources with type 'loki' first to get one, "
-        "then use list_loki_label_values with that datasourceUid and labelName 'container'. Return the list of container names."
+        "List the values for the label 'container' for the last 10 minutes from the Loki datasource."
     )
     final_content, tools_called, mcp_server = await run_llm_tool_loop(
         model, mcp_client, mcp_transport, prompt
--- a/tests/utils.py
+++ b/tests/utils.py
@@ -1,3 +1,4 @@
+import json
 import os
 from typing import List, Optional, Union
 
@@ -92,7 +93,7 @@ async def run_llm_tool_loop(
         for tool_call in response.choices[0].message.tool_calls:
             tool_name = tool_call.function.name
             args = (
-                __import__("json").loads(tool_call.function.arguments)
+                json.loads(tool_call.function.arguments)
                 if tool_call.function.arguments
                 else {}
             )


tests/navigation_test.py
```diff
@@ -2,171 +2,241 @@
...
+    assert_expected_tools_called(tools_called, "generate_deeplink")
+    test_case = make_test_case(prompt, final_content, mcp_server, tools_called)
+
+    mcp_metric = MCPUseMetric(threshold=MCP_EVAL_THRESHOLD)
+    output_metric = GEval(
+        name="OutputQuality",
+        criteria="Does the response contain a URL with custom query parameters?",
+        evaluation_params=[LLMTestCaseParams.INPUT, LLMTestCaseParams.ACTUAL_OUTPUT],
+        threshold=MCP_EVAL_THRESHOLD,
+    )
```
sd2k: We can validate that deterministically, not sure we need the LLM eval!

tests/utils.py
```diff
@@ -99,111 +24,76 @@ def convert_tool(tool: Tool) -> dict:
...
+    mcp_server: MCPServer,
+    tools_called: List[MCPToolCall],
+) -> LLMTestCase:
+    """Build LLMTestCase for DeepEval from prompt, final output, server, and recorded tool calls."""
+    return LLMTestCase(
+        input=input_text,
+        actual_output=actual_output,
+        mcp_servers=[mcp_server],
+        mcp_tools_called=tools_called,
     )
```
sd2k: not sure we get much from this function

tools/oncall.go
```diff
@@ -400,7 +400,7 @@ func listOnCallUsers(ctx context.Context, args ListOnCallUsersParams) ([]*aapi.U
 
 var ListOnCallUsers = mcpgrafana.MustTool(
 	"list_oncall_users",
-	"List users from Grafana OnCall. Can retrieve all users, a specific user by ID, or filter by username. Returns a list of user objects with their details. Supports pagination.",
+	"List users from Grafana OnCall (IRM plugin). Can retrieve all on call users, a specific user by ID, or filter by username. Returns a list of user objects with their details. Supports pagination.",
```
sd2k: The 'IRM plugin' part is a bit incongruous, perhaps it's better to explain that these users are different from Grafana users.

Also `all on call users` makes it sound like it will only return users currently on call!

tests/navigation_test.py
```diff
@@ -2,171 +2,241 @@
...
+            "Expected LLM to call generate_deeplink. No tool calls."
+        )
+
+    final_response = await acompletion(model=model, messages=messages, tools=tools)
+    final_content = final_response.choices[0].message.content or ""
+
+    assert "var-datasource=prometheus" in final_content, f"Expected var-datasource=prometheus in URL, got: {final_content}"
+    assert "refresh=30s" in final_content, f"Expected refresh=30s in URL, got: {final_content}"
+
+    assert_expected_tools_called(tools_called, "generate_deeplink")
```
sd2k: It seems like lots of the tests repeat some very similar logic to this, could we factor it out somehow? Perhaps that puts us back where we were before this PR though

ioanarm: yeah, we need to abstract this somehow.

tests/utils.py
ioanarm: oh the first version wasn't that 😭 didn't clean it up.

tests/loki_test.py
```diff
@@ -1,95 +1,67 @@
...
-        model, messages, tools, mcp_client, "list_datasources"
+async def test_loki_container_labels(
+    model: str,
+    mcp_client: ClientSession,
+    mcp_transport: str,
+):
+    prompt = (
+        "List the values for the label 'container' for the last 10 minutes using Loki. "
+        "If you don't have a Loki datasource UID, use list_datasources with type 'loki' first to get one, "
+        "then use list_loki_label_values with that datasourceUid and labelName 'container'. Return the list of container names."
```
sd2k: this is a bit unrealistically specific - real world prompts won't be anything like this at all. maybe we should make it more vague?

tests/utils.py
```diff
@@ -99,111 +27,132 @@ def convert_tool(tool: Tool) -> dict:
...
 
-    tool_call = response.choices[0].message.tool_calls[0]
-    assert tool_call.function.name == expected_tool_name, (
-        f"Expected {expected_tool_name} tool, got {tool_call.function.name}"
-    )
+    while response.choices and response.choices[0].message.tool_calls:
+        for tool_call in response.choices[0].message.tool_calls:
+            tool_name = tool_call.function.name
+            args = (
+                __import__("json").loads(tool_call.function.arguments)
```
sd2k: 🤔

sd2k (commented):
Generally LGTM, just a bit of weirdness!

tests/utils.py
ioanarm: cursor added it.. on it

tests/loki_test.py
ioanarm: I thought I updated it 😏

PR #522, opened 2026-01-30
feat: custom http headers support

ntibi:
Add ability to inject arbitrary HTTP headers into all Grafana API requests using the GRAFANA_EXTRA_HEADERS environment variable. The value should be a JSON object mapping header names to values.

This is useful for:
- Custom authentication schemes beyond API keys
- Integration with reverse proxies requiring specific headers

Changes:
- Standardize all HTTP clients to use consistent transport wrapping
- Add ExtraHeadersRoundTripper for consistent header injection
- Add BuildTransport() to consolidate transport creation logic
- Update all HTTP clients to use the new transport builder
- Add GRAFANA_EXTRA_HEADERS to server.json schema
- Add documentation and tests

### DIFF
--- a/tools/pyroscope.go
+++ b/tools/pyroscope.go
@@ -287,9 +287,9 @@ func fetchPyroscopeProfile(ctx context.Context, args FetchPyroscopeProfileParams
 func newPyroscopeClient(ctx context.Context, uid string) (*pyroscopeClient, error) {
 	cfg := mcpgrafana.GrafanaConfigFromContext(ctx)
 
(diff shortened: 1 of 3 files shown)


sd2k:
This seems reasonable, thanks for the PR. Can I ask what your specific use case is?

ntibi:
@sd2k My use case is accessing a grafana instance behind Cloudflare Access
So i need every outbound request to have some specific headers (`CF-Access-Client-Secret` + `CF-Access-Client-Id`)

PR #532, opened 2026-02-02
fix(ci): push Docker images with version tags, not just latest

franco-viotti:
## Summary

- The `docker.yml` workflow was configured to push versioned tags (e.g., `0.9.0`) alongside `latest`, but the `push.tags` trigger was never actually firing
- Looking at workflow run history, only the `goreleaser` workflow runs on tag pushes - the docker workflow never triggered
- This resulted in only `latest` being pushed to Docker Hub, with no version tags

## Changes

- Replace unreliable `push.tags` trigger with `workflow_run` trigger that fires after `goreleaser` completes
- Update tag detection logic to use `workflow_run` event data instead of `GITHUB_REF`
- Fix checkout refs to use the correct commit SHA from the triggering workflow
- Update `mcp-registry` job conditions to work with `workflow_run` events

## How it works now

When a tag like `v0.10.0` is pushed:
1. Goreleaser workflow runs and creates the GitHub release (this already works)
2. Docker workflow triggers automatically after goreleaser succeeds
3. Docker images are pushed with both `0.10.0` and `latest` tags
4. MCP registry is updated with the versioned image reference

## Test plan

- [ ] Verify workflow syntax is valid
- [ ] Test with a new release tag to confirm both version and latest tags are pushed

### DIFF
--- a/.github/workflows/docker.yml
+++ b/.github/workflows/docker.yml
@@ -4,13 +4,15 @@ on:
   push:
     branches:
       - main
-    tags:
-      - "v*.*.*"
   pull_request:
     branches:
       - main
-  release:
-    types: [published]
+  # Trigger after goreleaser completes on tags - this is more reliable than
+  # push.tags which wasn't triggering consistently
+  workflow_run:
+    workflows: ["goreleaser"]
+    types:
+      - completed
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


sd2k:
The goreleaser workflow isn't the one that handles the Docker images actually, it's the 'Build and push Docker image' workflow - it ran successfully [here](https://github.com/grafana/mcp-grafana/actions/runs/21441446865) for the latest release (v0.9.0) and pushed both 'latest' and '0.9.0' tags already, and I can see the tags in the Docker hub (https://hub.docker.com/r/grafana/mcp-grafana/tags)

What's the problem you're seeing?

franco-viotti:
> The goreleaser workflow isn't the one that handles the Docker images actually, it's the 'Build and push Docker image' workflow - it ran successfully [here](https://github.com/grafana/mcp-grafana/actions/runs/21441446865) for the latest release (v0.9.0) and pushed both 'latest' and '0.9.0' tags already, and I can see the tags in the Docker hub (https://hub.docker.com/r/grafana/mcp-grafana/tags)
> 
> What's the problem you're seeing?

I was looking at the wrong docker hub repo, sorry!

sd2k:
Ah yeah the mcp/grafana Docker repo is managed by Docker and seemingly unmaintained :disappointed:  See e.g. https://github.com/docker/mcp-registry/issues/757. No worries!

PR #564, opened 2026-02-10
feat(dashboard): support remove operation on array elements by index

tangcent:
## Description

Adds support for removing array elements by index in the `update_dashboard` patch operations.

### Motivation

While using the MCP tool to clean up a Grafana dashboard, I needed to remove several redundant panel rows that were already covered in another dashboard. The tool returned:

```
cannot remove array element panels[11] (not supported)
```

The only workaround was to fetch the entire panels array, manually filter out the unwanted elements, and send a `replace` with the full filtered array as the value. For a dashboard with many large panels, this meant the LLM had to hold all panel definitions in context just to remove three of them — a massive context window cost compared to a simple `{"op": "remove", "path": "$.panels[11]"}`.

The LLM also attempted to work around the limitation by generating same-index patterns like `[11, 11, 11]`, expecting each remove to shift the array — a reasonable but incorrect assumption that silently targets wrong elements.

### The concern: index shifting

Removing an array element shifts all subsequent indices. In a multi-operation batch applied sequentially, this means a second `remove` can target the wrong element:

```
Before: [A, B, C, D, E]
Remove index 1 → [A, C, D, E]
Remove index 3 → removes E (was D before the shift)
```

### The solution: auto-sort removes descending

Rather than erroring and forcing an extra LLM round trip, we automatically sort multiple array element removes on the same array from **highest to lowest** index via `sortArrayRemovesDescending`. This means callers can pass removes in any order (e.g. `[2, 1, 3]`) and they'll be reordered to `[3, 2, 1]` before applying. Duplicate indices on the same array are still rejected as they're almost certainly an LLM mistake.

The tool description for `update_dashboard` now also mentions the remove-by-index capability and the auto-reordering behavior.

### Changes

**`tools/dashboard.go`**
- `removeAtSegment()`: handle `segment.IsArray` with `slices.Delete` instead of returning "not supported" error
- `sortArrayRemovesDescending()`: auto-sort multi-remove operations from highest to lowest index; reject duplicate indices
- `updateDashboardWithPatches()`: call sort before processing operations
- `UpdateDashboard` tool description: mention array element removal and auto-reordering

**`tools/dashboard_unit_test.go`** (new)
- 5 property-based tests via `testing/quick` (200 iterations each)
- Edge case unit tests for `removeAtSegment`
- 7 test cases for `sortArrayRemovesDescending` including ascending reordering, mixed ops, and same-index rejection

**`tools/dashboard_test.go`**
- Integration test: append a panel then remove it by index

### Usage

```json
{"op": "remove", "path": "$.panels[2]"}
```

Multiple removes (highest index first):
```json
[
  {"op": "remove", "path": "$.panels[4]"},
  {"op": "remove", "path": "$.panels[2]"},
  {"op": "remove", "path": "$.panels[0]"}
]
```

---

Happy to adjust anything based on feedback. Feel free to close this if it doesn't align with the project's direction.

---

> [!NOTE]
> **Medium Risk**
> Touches dashboard mutation logic by changing how patch `remove` operations are applied and reordered; mistakes could delete the wrong elements, but the change is guarded with validation and extensive tests.
> 
> **Overview**
> `update_dashboard` patch operations now support removing array elements by index (e.g., `$.panels[2]`) instead of requiring callers to `replace` the full array.
> 
> To prevent multi-remove index shifting, array-element `remove` ops targeting the same array are automatically reordered from highest to lowest index and duplicate index removals are rejected. The tool description is updated accordingly, and new unit/property tests plus an integration test cover array removal, nested paths, bounds errors, and ordering behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 02b6c34f9168e46b59136b83b0ace08acd47aa73. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tools/dashboard.go
+++ b/tools/dashboard.go
@@ -5,6 +5,8 @@ import (
 	"encoding/json"
 	"fmt"
 	"regexp"
+	"slices"
+	"sort"
 	"strconv"
 	"strings"
 
@@ -69,6 +71,13 @@ func updateDashboard(ctx context.Context, args UpdateDashboardParams) (*models.P
 
 // updateDashboardWithPatches applies patch operations to an existing dashboard
 func updateDashboardWithPatches(ctx context.Context, args UpdateDashboardParams) (*models.PostDashboardOKBody, error) {
+	// Sort array element remove operations from highest to lowest index to avoid index-shifting issues
+	sortedOps, err := sortArrayRemovesDescending(args.Operations)
+	if err != nil {
+		return nil, err
+	}
+	args.Operations = sortedOps
+
 	// Get the current dashboard
 	dashboard, err := getDashboardByUID(ctx, GetDashboardByUIDParams{UID: args.UID})
 	if err != nil {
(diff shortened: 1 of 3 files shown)


tools/dashboard.go
sd2k: We should mention the new functionality here too 👍

tools/dashboard.go
```diff
@@ -129,6 +133,59 @@ func updateDashboardWithFullJSON(ctx context.Context, args UpdateDashboardParams
 	return dashboard.Payload, nil
 }
 
+// validateArrayRemoveOrder checks that multiple array element removes on the same array
+// are ordered from highest index to lowest. This prevents the index-shifting footgun
+// where removing a lower index first causes subsequent operations to target wrong elements.
+func validateArrayRemoveOrder(operations []PatchOperation) error {
```
sd2k: I'm wondering if rather than validating and erroring we could just always reverse-sort the input array, to avoid another round trip to the LLM. So if the caller passes [2, 1, 3] we just sort it to [3, 2, 1] instead of throwing an error. So this becomes something like `sortArrayRemovesDescending` instead.

tools/dashboard.go
```diff
@@ -546,7 +603,13 @@ func removeAtSegment(current map[string]interface{}, segment JSONPathSegment) er
 	}
 
 	if segment.IsArray {
-		return fmt.Errorf("cannot remove array element %s[%d] (not supported)", segment.Key, segment.Index)
+		arr, err := validateArrayAccess(current, segment)
+		if err != nil {
+			return err
+		}
+		arr = append(arr[:segment.Index], arr[segment.Index+1:]...)
+		current[segment.Key] = arr
```
sd2k: we should use `slices.Delete` here instead

tools/dashboard.go
```diff
@@ -6,7 +6,6 @@ import (
 	"fmt"
 	"regexp"
 	"strconv"
-
```
sd2k, line 10: can you keep this blank line to separate imports please 🙏

tools/dashboard.go
tangcent: Initially I was hesitant to add this to the description since I wasn't sure how often LLMs would actually use remove on array elements — my thinking was we could just guide them via error messages to keep the instruction surface small. But after using the MCP more since posting this PR, I've found removing array elements comes up pretty frequently (at least in my workflows). So yeah, makes sense to mention it upfront. Done.

tools/dashboard.go
tangcent: Had a similar concern about keeping the description lean, but you're right — auto-sorting avoids an unnecessary round trip to the LLM. Changed validateArrayRemoveOrder to sortArrayRemovesDescending that reorders the removes from highest to lowest index instead of erroring. Still rejects duplicate indices on the same array since that's almost certainly a mistake.

tools/dashboard.go
tangcent: Good call, switched to `slices.Delete` — cleaner and more idiomatic. Done.

tools/dashboard.go
tangcent: sorry about that, got it back.

sd2k (approved):
Thanks for addressing the feedback!

PR #566, opened 2026-02-10
feat: Alpine container build

MatteoMori8:
I have recently noticed that the current container image has a few CVEs. These seems to be linked to the image being based on Debian.

This PR is about migrating to Alpine to both make the image more secure and also smaller.

## How did I test it?
- I tried to locally build the image ✅ 
- I tried to run the container and query it ✅ 

---

> [!NOTE]
> **Low Risk**
> Build/runtime base images and provisioning steps change, but no application logic is modified; main risk is Alpine libc/package differences affecting runtime behavior.
> 
> **Overview**
> Switches the container build from Debian-based images to Alpine for both the Go builder (`golang:1.24-alpine3.23`) and runtime (`alpine:3.23.3`) stages.
> 
> Updates package/user setup accordingly by replacing `apt-get` with `apk` for upgrades and `ca-certificates`, and swapping `useradd` for `adduser` while keeping the non-root UID `1000` and the same built binary/entrypoint behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 8c4ae5a7dc17aadddaf2bf27556026ca53e31cf9. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,5 +1,5 @@
 # Build stage
-FROM golang:1.24-alpine3.23 AS builder
+FROM golang:1.24-alpine3.23@sha256:8bee1901f1e530bfb4a7850aa7a479d17ae3a18beb6e09064ed54cfd245b7191 AS builder
 
 # Set the working directory
 WORKDIR /app
@@ -17,7 +17,7 @@ COPY . .
 RUN go build -o mcp-grafana ./cmd/mcp-grafana
 
 # Final stage
-FROM alpine:3.23.3
+FROM alpine:3.23.3@sha256:25109184c71bdad752c8312a8623239686a9a2071e8825f20acb8f2198c3f659
 
 LABEL io.modelcontextprotocol.server.name="io.github.grafana/mcp-grafana"
 


sd2k:
We recently merged https://github.com/grafana/mcp-grafana/pull/551 which should help address issues, we just haven't done a release yet. I'd rather avoid switching the default image to alpine if possible, although we may be able to provide a separate image.

MatteoMori8:
Hey @sd2k thanks for having a look at this 😄 

Point taken. However, may I please ask why you would prefer not to move to Alpine? When it comes to security compliance, I noticed that running on Alpine makes things quite easier so I thought it would be a reasonable move here as well

> although we may be able to provide a separate image.

That would be amazing! Thanks

sd2k:
At this point it's mostly because of backwards compatibility (if someone has an image based on this one and is running further `apt` commands, for example, it will fail). But I'm not at all averse to adding an `:alpine` tag alongside the others!

MatteoMori8:
Right; got it, thanks! Fair enough. 
If you guys have time to add an `:alpine` release as well, that would definitely make our life easier 😄 

Happy to close this PR

MatteoMori8:
@sd2k - Hey! Would you guys benefit from me raising and Issue/feature for the Alpine image development? I am asking because this would make it a bit easier to see if/when progress happen

sd2k:
@MatteoMori8 does https://github.com/grafana/mcp-grafana/pull/568 look good to you?

MatteoMori8:
@sd2k - It does, thank you 🙏 This will make conversations with my security team much easier!

I see that a CI test failed but it looks to be caused by some `clickhouse_test`; unfortunately I am not too sure what that is though

Issue #610, opened 2026-02-28
generate_deeplink: allow configuring host


aviau:
In our setup, `mcp-grafana` speaks to grafana using an internal hostname. This results in links generated by `generate_deeplink` being unusable for our users.

It would be great if there was a setting such as `GRAFANA_PUBLIC_URL` that allowed overriding the host used by `generate_deeplink`!

PR #611, opened 2026-02-28
generateDeepLink: add support for PUBLIC_GRAFANA_URL

aviau:
Closes https://github.com/grafana/mcp-grafana/issues/610

Feel free to edit my PR directly if you have minor comments, I won't be sad :)

> [!NOTE]
> **Low Risk**
> Low risk: adds an optional `PublicURL` used only for link generation while preserving existing `GRAFANA_URL`/`X-Grafana-URL` behavior and fallbacks.
> 
> **Overview**
> Adds support for a separate public-facing Grafana base URL via `GRAFANA_PUBLIC_URL` and `X-Grafana-Public-URL`, storing it as `GrafanaConfig.PublicURL` (header takes precedence over env).
> 
> Updates `generate_deeplink` to build user-facing links from `PublicURL` when provided, otherwise falling back to the configured `URL`, and extends unit tests to cover the new precedence and link behavior.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 0173584ca708ccb7f6a9821023a41081c3107c07. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/mcpgrafana.go
+++ b/mcpgrafana.go
@@ -30,6 +30,7 @@ const (
 	defaultGrafanaURL  = "http://" + defaultGrafanaHost
 
 	grafanaURLEnvVar                 = "GRAFANA_URL"
+	grafanaPublicURLEnvVar           = "GRAFANA_PUBLIC_URL"
 	grafanaServiceAccountTokenEnvVar = "GRAFANA_SERVICE_ACCOUNT_TOKEN"
 	grafanaAPIEnvVar                 = "GRAFANA_API_KEY" // Deprecated: use GRAFANA_SERVICE_ACCOUNT_TOKEN instead
 	grafanaOrgIDEnvVar               = "GRAFANA_ORG_ID"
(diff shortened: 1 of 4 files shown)


sd2k:
This looks like a dupe of https://github.com/grafana/mcp-grafana/pull/408 - I added a comment mentioning my preferred method of handling this in there, would that work for you instead of requiring another setting?

aviau:
Gah. I wish I saw it before, my bad!

> Thanks @greedy52. I'm curious whether, for your use case, it would work if we fetched Grafana's public URL from its settings instead (/api/frontend/settings in the appUrl field of the JSON). If we can avoid adding another config var that would be good.

Yeah for us that would work. Our grafana instance is configured with the proper public URL. It would certainly be a better default than the current behaviour.

aviau:
My quick search does not indicate that this is available in the grafana client right now though?

sd2k:
Yeah we'll probably have to fetch it manually and plumb it through context somehow.

Edit: actually, we should use our own custom `GrafanaClient` struct in context, which embeds the existing `GrafanaHTTPAPI` and also stores the public URL in a field. We can fetch the `/api/frontend/settings` endpoint in `NewGrafanaClient` using a custom HTTP request. This will also help later if we ever need to store extra things like the Grafana version or feature toggles.

sd2k:
Superseded by #664, thanks for the PR!

aviau:
Thanks to you!

PR #636, opened 2026-03-08
fix(prometheus): support GET-only datasources in query tools

### LINKED ISSUE #632: query_prometheus fails with 500 for datasources configured with httpMethod: GET
## Description

`query_prometheus` (and other Prometheus query tools) fail with `server_error: server error: 500` for any Prometheus datasource configured with `httpMethod: GET` in its `jsonData`. Datasources configured with `httpMethod: POST` work fine.

## Root Cause

`promClientFromContext()` in `tools/prometheus.go` creates a standard `prometheus/client_golang` API client pointing at the Grafana datasource resources endpoint:

```go
url := fmt.Sprintf("%s/api/datasources/uid/%s/resources", strings.TrimRight(cfg.URL, "/"), uid)
```

The Prometheus client's `DoGetFallback` method always sends **POST first**, then falls back to GET only on 405 or 501 responses. However, Grafana's datasource resources API returns **500** (not 405) when it receives a POST for a datasource configured with `httpMethod: GET`. This means the GET fallback never triggers.

## Steps to Reproduce

1. Configure a Prometheus datasource in Grafana with `jsonData.httpMethod: "GET"`
2. Use the `query_prometheus` MCP tool to query that datasource
3. Observe: 500 error

The same datasource works fine when queried directly via Grafana's `/api/ds/query` endpoint or the `/api/datasources/proxy/{id}/` endpoint.

## Suggested Fix

In `promClientFromContext()`, read the datasource's `jsonData.httpMethod` setting (the datasource info is already fetched on line 34). If the method is GET, wrap the transport with a `RoundTripper` that converts POST requests to GET (moving form-encoded body params to URL query string). This avoids the unnecessary POST→500→failure cycle entirely.

Alternatively, the client could be configured to always use GET (which both Prometheus and Thanos accept for all query endpoints), avoiding the `DoGetFallback` path altogether.

## Workaround

Change the affected datasources' `httpMethod` from GET to POST. This is safe since the Prometheus HTTP API handles both methods identically, and POST is the modern Grafana default.

## Related

There is also a [related issue in Grafana core ](https://github.com/grafana/grafana/issues/119686)— the resources API should return 405 instead of 500 for method mismatches, which would allow the client's existing fallback to work. Filed separately against `grafana/grafana`.

SergioChan:
## Summary
- read datasource `jsonData.httpMethod` in `promClientFromContext`
- when the datasource is configured for `GET`, wrap the Prometheus client transport to convert outgoing form-encoded POST requests into GET requests against the datasource resources endpoint
- keep existing behavior unchanged for POST-configured datasources
- add focused unit tests for method detection and POST→GET conversion behavior

## Testing
- `go test ./tools -run 'TestDatasourcePrometheusHTTPMethod|TestPrometheusDatasourceHTTPMethodRoundTripper'`

## Related
Fixes #632

---

> [!NOTE]
> **Medium Risk**
> Changes how Prometheus requests are sent for GET-configured datasources by rewriting HTTP methods and bodies, which could affect query compatibility or edge cases around request encoding; POST-configured datasources are unchanged.
> 
> **Overview**
> Prometheus tooling now respects datasource `jsonData.httpMethod`: when configured for `GET`, `promClientFromContext` wraps the Prometheus client transport to transparently convert form-encoded `POST` requests into `GET` requests with equivalent query parameters.
> 
> Adds unit coverage for method detection defaults and for the POST→GET conversion behavior (including non-form POST passthrough and invalid form error handling).
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f199f4b2e3a86652096536922c49e90a32cf39bb. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tools/prometheus.go
+++ b/tools/prometheus.go
@@ -107,6 +107,11 @@ func (p *prometheusDatasourceHTTPMethodRoundTripper) RoundTrip(req *http.Request
 		return p.base.RoundTrip(req)
 	}
 
+	contentType := strings.ToLower(req.Header.Get("Content-Type"))
+	if req.Body != nil && contentType != "" && !strings.HasPrefix(contentType, "application/x-www-form-urlencoded") {
+		return p.base.RoundTrip(req)
+	}
+
 	bodyBytes := []byte{}
 	if req.Body != nil {
 		var err error
@@ -122,19 +127,21 @@ func (p *prometheusDatasourceHTTPMethodRoundTripper) RoundTrip(req *http.Request
 	converted.Body = nil
 	converted.GetBody = nil
 	converted.ContentLength = 0
-	converted.Header = req.Header.Clone()
 	converted.Header.Del("Content-Type")
 	converted.Header.Del("Content-Length")
 
-	query := converted.URL.Query()
-	if values, parseErr := url.ParseQuery(string(bodyBytes)); parseErr == nil {
-		for key, vals := range values {
-			for _, val := range vals {
-				query.Add(key, val)
-			}
-		}
-		converted.URL.RawQuery = query.Encode()
+	values, parseErr := url.ParseQuery(string(bodyBytes))
+	if parseErr != nil {
+		return nil, fmt.Errorf("parsing form body for GET conversion: %w", parseErr)
 	}
+
+	query := converted.URL.Query()
+	for key, vals := range values {
+		for _, val := range vals {
+			query.Add(key, val)
+		}
+	}
+	converted.URL.RawQuery = query.Encode()
 
 	return p.base.RoundTrip(converted)
 }
--- a/tools/prometheus_unit_test.go
+++ b/tools/prometheus_unit_test.go
@@ -32,31 +32,98 @@ type roundTripFunc func(*http.Request) (*http.Response, error)
 func (f roundTripFunc) RoundTrip(req *http.Request) (*http.Response, error) { return f(req) }
 
 func TestPrometheusDatasourceHTTPMethodRoundTripper(t *testing.T) {
-	var gotMethod string
-	var gotQuery string
+	t.Run("converts form POST body into GET query", func(t *testing.T) {
+		var gotMethod string
+		var gotQuery string
 
-	rt := &prometheusDatasourceHTTPMethodRoundTripper{base: roundTripFunc(func(r *http.Request) (*http.Response, error) {
-		gotMethod = r.Method
-		gotQuery = r.URL.RawQuery
-		return &http.Response{
-			StatusCode: http.StatusOK,
-			Body:       io.NopCloser(strings.NewReader("ok")),
-			Header:     make(http.Header),
-		}, nil
-	})}
+		rt := &prometheusDatasourceHTTPMethodRoundTripper{base: roundTripFunc(func(r *http.Request) (*http.Response, error) {
+			gotMethod = r.Method
+			gotQuery = r.URL.RawQuery
+			return &http.Response{
+				StatusCode: http.StatusOK,
+				Body:       io.NopCloser(strings.NewReader("ok")),
+				Header:     make(http.Header),
+			}, nil
+		})}
 
-	req, err := http.NewRequest(http.MethodPost, "http://example.test/api/datasources/uid/abc/resources/api/v1/query", strings.NewReader("query=up&time=1234"))
-	require.NoError(t, err)
-	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+		req, err := http.NewRequest(http.MethodPost, "http://example.test/api/datasources/uid/abc/resources/api/v1/query", strings.NewReader("query=up&time=1234"))
+		require.NoError(t, err)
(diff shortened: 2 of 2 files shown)


tools/prometheus.go
```diff
@@ -79,6 +86,59 @@ func promClientFromContext(ctx context.Context, uid string) (promv1.API, error)
...
+}
+
+type prometheusDatasourceHTTPMethodRoundTripper struct {
+	base http.RoundTripper
+}
+
+func (p *prometheusDatasourceHTTPMethodRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
+	if req.Method != http.MethodPost {
+		return p.base.RoundTrip(req)
+	}
```
sd2k, line 108: `req.Clone()` already deep-clones headers, so this line replaces the clone with an identical clone. You can drop it:

```suggestion
	converted.Header.Del("Content-Type")
	converted.Header.Del("Content-Length")
```

tools/prometheus.go
```diff
@@ -79,6 +86,59 @@ func promClientFromContext(ctx context.Context, uid string) (promv1.API, error)
...
+			return nil, err
+		}
+		_ = req.Body.Close()
+	}
+
+	converted := req.Clone(req.Context())
+	converted.Method = http.MethodGet
+	converted.Body = nil
+	converted.GetBody = nil
+	converted.ContentLength = 0
```
sd2k, line 129: If `url.ParseQuery` fails, the body params are silently dropped and the GET request is sent without them — Prometheus will return a confusing "missing parameter" error rather than something actionable. Better to surface it:

```suggestion
	values, parseErr := url.ParseQuery(string(bodyBytes))
	if parseErr != nil {
		return nil, fmt.Errorf("parsing form body for GET conversion: %%w", parseErr)
	}
	for key, vals := range values {
		for _, val := range vals {
			query.Add(key, val)
		}
	}
	converted.URL.RawQuery = query.Encode()
```

tools/prometheus.go
```diff
@@ -79,6 +86,59 @@ func promClientFromContext(ctx context.Context, uid string) (promv1.API, error)
...
+	data, ok := jsonData.(map[string]any)
+	if !ok {
+		return http.MethodPost
+	}
+	method, _ := data["httpMethod"].(string)
+	if strings.EqualFold(method, http.MethodGet) {
+		return http.MethodGet
+	}
+	return http.MethodPost
+}
```
sd2k, line 99: This will attempt to parse any POST body as URL-encoded form data, regardless of Content-Type. The Prometheus client always sends `application/x-www-form-urlencoded` for query endpoints so it works today, but a defensive guard would be safer:

```go
if req.Body != nil && req.Header.Get("Content-Type") == "application/x-www-form-urlencoded" {
```

#633 had this check — worth carrying over.

tools/prometheus_unit_test.go
```diff
@@ -11,6 +14,51 @@ import (
...
+	req, err := http.NewRequest(http.MethodPost, "http://example.test/api/datasources/uid/abc/resources/api/v1/query", strings.NewReader("query=up&time=1234"))
+	require.NoError(t, err)
+	req.Header.Set("Content-Type", "application/x-www-form-urlencoded")
+
+	resp, err := rt.RoundTrip(req)
+	require.NoError(t, err)
+	defer resp.Body.Close()
+	_, _ = io.ReadAll(resp.Body)
+
+	assert.Equal(t, http.MethodGet, gotMethod)
```
sd2k: Good coverage of the happy path. A couple more cases would round it out:
- GET request passes through unchanged (verifies the early return)
- POST with nil body (no panic)
- POST with non-form-encoded Content-Type (once the guard is added)

SergioChan:
﻿Thanks for the detailed review feedback 鈥?I鈥檓 working through the requested updates and will post a focused patch (or direct answers) shortly.

SergioChan:
Thanks for the detailed review — I pushed a follow-up commit addressing the requested fixes: f199f4b

### What changed
- Return a clear error when form parsing fails during POST→GET conversion (`parsing form body for GET conversion: ...`) instead of silently dropping params.
- Add a content-type guard so only `application/x-www-form-urlencoded` POST bodies are converted; non-form POST requests are now passed through unchanged.
- Remove redundant header re-cloning after `req.Clone()`.
- Extend unit coverage for:
  - non-form POST passthrough,
  - nil-body POST conversion,
  - invalid form body parse error path.

### Validation
- `git diff --check` (pass)
- `go test ./tools -run "TestDatasourcePrometheusHTTPMethod|TestPrometheusDatasourceHTTPMethodRoundTripper"` is currently not runnable in this environment because the repository requires downloading Go 1.24.9 toolchain and the download fails due insufficient disk space.

I’ll also re-run the targeted tests once toolchain space is available.

sd2k:
This has been superseded by #633 and #647 now, the fix is already in. Thanks!

PR #639, opened 2026-03-09
feat(tools): add PromQL support for non-native datasources (stackdriver)

jacobstr:
## Summary

- Adds a `backendPromClient` adapter implementing `promv1.API` that routes PromQL queries through Grafana's `/api/ds/query` endpoint for datasources that support PromQL but don't implement the Prometheus HTTP API at the resource proxy
- `promClientFromContext()` now checks datasource type and routes: native types (prometheus, mimir, cortex, thanos) use the existing resource proxy path; backend types (stackdriver) use the new `/api/ds/query` adapter
- Metadata operations (LabelNames, LabelValues, Metadata) return descriptive errors for non-native datasources
- Maps `stackdriver` → `prometheus` in `normalizeDatasourceType()` so panel queries also work

## Motivation

Google Cloud Monitoring (`stackdriver`) supports PromQL but its Grafana plugin doesn't expose the Prometheus HTTP API at the resource proxy — queries return 502. This change enables `query_prometheus` and `query_prometheus_histogram` to work with stackdriver datasources by routing through Grafana's backend query endpoint instead.

## Test plan

- [x] `go build ./...` compiles
- [x] New unit tests pass: type classification, payload construction, data frame → Matrix/Vector conversion, metadata error messages
- [x] Existing Prometheus unit tests pass (no regressions)
- [ ] Manual test against a stackdriver datasource with a simple PromQL expression
- [ ] Manual test against a native Prometheus datasource to confirm no regression

---

> [!NOTE]
> **Medium Risk**
> Adds a new HTTP query path for PromQL via Grafana’s backend `/api/ds/query`, including response frame parsing and custom auth headers; mistakes could break PromQL queries for stackdriver or subtly change returned timestamps/series. Native Prometheus-family datasources remain on the existing resource-proxy path, limiting blast radius.
> 
> **Overview**
> Adds PromQL support for non-native datasources (currently `stackdriver`) by introducing a `backendPromClient` that implements `promv1.API` and routes `Query`/`QueryRange` through Grafana’s `/api/ds/query`, converting returned data frames into Prometheus `model.Matrix`/`model.Vector`.
> 
> Updates `promClientFromContext` to detect datasource type and choose between the existing native resource-proxy client (`prometheus`/`mimir`/`cortex`/`thanos`) and the new backend query client, with explicit errors for unsupported types and for metadata-style operations on backend-routed datasources. `run_panel_query` is extended to treat `stackdriver` panels as PromQL-capable and to emit stackdriver-specific “empty result” hints, and new unit tests cover type routing, payload building, frame conversion, and unsupported metadata errors.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6a1c2e4aa8d762fc05ae251cfbfbe00f4b083334. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tools/prometheus_backend_test.go
+++ b/tools/prometheus_backend_test.go
@@ -311,6 +311,6 @@ func TestBackendPromClientMetadataErrors(t *testing.T) {
 }
 
 func TestNormalizeDatasourceTypeStackdriver(t *testing.T) {
-	assert.Equal(t, "prometheus", normalizeDatasourceType("stackdriver"))
-	assert.Equal(t, "prometheus", normalizeDatasourceType("Stackdriver"))
+	assert.Equal(t, "stackdriver", normalizeDatasourceType("stackdriver"))
+	assert.Equal(t, "stackdriver", normalizeDatasourceType("Stackdriver"))
 }
--- a/tools/run_panel_query.go
+++ b/tools/run_panel_query.go
@@ -216,7 +216,7 @@ func runSinglePanelQuery(ctx context.Context, params singlePanelQueryParams) (*P
 	var results interface{}
 
 	switch normalizeDatasourceType(datasourceType) {
-	case "prometheus":
+	case "prometheus", "stackdriver":
 		results, err = executePrometheusQuery(ctx, datasourceUID, query, params.Start, params.End)
 	case "loki":
 		results, err = executeLokiQuery(ctx, datasourceUID, query, params.Start, params.End)
@@ -752,7 +752,7 @@ func normalizeDatasourceType(dsType string) string {
 	case lower == "cloudwatch":
 		return "cloudwatch"
 	case lower == "stackdriver":
-		return "prometheus" // stackdriver supports PromQL, route through Prometheus query path
+		return "stackdriver" // stackdriver supports PromQL but needs its own hints
 	case strings.Contains(lower, "clickhouse"):
 		return "clickhouse"
 	default:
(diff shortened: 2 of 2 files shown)


jacobstr:
Addressing the two Bugbot findings:

**1. Misleading empty-result hints for stackdriver (Medium)** — Fixed in 6a1c2e4. `normalizeDatasourceType` now returns `"stackdriver"` instead of `"prometheus"`, and `generatePanelQueryHints` has a dedicated stackdriver case with appropriate guidance (referencing Cloud Monitoring PromQL limitations and the Grafana Explore UI, rather than `list_prometheus_metric_names` which is unsupported for backend-routed datasources). Query routing still correctly sends stackdriver through `executePrometheusQuery` via a combined case.

**2. Duplicated `/api/ds/query` HTTP execution logic (Low)** — Acknowledged. The transport setup and HTTP plumbing in `executeDSQuery` mirrors `executeGrafanaDSQuery`, but the two have meaningful differences: `backendPromClient` adds `ds_type` query params, `X-Datasource-Uid`/`X-Plugin-Id` headers, uses `io.LimitReader`, and deserializes into typed `dsQueryResponse` structs (vs `map[string]interface{}`). It also needs to be a struct method to implement `promv1.API`. Consolidating would require an options/config pattern that adds more abstraction than the duplication warrants at this stage. Happy to revisit if more backend datasource types are added.

Additionally, the latest commit (1a5a100) fixes the core issue that was causing 502/500 errors: the `/api/ds/query` payload now matches what the Grafana UI sends — `queryType: "promQL"`, `promQLQuery` wrapper with `projectName`, and a `timeSeriesList` stub (required by the Cloud Monitoring plugin even for PromQL queries). This was validated against a production stackdriver datasource. Given that [MQL is deprecated](https://docs.cloud.google.com/stackdriver/docs/deprecations/mql) and PromQL is Google's recommended replacement, this backend routing path is the way forward for Cloud Monitoring support.

sd2k:
Thanks for the contribution! I already had a WIP branch for this which tackles it a bit more holistically including support for the metadata tools, and also adds integration tests against a cloud datasource. (the PR is [here](https://github.com/grafana/mcp-grafana/pull/647)). I think it will do the same job as this!

PR #669, opened 2026-03-19
Add OpenSearch support

LucasHild:
Add a new tool for querying OpenSearch. The tool is disabled by default and needs to be enabled using a flag.

I am using the `/api/ds/query` endpoint instead of `/api/datasources/proxy/uid/{uid}/_msearch` to support SigV4 signature for AWS.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new OpenSearch query path (via `/api/ds/query`) and refactors existing Elasticsearch querying behind a shared backend interface, which could affect query behavior and error handling for existing `query_elasticsearch` users.
> 
> **Overview**
> Adds **OpenSearch support** to the existing `query_elasticsearch` tool by routing OpenSearch datasources through Grafana’s `/api/ds/query` (plugin backend) while keeping Elasticsearch on the `_msearch` proxy path.
> 
> Refactors Elasticsearch querying into a new backend abstraction (`esBackend`), adds OpenSearch-specific index-pattern validation and frame-to-document conversion, and updates docs/CLI help to describe Elasticsearch+OpenSearch behavior (including *DSL JSON only for Elasticsearch*).
> 
> Extends the dev/test environment to provision and seed an OpenSearch instance (docker-compose + provisioning + seed script) and adds integration/E2E coverage to ensure OpenSearch queries return real seeded log data.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0d2546f5eaa57a32ddd05be8307d3b949b7dc19b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


ioanarm:
Hello @LucasHild! Thanks for putting this together! We’re trying to avoid growing the MCP tool list when the use case is this close to what we already have. OpenSearch and Elasticsearch are similar under the hood (Lucene, comparable query styles), so `query_opensearch` and `query_elasticsearch`  end up overlapping a lot, and we’d be maintaining two parallel paths long term. 

Maybe we could consider extending query_elasticsearch with an opensearch path. Would you be open to try and extend  `query_elasticsearch` tool instead? Shared bits (auth, limits, time range, document shape) can stay in one place and the request/response handling is the one that needs to diverge.

Thanks again! Happy to discuss more or have another look if you think this is something that you'd like to explore.

sd2k:
@LucasHild Hi! Just checking if you saw @ioanarm 's comment above and have any input on whether we could re-use the existing Elasticsearch tools for Opensearch, too?

LucasHild:
Hi @ioanarm and @sd2k, sorry for the late response.

I kept the two tool calls separate for the following reasons:

1. If a user mentions the OpenSearch datasource in their query (e.g. "Can you query the OpenSearch datasource for the last 10 log entries ..."), the LLM will have an easier time figuring out the right tool call.
2. The underlying implementation differs quite a bit. The OpenSearch tool calls a different endpoint, it uses a different wire format, and the response parsing is different. The overlap is mainly in the tool call parameters and in the document structs.
3. Exposing an additional tool won't lead to context bloat since both tools need to be individually enabled with `--enabled-tools`.

If you prefer merging them, I am happy to adjust the implementation. Just let me know.

sd2k:
@LucasHild Thanks for the clarity. I think I am still leaning towards merging. Addressing your points below:

1. To work around this, perhaps we can rename the `query_elasticsearch` tool to reflect the fact that it also supports OpenSearch, or be very clear in the description that it supports both. And even if the user didn't include the datasource type in their question (e.g. "Query my prod logs datasource") the LLM would already have to call `list_datasources` and find the datasource with the right name so it can get the UID, at which point it would know the type... I just think we could make this work :thinking: 
2. True, but the LLM shouldn't have to care about this; if the tools have the same parameters and the descriptions overlap significantly, it'd be better to have a single tool and separate backends (similar to our `prom_backend.go` file which abstracts over various Prometheus-like data sources such as VictoriaMetrics, Stackdriver, etc).
3. That is true for users who are in charge of running their own MCP server, but if someone is connecting to a remote MCP server (e.g. one run by their org, or a generic hosted one such as mcp.grafana.com), they won't be able to control those flags and limit the tools in the same way, so there will be bloat by default.

Perhaps we could model this in the same way we do for Prometheus and have a Lucene/Elasticsearch backend interface similar to `promBackend`?

tools/es_backend.go
```diff
@@ -0,0 +1,287 @@
...
+
+	client := &http.Client{
+		Transport: mcpgrafana.NewUserAgentTransport(transport),
+	}
+
+	return &elasticsearchBackend{
+		httpClient: client,
+		baseURL:    url,
+	}, nil
+}
```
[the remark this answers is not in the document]
LucasHild: Resolved

LucasHild:
Hey @sd2k, thanks for the input! I just adapted the implementation. Do you mind taking another look?

tools/es_backend_opensearch.go
```diff
@@ -0,0 +1,223 @@
...
+	payload := map[string]interface{}{
+		"queries": []map[string]interface{}{
+			{
+				"refId": "A",
+				"datasource": map[string]interface{}{
+					"uid":  b.datasourceUID,
+					"type": openSearchDatasourceType,
+				},
+				"query":           luceneQuery,
+				"queryType":       "lucene",
```
sd2k, line 120: The tool exposed to the LLM indicates that we support both Lucene syntax and Query DSL, but it looks like we only handle Lucene for OpenSearch. Is that because OpenSearch doesn't accept the Query DSL? If so, we should make that clear in the tool parameter descriptions so the LLM doesn't try to use Query DSL with OpenSearch. Otherwise we could do something similar to ES and pass the DSL down.

tools/es_backend_opensearch.go
```diff
@@ -0,0 +1,223 @@
...
+
+	// Build the Lucene query with index filter.
+	// The OpenSearch plugin uses the datasource's configured index pattern by default,
+	// so we prepend _index:<pattern> to filter by the user-specified index.
+	luceneQuery := query
+	if luceneQuery == "*" || luceneQuery == "" {
+		luceneQuery = "_index:" + index
+	} else {
+		luceneQuery = "_index:" + index + " AND (" + luceneQuery + ")"
+	}
```
sd2k: I wonder if this manipulation of the query will be confusing to the model. Can we pass the index any other way, or is this the way it's supposed to be done?

Note: this query might leak into the error message returned to the LLM if the query fails, which may add to the confusion (or may help the LLM debug and fix the next call 😅 )

sd2k (commented):
Generally LGTM! A few questions around the query we pass, we should clarify a couple of those before merging but it might just come down to better tool descriptions.

tools/es_backend_opensearch.go
LucasHild: I just checked the implementation of the OpenSearch plugin and noticed that overriding the index won't work as I imagined it to work. It will lead to an empty search result.

The OpenSearch plugin’s design assumes one datasource = one index pattern. The plugin’s `_msearch` request always uses the datasource’s configured index pattern. If the datasource is configured with `logs-*` but the user requests `metrics-*`, the `_msearch` header restricts the search to `logs-*` indices, meaning the `_index:metrics-*` filter will match nothing.

Therefore, I just changed the implementation and added validation to the tool call backend. If the requested index does not match the datasource's index, the tool call will return an error.

Let me know if you would prefer a different implementation.

tools/es_backend_opensearch.go
LucasHild: The OpenSearch plugin only supports the Lucene syntax and PPL ([see plugin constants](https://github.com/grafana/opensearch-datasource/blob/main/pkg/opensearch/models.go#L134)).

The only way to use Query DSL would be to use the proxy endpoint. However, this would drop support for AWS SigV4 which is a core feature of OpenSearch.

I adjusted the tool description to make it clear that OpenSearch only supports Lucene syntax.

tools/es_backend_opensearch.go
sd2k: Sounds good, thanks!

tools/es_backend_opensearch.go
sd2k: Sounds good, thank you.

tools/es_backend_opensearch.go
```diff
@@ -0,0 +1,258 @@
...
+			}
+		}
+		return nil, fmt.Errorf("opensearch query failed with status %d: %s", resp.StatusCode, string(bodyBytes))
+	}
+
+	// Parse the /api/ds/query response
+	var result dsQueryResponse
+	if err := json.Unmarshal(bodyBytes, &result); err != nil {
+		return nil, fmt.Errorf("unmarshalling response: %w", err)
+	}
```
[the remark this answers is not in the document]
sd2k: I will refactor this in a later PR.

tools/es_backend_opensearch.go
```diff
@@ -0,0 +1,258 @@
...
+}
+
+// indexMatchesPattern reports whether the given index is compatible with the
+// datasource's configured index pattern. It handles simple wildcard patterns
+// (e.g., "logs-*") using filepath.Match semantics.
+func indexMatchesPattern(pattern, index string) bool {
+	if pattern == "" || pattern == index {
+		return true
+	}
+	matched, err := filepath.Match(pattern, index)
```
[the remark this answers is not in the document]
sd2k: @LucasHild this seems valid, could you switch to `path.Match` instead?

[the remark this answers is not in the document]
LucasHild: @sd2k Good point. I just adjusted it.

sd2k (approved):
Nice, thank you!

PR #672, opened 2026-03-22
feat(pyroscope): add series query and unified query tool

MikeHsu0618:
## Summary
- Add `fetch_pyroscope_series` tool to query time-series metrics from Pyroscope — the "when" dimension of profiling (aggregated values over time intervals).
- Add `fetch_pyroscope` unified tool combining profile (flamegraph) and metrics (time-series) in a single call via `query_type` parameter (`profile`, `metrics`, or `both`).
- `formatSeriesResponse` computes per-series min/max/avg/peak statistics for concise LLM-friendly output.

## Test plan
- [x] `TestFormatSeriesResponse_Empty` — nil series returns message string
- [x] `TestFormatSeriesResponse_SingleSeries` — min/max/avg/peak/points calculation
- [x] `TestFormatSeriesResponse_MultipleSeries` — multi-series grouping
- [x] `TestFormatSeriesResponse_ZeroPointsSkipped` — empty-point series filtered
- [x] `TestFormatSeriesResponse_MinValResetWhenAllZero` — all-zero edge case
- [x] `TestFetchPyroscopeUnified_QueryTypeValidation` — invalid query_type rejected
- [x] `TestAutoStepCalculation` — auto-step: 1h→72s, 5m→15s floor, 24h→1728s
- [ ] Integration tests for `fetch_pyroscope_series` and `fetch_pyroscope` (in `pyroscope_test.go`, requires docker-compose services)

---

> [!NOTE]
> **Medium Risk**
> Adds a new Pyroscope querying surface that fans out to two backend calls (render + series) and changes the MCP tool contract/output to JSON, which may affect downstream consumers and error handling.
> 
> **Overview**
> Introduces a new `query_pyroscope` MCP tool that can return a profile (`/render` DOT call graph), metrics (Pyroscope `SelectSeries` time-series), or **both** in one call via `query_type`, including auto step sizing and optional `group_by`.
> 
> Removes the old `fetch_pyroscope_profile` tool registration/implementation and updates integration tests to validate the new JSON response shape, plus adds unit tests for `buildSeriesResponse` and `query_type` validation.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 78b83b549f7c0986c12c4ad3d8e7e22f48f9411c. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

### DIFF
--- a/tools/pyroscope.go
+++ b/tools/pyroscope.go
@@ -2,8 +2,10 @@ package tools
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"io"
+	"math"
 	"net/http"
 	"net/url"
 	"regexp"
@@ -23,7 +25,7 @@ func AddPyroscopeTools(mcp *server.MCPServer) {
 	ListPyroscopeLabelNames.Register(mcp)
 	ListPyroscopeLabelValues.Register(mcp)
 	ListPyroscopeProfileTypes.Register(mcp)
-	FetchPyroscopeProfile.Register(mcp)
+	QueryPyroscope.Register(mcp)
 }
 
 const listPyroscopeLabelNamesToolPrompt = `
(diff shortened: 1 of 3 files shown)


MikeHsu0618:
> I like the idea of being able to query Pyroscope for series, but think we should avoid adding any more tools for it. Instead we can just have the unified query tool (I think `query_pyroscope` is more aligned to others than `fetch_pyroscope`) with a `query_type` argument (`series`, `profiles`, `both`). WDYT?

Hi @sd2k 
Hello, I'm glad we are on the same page.

I didn't remove the original method and its implementation; I kept them solely for backward compatibility. My ultimate goal is to use a more generic function that includes a "queryType" parameter.

If you agree, I can remove the existing fetch function and unify everything under this "fetchPyroscope" method, which takes three parameters.

Thank you.

sd2k:
Yep no need to worry too much about back-compat since agents automatically get the current version anyway. Thanks!

MikeHsu0618:
> Yep no need to worry too much about back-compat since agents automatically get the current version anyway. Thanks!

Hi @sd2k 
I I kept only QueryPyroscope and supported three parameters.

I also changed the naming from `FetchPyroscope` to `QueryPyroscope` to align with the existing naming style, similar to QueryLokiLog and QueryPrometheus.

tools/pyroscope.go
```diff
@@ -458,3 +387,178 @@ func cleanupDotProfile(profile string) string {
...
+	if queryType == "" {
+		queryType = "both"
+	}
+	if queryType != "profile" && queryType != "metrics" && queryType != "both" {
+		return "", fmt.Errorf("invalid query_type %q: must be \"profile\", \"metrics\", or \"both\"", args.QueryType)
+	}
+
+	// Common setup
+	matchers := stringOrDefault(args.Matchers, "{}")
+	matchersRegex := regexp.MustCompile(`^\{.*\}$`)
```
sd2k: no need to recompile this on every call, this can be factored out into a global

tools/pyroscope_unit_test.go
```diff
@@ -0,0 +1,178 @@
...
+			duration: 24 * time.Hour,
+			wantStep: 1728,
+		},
+	}
+
+	for _, tc := range tests {
+		t.Run(tc.name, func(t *testing.T) {
+			durationSec := tc.duration.Seconds()
+			step := math.Max(durationSec/50.0, 15.0)
+			assert.InDelta(t, tc.wantStep, step, 0.01)
```
sd2k: we're duplicating the logic here so not actually exercising the code, we should either factor it out and test the factored out function or just delete this test.

tools/pyroscope.go
```diff
@@ -458,3 +387,178 @@ func cleanupDotProfile(profile string) string {
...
+// ---------------------------------------------------------------------------
+// query_pyroscope — unified tool: profile + metrics + both
+// ---------------------------------------------------------------------------
+
+const queryPyroscopeToolPrompt = `
+Unified Pyroscope query tool that mirrors Grafana's query_type parameter. Combines profile (flamegraph) and
+metrics (time-series) in a single call. Profile data shows WHICH functions consume resources; metrics data
+shows WHEN consumption spiked. Use query_type="both" for complete analysis in one call.
+
+query_type options (matches Grafana's PyroscopeQueryType):
```
sd2k:
```suggestion
query_type options (extends Grafana's PyroscopeQueryType):
```

since Grafana's type doesn't actually have "both".

tools/pyroscope.go
```diff
@@ -458,3 +387,188 @@ func cleanupDotProfile(profile string) string {
...
+}
+
+// ---------------------------------------------------------------------------
+// query_pyroscope — unified tool: profile + metrics + both
+// ---------------------------------------------------------------------------
+
+const queryPyroscopeToolPrompt = `
+Unified Pyroscope query tool that mirrors Grafana's query_type parameter. Combines profile (flamegraph) and
+metrics (time-series) in a single call. Profile data shows WHICH functions consume resources; metrics data
+shows WHEN consumption spiked. Use query_type="both" for complete analysis in one call.
```
sd2k:
```suggestion
Unified Pyroscope query tool for fetching profiles or metrics from Pyroscope. Profile data shows WHICH functions consume resources; metrics data
shows WHEN consumption spiked. Use query_type="both" for complete analysis in one call.
```

sd2k (approved):
LGTM, just one prompt suggestion! Thanks!

PR #746, opened 2026-04-14
fix: fix get_panel_image error behind api proxy, with subpath grafana support, close #745

### LINKED ISSUE #745: get_panel_image fails silently behind API proxies, on auth errors, and on subpath/legacy Grafana deployments
### Problem

`get_panel_image` in `tools/rendering.go` has three robustness issues that cause it to
return corrupt results or fail in common non-default Grafana deployments.

#### 1. Corrupt image when behind API proxies

The response body is unconditionally base64-encoded as raw image bytes, with MIME type
hardcoded to `image/png`. Some API gateways/reverse proxies wrap the rendered image in
a JSON envelope like `{"base64": "<png-data>"}` — the tool encodes the JSON text itself,
producing a silently corrupted image with no error.

**How to reproduce**: Deploy Grafana behind an API proxy that wraps render responses in
JSON → call `get_panel_image` → returned "image" is base64-encoded JSON, not a valid PNG.

#### 2. No error on authentication/redirect failures

When auth credentials are invalid, Grafana may return an HTML login page with HTTP 200.
The tool treats this HTML as image bytes and returns it as a "successful" PNG — the caller
gets no error signal but the result is unusable.

**How to reproduce**: Set an incorrect `GRAFANA_API_KEY` → call `get_panel_image` →
returns successfully, but the data is base64-encoded HTML markup.

#### 3. Incompatible with subpath deployments and older Grafana versions

`buildRenderURL` only generates the modern `/render/d/{uid}?viewPanel={id}` path.
Grafana instances served from a subpath (e.g. `https://host/grafana/`) or older Grafana
versions that only support `/render/d-solo/{uid}?panelId={id}` receive 404 or routing
errors. Related upstream issue: [grafana/grafana#115529](https://github.com/grafana/grafana/issues/115529).

**How to reproduce**: Configure Grafana with `serve_from_sub_path = true` under a
reverse proxy (e.g. Nginx proxying `/grafana/` → Grafana) → call `get_panel_image` →
render request routes incorrectly, returns 404.

### Expected Behavior

1. Inspect the response `Content-Type` header and handle accordingly:
   - `image/*` → base64-encode, use actual MIME type from header
   - `application/json` → parse `{"base64": "..."}` envelope
   - `text/html` → return clear error: likely an auth or redirect issue
   - Other → return explicit "unexpected Content-Type" error
2. Support a `GRAFANA_RENDER_MODE=legacy` environment variable to switch to
   `/render/d-solo/{uid}` + `panelId` for broader compatibility.

### Environment

- **mcp-grafana**: current main (post PR #448)
- **Grafana**: any version; most impactful on subpath deployments and pre-v11 instances
- **Triggers**: Grafana behind API proxy / invalid credentials / subpath configuration

### Rendering results

- mcp-grafana rendering image tool:

Image

- fix with the legacy mode and suitable response `Content-Type` header setting:

Image

leoswing:
## Summary

Fixes #745

The `get_panel_image` tool fails in two real-world deployment scenarios:

1. **API proxy environments** — Some reverse proxies / API gateways wrap the
   Grafana render response in a JSON envelope (`{"base64": "..."}`) instead of
   passing the raw PNG through. The tool previously assumed every 200 OK
   response was a raw image, producing corrupt output silently.

2. **Subpath / older Grafana deployments** — Grafana instances served under a
   subpath (e.g. `https://example.com/grafana/`) or running older versions may
   not support the `/render/d/?viewPanel=` endpoint, causing render failures
   with no workaround.

### Content-Type aware response handling (`getPanelImage`)

- Inspect the `Content-Type` header and dispatch accordingly:
  | Content-Type | Behavior |
  |---|---|
  | `image/*` | Base64-encode raw bytes, preserve actual MIME type |
  | `application/json` | Unwrap `{"base64": "..."}` JSON envelope from API proxies |
  | `text/html` | Return actionable error (likely auth redirect) |
  | other | Return error with the unexpected Content-Type |
- Previously: blindly base64-encoded the body and hardcoded `image/png`.

### Legacy render mode (`buildRenderURL`)

- Add `GRAFANA_RENDER_MODE=legacy` environment variable opt-in.
- When enabled, panel renders use `/render/d-solo/{uid}?panelId=N` instead of
  `/render/d/{uid}?viewPanel=N`, which has broader compatibility with subpath
  deployments and older Grafana versions.
- Full dashboard renders (no panelId) are unaffected — always use `/render/d/`.

## Test plan

- [x] `TestBuildRenderURL_LegacyMode`: 3 cases covering d-solo path, full
  dashboard fallback, and subpath base URL
- [x] `TestGetPanelImage/JSON_envelope_response_from_API_proxy`: valid JSON
  envelope is correctly unwrapped
- [x] `TestGetPanelImage/JSON_response_without_base64_field`: missing field
  returns descriptive error
- [x] `TestGetPanelImage/HTML_response_returns_auth_error`: HTML detected as
  auth/redirect issue
- [x] `TestGetPanelImage/Unexpected_content_type_returns_error`: unknown
  Content-Type is rejected
- [x] All existing tests pass (modern mode behavior unchanged)

---

> [!NOTE]
> **Medium Risk**
> Moderate risk: changes how `get_panel_image` interprets successful render responses and how panel render URLs are constructed, which could affect compatibility across Grafana deployments. Behavior is gated for legacy mode via env var and covered by added unit tests.
> 
> **Overview**
> Fixes `get_panel_image` to be **Content-Type aware**: it now accepts `image/*` responses (preserving the MIME type), unwraps `application/json` responses that contain a `{ "base64": ... }` envelope, and returns clearer errors for `text/html` (auth/redirect) or unexpected content types.
> 
> Adds an opt-in **legacy render mode** via `GRAFANA_RENDER_MODE=legacy` that switches panel renders to `/render/d-solo/{uid}?panelId=...` (while keeping full-dashboard renders on `/render/d/{uid}`), improving compatibility with subpath and older Grafana setups. Unit tests cover the new URL behavior and the new response-handling branches.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 144247051540178f93b10adf6f6718c4d62d512c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tools/rendering.go
+++ b/tools/rendering.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -83,6 +84,7 @@ type RenderTimeRange struct {
 
 func getPanelImage(ctx context.Context, args GetPanelImageParams) (*mcp.CallToolResult, error) {
 	config := mcpgrafana.GrafanaConfigFromContext(ctx)
+
 	baseURL := strings.TrimRight(config.URL, "/")
 
 	if baseURL == "" {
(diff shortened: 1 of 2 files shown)


tools/rendering.go
```diff
@@ -146,32 +148,75 @@ func getPanelImage(ctx context.Context, args GetPanelImageParams) (*mcp.CallTool
...
+			return nil, fmt.Errorf("render response JSON does not contain a 'base64' field")
+		}
+		base64Data = envelope.Base64
+
+	case strings.HasPrefix(contentType, "text/html"):
+		return nil, fmt.Errorf("render endpoint returned HTML instead of an image (likely an authentication or redirect issue). Verify that GRAFANA_URL is reachable with the configured credentials")
+
+	default:
+		return nil, fmt.Errorf("unexpected Content-Type from render endpoint: %s (expected image/png or application/json)", contentType)
+	}
```
sd2k, line 185: I do not think this should be the responsibility of the MCP server to figure out, we're baking in an assumption that the envelope has `base64` as the key which is surely not applicable to all API proxies. I would expect the API proxy to be configurable to _not_ do this for certain endpoints.

leoswing: Thanks for the feedback @sd2k 

The context here is our enterprise Grafana deployments behind centrally managed API
gateways that enforce unified JSON response wrapping across all upstream services
(e.g. via response transformation middleware or API management layers like
Kong API Gateway). These policies are typically controlled by a separate infra team
and cannot be exempted per-endpoint by the Grafana deployer.

Without this fallback, `get_panel_image` is completely unusable in such
environments with no workaround.

With the JSON response, using the mcp-grafana wouldn't work for this situation, so the envelope codes are provided. Could you please help to review the code changes? Thanks a lot~

leoswing:
Dear @sd2k , I've opened a [new PR](https://github.com/grafana/mcp-grafana/pull/747) to solve this situation, with an extra header accept, could you please help to review the new PR?  That PR is  changed with tiny code.

PR #751, opened 2026-04-15
feat(rendering): support image legacy d-solo render mode, close #750

### LINKED ISSUE #750: Support legacy /d-solo/ render path for single panel image rendering
### Problem

The `get_panel_image` tool currently renders single panels using the `/render/d/{uid}?viewPanel={id}` path. This loads the entire dashboard and opens the panel view overlay, which is heavier and slower than necessary for single panel rendering.

Some Grafana deployments (particularly older versions or those behind custom proxies) rely on the legacy `/render/d-solo/{uid}?panelId={id}` endpoint, which renders only the target panel without loading the full dashboard. The current implementation does not support this render path.

### Proposal

Add a `GRAFANA_RENDER_MODE` environment variable. When set to `legacy`, single panel rendering switches to the `/render/d-solo/` path with the `panelId` query parameter. When unset or set to any other value, the current behavior is preserved.

| Mode | Render path | Panel param |
|------|-------------|-------------|
| default | `/render/d/{uid}` | `viewPanel` |
| `legacy` | `/render/d-solo/{uid}` | `panelId` |

This only affects single panel rendering (when `panelId` is specified). Full dashboard rendering always uses `/render/d/`.

leoswing:
## Summary

- Add `GRAFANA_RENDER_MODE` environment variable to control single panel render path
- When set to `legacy`, use `/render/d-solo/{uid}?panelId={id}` instead of `/render/d/{uid}?viewPanel={id}`
- Default behavior is unchanged; full dashboard rendering is unaffected

Closes #750 

## Test plan

- [x] Unit test: `buildRenderURL` verifies legacy mode produces `/d-solo/` path with `panelId` param
- [x] Unit test: default mode still produces `/d/` path with `viewPanel` param (with `notContains` assertions)
- [x] Integration test: end-to-end `getPanelImage` verifies HTTP request path and query params in legacy mode
- [x] All existing tests pass with no regressions

---

> [!NOTE]
> **Medium Risk**
> Changes the URL/parameter used for single-panel Grafana renders, which could break image rendering against Grafana instances expecting the previous `viewPanel` behavior. Covered by updated unit and integration tests, but still impacts a key external integration point.
> 
> **Overview**
> Single-panel image rendering now uses Grafana’s dedicated ` /render/d-solo/{uid}` endpoint with a `panelId` query parameter instead of rendering the full dashboard via `/render/d/{uid}` with `viewPanel`.
> 
> Tests were updated to assert the new path/param and to explicitly ensure the old `/render/d` + `viewPanel` form is not used for panel renders.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f574e9a9793e0474720d2408876a7dd99b17af50. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tools/rendering.go
+++ b/tools/rendering.go
@@ -8,7 +8,6 @@ import (
 	"io"
 	"net/http"
 	"net/url"
-	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -18,11 +17,6 @@ import (
 	"github.com/mark3labs/mcp-go/server"
 
 	mcpgrafana "github.com/grafana/mcp-grafana"
-)
-
-const (
-	grafanaRenderModeEnvVar = "GRAFANA_RENDER_MODE"
-	renderModeLegacy        = "legacy"
 )
 
 // StringOrSlice is a type that can be unmarshaled from either a JSON string
@@ -180,18 +174,17 @@ func buildRenderURL(baseURL string, args GetPanelImageParams) (string, error) {
 	// Strip trailing slashes from base URL for consistent URL construction
 	baseURL = strings.TrimRight(baseURL, "/")
 
-	// Build the render path
-	renderPath := fmt.Sprintf("/render/d/%s", args.DashboardUID)
-	panelParamKey := "viewPanel"
-
-	// Use solo path and panel ID parameter for single panel rendering in legacy mode
-	if args.PanelID != nil && strings.EqualFold(os.Getenv(grafanaRenderModeEnvVar), renderModeLegacy) {
-		renderPath = fmt.Sprintf("/render/d-solo/%s", args.DashboardUID)
-		panelParamKey = "panelId"
-	}
-
 	// Build query parameters
 	params := url.Values{}
+
+	// Single-panel renders use the purpose-built /d-solo route (lighter than loading
+	// the full dashboard with viewPanel). Full dashboard renders use /d as default.
+	renderPath := fmt.Sprintf("/render/d/%s", args.DashboardUID)
+
+	if args.PanelID != nil {
+		renderPath = fmt.Sprintf("/render/d-solo/%s", args.DashboardUID)
+		params.Set("panelId", strconv.Itoa(*args.PanelID))
+	}
 
 	// Set dimensions
 	width := 1000
@@ -211,11 +204,6 @@ func buildRenderURL(baseURL string, args GetPanelImageParams) (string, error) {
 		scale = *args.Scale
 	}
 	params.Set("scale", strconv.Itoa(scale))
-
-	// Add panel ID if specified (for single panel rendering)
-	if args.PanelID != nil {
-		params.Set(panelParamKey, strconv.Itoa(*args.PanelID))
-	}
 
 	// Add time range
 	if args.TimeRange != nil {
--- a/tools/rendering.go
+++ b/tools/rendering.go
@@ -8,6 +8,7 @@ import (
 	"io"
 	"net/http"
 	"net/url"
+	"os"
 	"strconv"
 	"strings"
 	"time"
@@ -17,6 +18,11 @@ import (
 	"github.com/mark3labs/mcp-go/server"
 
 	mcpgrafana "github.com/grafana/mcp-grafana"
+)
+
+const (
+	grafanaRenderModeEnvVar = "GRAFANA_RENDER_MODE"
+	renderModeLegacy        = "legacy"
 )
 
 // StringOrSlice is a type that can be unmarshaled from either a JSON string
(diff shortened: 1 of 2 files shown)


sd2k:
@leoswing thanks for the contribution. What version of Grafana are you using that requires `/d-solo`?

sd2k:
Thanks for the contribution! The diagnosis here is spot on — `/render/d-solo/` is the right route for single panel rendering and is lighter/faster than loading the full dashboard with `viewPanel`.

However, I don't think we need the env var or the "legacy" framing. `/d-solo/` is still actively maintained in Grafana — it's registered in both the backend routes (`pkg/api/api.go`) and frontend (`SoloPanelPage`), it's what Grafana's own internal screenshot service uses, and it's documented. It was never deprecated.

Since we require Grafana >= 9.0, both routes work on all supported versions. The simpler fix would be to just **always use `/render/d-solo/{uid}?panelId={id}` when rendering a single panel** (i.e. when `PanelID` is set), and keep `/render/d/{uid}` for full dashboard renders. No env var, no mode switching — just use the purpose-built route for the job.

Would you be open to simplifying the PR to just flip the default? The change to `buildRenderURL` would be something like:

```go
renderPath := fmt.Sprintf("/render/d/%s", args.DashboardUID)
panelParamKey := "viewPanel"

if args.PanelID != nil {
    renderPath = fmt.Sprintf("/render/d-solo/%s", args.DashboardUID)
    panelParamKey = "panelId"
}
```

No env var needed.

leoswing:
> Thanks for the contribution! The diagnosis here is spot on — `/render/d-solo/` is the right route for single panel rendering and is lighter/faster than loading the full dashboard with `viewPanel`.
> 
> However, I don't think we need the env var or the "legacy" framing. `/d-solo/` is still actively maintained in Grafana — it's registered in both the backend routes (`pkg/api/api.go`) and frontend (`SoloPanelPage`), it's what Grafana's own internal screenshot service uses, and it's documented. It was never deprecated.
> 
> Since we require Grafana >= 9.0, both routes work on all supported versions. The simpler fix would be to just **always use `/render/d-solo/{uid}?panelId={id}` when rendering a single panel** (i.e. when `PanelID` is set), and keep `/render/d/{uid}` for full dashboard renders. No env var, no mode switching — just use the purpose-built route for the job.
> 
> Would you be open to simplifying the PR to just flip the default? The change to `buildRenderURL` would be something like:
> 
> ```go
> renderPath := fmt.Sprintf("/render/d/%s", args.DashboardUID)
> panelParamKey := "viewPanel"
> 
> if args.PanelID != nil {
>     renderPath = fmt.Sprintf("/render/d-solo/%s", args.DashboardUID)
>     panelParamKey = "panelId"
> }
> ```
> 
> No env var needed.



> Thanks for the contribution! The diagnosis here is spot on — `/render/d-solo/` is the right route for single panel rendering and is lighter/faster than loading the full dashboard with `viewPanel`.
> 
> However, I don't think we need the env var or the "legacy" framing. `/d-solo/` is still actively maintained in Grafana — it's registered in both the backend routes (`pkg/api/api.go`) and frontend (`SoloPanelPage`), it's what Grafana's own internal screenshot service uses, and it's documented. It was never deprecated.
> 
> Since we require Grafana >= 9.0, both routes work on all supported versions. The simpler fix would be to just **always use `/render/d-solo/{uid}?panelId={id}` when rendering a single panel** (i.e. when `PanelID` is set), and keep `/render/d/{uid}` for full dashboard renders. No env var, no mode switching — just use the purpose-built route for the job.
> 
> Would you be open to simplifying the PR to just flip the default? The change to `buildRenderURL` would be something like:
> 
> ```go
> renderPath := fmt.Sprintf("/render/d/%s", args.DashboardUID)
> panelParamKey := "viewPanel"
> 
> if args.PanelID != nil {
>     renderPath = fmt.Sprintf("/render/d-solo/%s", args.DashboardUID)
>     panelParamKey = "panelId"
> }
> ```
> 
> No env var needed.

Dear @sd2k , thanks for your quick response, I'm glad to do this~  let me simplify the PR and upgrade the code changes

leoswing:
> Thanks for the contribution! The diagnosis here is spot on — `/render/d-solo/` is the right route for single panel rendering and is lighter/faster than loading the full dashboard with `viewPanel`.

Dear @sd2k , thanks for your kindness advises， I've already updated the PR  codes, would you please help to review the PR again? Thanks for your time helping this ~

leoswing:
Hi , dear @sd2k , PR codes has been updated already, could you please help to review this PR change again?

sd2k (approved):
Thanks!

PR #754, opened 2026-04-17
fix: register ephemeral sessions to fix horizontal scaling of proxied tools

### LINKED ISSUE #749: Horizontal scaling: multiple replicas fail silently due to per-pod session state
## Summary

Running `mcp-grafana` in `streamable-http` mode with multiple Kubernetes replicas causes **silent failures for proxied tools** (e.g., Tempo). Base tools (Prometheus, Loki, dashboards, etc.) continue to work, but per-session tools fail to register when requests land on a pod that didn't handle the `initialize` call. The upstream Helm chart ([`grafana-mcp/values.yaml`](https://github.com/grafana-community/helm-charts/blob/main/charts/grafana-mcp/values.yaml)) exposes a `replicas` setting and `sessionAffinity` / `sessionAffinityConfig` fields with no warning about this limitation.

## Environment

- **Transport**: `streamable-http` (`-t streamable-http`)
- **Infrastructure**: Kubernetes (EKS) with AWS ALB Ingress
- **Proxied tools**: Enabled (Tempo datasources discovered via MCP)
- **mcp-grafana version**: v0.11.6
- **SDK**: `github.com/mark3labs/mcp-go` v0.46.0

## Background: SDK-level fix in mcp-go v0.43.1

The mcp-go SDK had a related issue (mark3labs/mcp-go#636) where `InsecureStatefulSessionIdManager` became the default and broke multi-instance deployments with "Invalid session ID" errors. This was fixed in mcp-go v0.43.1 (mark3labs/mcp-go#641) by introducing `StatelessGeneratingSessionIdManager` as the new default — it generates session IDs but only validates format, not local existence.

**That SDK fix solves Layer 1 of the problem** (session ID rejection). However, it does **not** solve the application-layer session state issue described below — proxied tools still silently fail on the wrong pod.

## What we tried

We deployed with `replicas: 2` and configured ALB cookie-based sticky sessions:

```yaml
alb.ingress.kubernetes.io/target-group-attributes: stickiness.enabled=true,stickiness.lb_cookie.duration_seconds=3600
```

This didn't help because MCP clients (including Claude Code / Claude CLI) communicate via `POST` requests with an `Mcp-Session-Id` header — they don't persist or return cookies between requests. ALB cookie-based stickiness has no effect.

## Root cause investigation

We traced the request flow through three layers of in-memory state that are all pod-local:

### Layer 1: SDK session ID validation (FIXED in mcp-go v0.43.1)

With the current default `StatelessGeneratingSessionIdManager`, the SDK validates session ID **format only**, not existence. So requests hitting a different pod won't be rejected at this layer. This was the fix from mark3labs/mcp-go#641.

> Relevant code: `WithStateLess(dt.proxied)` in `cmd/mcp-grafana/main.go:381`. Since `dt.proxied` is the `--disable-proxied` flag (default `false`), this calls `WithStateLess(false)`, which is a no-op — the default `StatelessGeneratingSessionIdManager` remains active.

### Layer 2: SDK session registration (`MCPServer.sessions` sync.Map) — STILL BROKEN

When an `initialize` request hits Pod A, the SDK creates a `streamableHttpSession` and registers it in `MCPServer.sessions` (a `sync.Map`). When a subsequent `tools/list` request hits Pod B:

1. Session ID format validation passes (Layer 1 — fixed)
2. `MCPServer.sessions.Load(sessionID)` — **not found** (registered on Pod A)
3. `activeSessions.Load(sessionID)` — **not found**
4. SDK creates an **ephemeral session** that is never registered with `MCPServer`

> Relevant code: `streamable_http.go:396-417` in `mcp-go`

### Layer 3: Application `SessionManager` + `AddSessionTools` failure — STILL BROKEN

The `OnBeforeListTools` hook calls `InitializeAndRegisterProxiedTools`, which:

1. Finds the session missing from `SessionManager` — creates a new one (fallback at `proxied_tools.go:335-342`)
2. Runs datasource discovery successfully
3. Calls `server.AddSessionTools(sessionID, serverTools...)` — this checks `MCPServer.sessions.Load(sessionID)` and returns `ErrSessionNotFound` because the ephemeral session was never registered
4. Logs a warning `"failed to add session tools"` and continues silently

> Relevant code: `proxied_tools.go:423`, `server/session.go:365-368` in `mcp-go`

### Result

- **Base tools** (registered globally via `server.AddTool()` at startup) — **work on all pods**
- **Proxied tools** (registered per-session via `server.AddSessionTools()`) — **silently fail to register on the wrong pod**, so `tools/list` returns only base tools and `tools/call` for a proxied tool returns "tool not found"

## Suggested fix

The stdio transport already registers proxied tools globally via `InitializeAndRegisterServerTools()` (`proxied_tools.go:260-323`). The HTTP/SSE transports could use the same server-global pattern instead of per-session registration. This would:

1. Run datasource discovery once at startup (or on a periodic refresh) instead of per-session
2. Register proxied tools via `server.AddTool()` (global) instead of `server.AddSessionTools()` (per-session)
3. Eliminate the session affinity requirement entirely — the server becomes stateless and can scale horizontally

Additionally, the upstream Helm chart should document this limitation. The `replicas` field and `sessionAffinity` / `sessionAffinityConfig` fields in `grafana-mcp/values.yaml` currently have no warnings about stateful session behavior when proxied tools are enabled.

## Workaround

For now, run with `replicas: 1`. The MCP server itself is lightweight — the bottleneck is typically the downstream Grafana API, not the MCP server.

If you don't use proxied tools (`--disable-proxied`), multiple replicas should work since all base tools are registered globally and the session ID manager is format-only.

sd2k:
## Summary

Fixes #749. When running mcp-grafana in `streamable-http` mode with multiple Kubernetes replicas, proxied tools (e.g., Tempo) silently fail when a request lands on a pod that didn't handle the `initialize` call.

**Root cause:** The mcp-go SDK creates an ephemeral `streamableHttpSession` for POST requests with unknown session IDs, but never registers it in `MCPServer.sessions`. When `InitializeAndRegisterProxiedTools` calls `AddSessionTools`, it fails with `ErrSessionNotFound` because the session isn't in the map.

**Fix:** Call `MCPServer.RegisterSession` in the `OnBeforeListTools` and `OnBeforeCallTool` hooks before registering proxied tools. This ensures the ephemeral session is in the SDK's session map so `AddSessionTools` succeeds. `RegisterSession` uses `LoadOrStore` internally, so it's a no-op for already-registered sessions.

Also adds `SessionManager.SetMCPServer` so the session reaper can call `UnregisterSession` for reaped sessions, preventing a memory leak in the SDK's session map.

## Changes

- `cmd/mcp-grafana/main.go`: Add `ensureSessionRegistered` helper called in both tool hooks before proxied tool initialization
- `session.go`: Add `mcpServer` field and `SetMCPServer` method to `SessionManager`; reaper now calls `MCPServer.UnregisterSession` for reaped sessions
- `session_horizontal_scaling_test.go`: Tests verifying the fix (RegisterSession enables AddSessionTools, reaper cleans up MCPServer sessions)

## Test plan

- [x] New unit tests verify `AddSessionTools` fails without `RegisterSession` and succeeds with it
- [x] New unit tests verify the reaper calls `UnregisterSession` when `SetMCPServer` is set
- [x] All existing tests pass (`go test ./...`)
- [x] `golangci-lint run` passes with 0 issues
- [x] `go vet ./...` passes
- [x] Race detector clean on new tests
- [ ] Integration testing with multiple replicas in Kubernetes (requires manual verification by reporter)

---

> [!NOTE]
> **Medium Risk**
> Touches MCP session registration/reaping and per-request hooks, which can impact tool availability and session lifecycle across transports. Risk is mitigated by being scoped to proxied-tools paths and covered by new unit tests.
> 
> **Overview**
> Ensures proxied tools work when `streamable-http` requests hit a different replica than the one that handled `initialize`, by registering any ephemeral session in `MCPServer.sessions` during `OnBeforeListTools`/`OnBeforeCallTool` before calling `InitializeAndRegisterProxiedTools`.
> 
> Adds `SessionManager.SetMCPServer` so the idle-session reaper also calls `MCPServer.UnregisterSession` for reaped sessions, preventing leaked entries in the SDK’s internal session map, and introduces focused tests covering both the cross-pod registration behavior and reaper cleanup.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4b926b3d0bcb6877e1af662eadb2835677daca43. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/mcp-grafana/main.go
+++ b/cmd/mcp-grafana/main.go
@@ -136,8 +136,10 @@ func newServer(transport string, dt disabledTools, obs *observability.Observabil
 		mcpgrafana.WithSessionTTL(time.Duration(sessionIdleTimeoutMinutes) * time.Minute),
 	)
 
-	// Declare variable for ToolManager that will be initialized after server creation
+	// Declare variables that will be initialized after server creation.
+	// The hooks below capture these by pointer, so they must be declared first.
 	var stm *mcpgrafana.ToolManager
+	var s *server.MCPServer
 
 	// Create hooks
 	hooks := &server.Hooks{
@@ -149,9 +151,24 @@ func newServer(transport string, dt disabledTools, obs *observability.Observabil
 	// (stdio mode is handled by InitializeAndRegisterServerTools; per-session tools
 	// are not supported).
 	if transport != "stdio" && !dt.proxied {
+		// ensureSessionRegistered registers an ephemeral session in MCPServer.sessions
+		// if it's not already there. This is needed for horizontal scaling: when a
+		// request lands on a pod that didn't handle the initialize call, the SDK
+		// creates an ephemeral session that isn't registered, causing AddSessionTools
+		// to fail with ErrSessionNotFound. RegisterSession uses LoadOrStore
+		// internally, so this is a no-op for already-registered sessions.
+		ensureSessionRegistered := func(ctx context.Context) {
+			if s != nil {
+				if session := server.ClientSessionFromContext(ctx); session != nil {
+					_ = s.RegisterSession(ctx, session)
+				}
+			}
+		}
+
 		// OnBeforeListTools: Discover, connect, and register tools
 		hooks.OnBeforeListTools = []server.OnBeforeListToolsFunc{
 			func(ctx context.Context, id any, request *mcp.ListToolsRequest) {
+				ensureSessionRegistered(ctx)
 				if stm != nil {
 					if session := server.ClientSessionFromContext(ctx); session != nil {
 						stm.InitializeAndRegisterProxiedTools(ctx, session)
@@ -163,6 +180,7 @@ func newServer(transport string, dt disabledTools, obs *observability.Observabil
 		// OnBeforeCallTool: Fallback in case client calls tool without listing first
 		hooks.OnBeforeCallTool = []server.OnBeforeCallToolFunc{
 			func(ctx context.Context, id any, request *mcp.CallToolRequest) {
+				ensureSessionRegistered(ctx)
 				if stm != nil {
 					if session := server.ClientSessionFromContext(ctx); session != nil {
 						stm.InitializeAndRegisterProxiedTools(ctx, session)
@@ -175,7 +193,7 @@ func newServer(transport string, dt disabledTools, obs *observability.Observabil
 	// Merge observability hooks with existing hooks
 	hooks = observability.MergeHooks(hooks, obs.MCPHooks())
 
-	s := server.NewMCPServer("mcp-grafana", mcpgrafana.Version(),
+	s = server.NewMCPServer("mcp-grafana", mcpgrafana.Version(),
 		server.WithInstructions(`
 This server provides access to your Grafana instance and the surrounding ecosystem.
 
@@ -202,6 +220,10 @@ Note that some of these capabilities may be disabled. Do not try to use features
 
 	// Initialize ToolManager now that server is created
 	stm = mcpgrafana.NewToolManager(sm, s, mcpgrafana.WithProxiedTools(!dt.proxied))
+
+	// Give the SessionManager a reference to the MCPServer so the reaper can
+	// unregister sessions from the SDK's internal session map.
+	sm.SetMCPServer(s)
 
 	dt.addTools(s)
 	return s, stm, sm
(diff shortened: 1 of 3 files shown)


sd2k:
cursor review

ioanarm (approved):
🚀

PR #755, opened 2026-04-17
docs: clarify UTC default for naive timestamps via server instructions

sd2k:
## Summary

A customer reported that their AI client searched for logs "today" which it
interpreted as `04/15/2026 00:00:00 - 04/15/2026 23:59:59` and found no
results. Turned out the search was UTC, but they were in CDT -- 9 PM CDT is
2 AM UTC next day, so their logs were outside that range.

Root cause: `tools/time_helpers.go` uses `gtime.TimeRange{Now: time.Now()}`
which interprets naive timestamps (no tz offset) in the server's local
time -- UTC in hosted containers. The AI client had no way to know this.

## Approach (revised)

The original draft of this PR added a UTC-hint clause to the `jsonschema`
tag of every time-range parameter across ~14 tool files (~45 fields). That
made the hint visible where the client sees the schema, but it ballooned
the size of every `list_tools` response -- and that response is emitted
per session on every client reconnect.

This revision takes a cheaper approach: a single sentence in the MCP
server `WithInstructions(...)` block. Server instructions are emitted
once per session and cover every tool, so one line does the job of ~45
tag edits at a fraction of the token cost.

The added line:

> Timestamp parameters without a timezone offset are interpreted as UTC.
> Include an offset like '-05:00' or use relative syntax like 'now-1h' to
> query in a different timezone.

## Also included

An internal `timeParamHint()` helper in `tools/time_helpers.go` now
enriches `parseStartTime` / `parseEndTime` error messages with the same
guidance, so callers that pass a malformed timestamp get the hint inline
in the tool error.

## Intentionally skipped

- Per-tool jsonschema tag edits -- cost outweighs benefit now that the
  instructions block covers the same ground.
- `grafana-assistant-app` hosted MCP has its own instructions block to
  update in that repo; tracked separately.

## Token impact

Regenerated `baseline.json` locally via `make token-baseline`:

| provider / model                      | before (tag-per-tool) | after (instructions only) | delta   |
| ------------------------------------- | --------------------- | ------------------------- | ------- |
| tiktoken / gpt-4o                     | 13,922                | 12,611                    | -1,311  |
| anthropic / claude-sonnet-4-5         | 17,937                | 16,572                    | -1,365  |

(These numbers are list_tools-only; the one-time instructions line adds
~30 tokens at session start, which is independent of the per-call
list_tools cost.)

## Test plan

- [x] `make lint-jsonschema` passes
- [x] `go vet ./...` passes
- [x] `make test-unit` passes
- [ ] Confirm with a real AI client that the new server instruction
      reduces the incidence of silent UTC misinterpretation

Generated with Claude Code

---

> [!NOTE]
> **Low Risk**
> Documentation and error-message-only changes; no query semantics or time parsing behavior is modified.
> 
> **Overview**
> Adds a server-wide instruction noting that timestamp parameters without timezone offsets are treated as **UTC**, and suggests using an explicit offset or relative `now-...` syntax.
> 
> Updates `parseStartTime`/`parseEndTime` to wrap `gtime` parse errors with an additional hint so malformed time inputs return actionable timezone guidance inline.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit fd0b552b8f66790c86736ad21b81e47a93295e66. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tools/time_helpers.go
+++ b/tools/time_helpers.go
@@ -1,10 +1,19 @@
 package tools
 
 import (
+	"fmt"
 	"time"
 
 	"github.com/grafana/grafana-plugin-sdk-go/backend/gtime"
 )
+
+// timeParamHint returns the canonical phrase appended to time-parameter parse
+// errors. It warns callers that naive timestamps (no timezone offset) are
+// interpreted as UTC by the server, and suggests including an explicit offset
+// like '-05:00' or using the relative 'now-Xh' syntax accepted by gtime.
+func timeParamHint() string {
+	return "Timestamps without a timezone offset are interpreted as UTC; include an offset like '-05:00'"
+}
 
 // parseStartTime parses start time strings in various formats.
 // Supports: "now", "now-Xs/m/h/d/w", RFC3339, ISO dates, and Unix timestamps.
@@ -17,7 +26,11 @@ func parseStartTime(timeStr string) (time.Time, error) {
 		From: timeStr,
 		Now:  time.Now(),
 	}
-	return tr.ParseFrom()
+	t, err := tr.ParseFrom()
+	if err != nil {
+		return time.Time{}, fmt.Errorf("%w. %s", err, timeParamHint())
+	}
+	return t, nil
 }
 
 // parseEndTime parses end time strings in various formats.
@@ -31,5 +44,9 @@ func parseEndTime(timeStr string) (time.Time, error) {
 		To:  timeStr,
 		Now: time.Now(),
 	}
-	return tr.ParseTo()
+	t, err := tr.ParseTo()
+	if err != nil {
+		return time.Time{}, fmt.Errorf("%w. %s", err, timeParamHint())
+	}
+	return t, nil
 }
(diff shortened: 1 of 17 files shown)


tools/time_helpers.go
```diff
@@ -1,11 +1,20 @@
...
 
 	"github.com/grafana/grafana-plugin-sdk-go/backend/gtime"
 )
 
+// timeParamHint returns the canonical phrase appended to time-parameter parse
+// errors. It warns callers that naive timestamps (no timezone offset) are
+// interpreted as UTC by the server, and suggests including an explicit offset
+// like '-05:00' or using the relative 'now-Xh' syntax accepted by gtime.
+func timeParamHint() string {
+	return "Timestamps without a timezone offset are interpreted as UTC; include an offset like '-05:00'"
```
ioanarm, line 15: Should we mention the relative syntax here too?

sd2k: Good call — done in fd0b552. The hint now reads: `include an offset like '-05:00' or use relative syntax like 'now-1h'`, matching the server instructions text.

sd2k: Good call, done!

ioanarm (approved):
🚀

PR #758, opened 2026-04-20
fix: encode Basic Auth per RFC 7617 in proxied_client.go

### LINKED ISSUE #757: Basic Auth in proxied_client.go violates RFC 7617 for special chars
## TL;DR

`NewProxiedClient` constructs `Authorization: Basic ...` by calling `(*url.Userinfo).String()` and base64-encoding the result. That percent-escapes reserved characters per RFC 3986. RFC 7617 Basic Auth wants the raw `username ":" password` bytes instead. Every other Basic-Auth site in this repo already does it the RFC 7617 way via `SetBasicAuth` or an equivalent raw concat. `proxied_client.go` is the one outlier.

## The spec

[RFC 7617 §2](https://www.rfc-editor.org/rfc/rfc7617#section-2) defines the Basic scheme's credential form as:

> `user-pass = userid ":" password`

The bytes above are then base64-encoded. No percent-escaping, no URL encoding. Just the literal user, colon, literal password. The Go standard library codifies this in [`(*http.Request).SetBasicAuth`](https://pkg.go.dev/net/http#Request.SetBasicAuth):

```go
// From src/net/http/request.go:
func basicAuth(username, password string) string {
    auth := username + ":" + password
    return base64.StdEncoding.EncodeToString([]byte(auth))
}
```

Raw bytes. No escaping.

## What this repo does

[`proxied_client.go:34-37`](https://github.com/grafana/mcp-grafana/blob/main/proxied_client.go#L34-L37):

```go
} else if config.BasicAuth != nil {
    auth := config.BasicAuth.String()
    headers["Authorization"] = "Basic " + base64.StdEncoding.EncodeToString([]byte(auth))
}
```

`(*url.Userinfo).String()` is designed for URL serialization, not Basic Auth. From the Go stdlib (slightly paraphrased):

```go
func (u *Userinfo) String() string {
    if u == nil {
        return ""
    }
    s := escape(u.username, encodeUserPassword)
    if u.passwordSet {
        s += ":" + escape(u.password, encodeUserPassword)
    }
    return s
}
```

The `escape(..., encodeUserPassword)` step percent-escapes `:`, `@`, `/`, `%`, space, and other characters that would be ambiguous in a URL's userinfo segment per RFC 3986. That behavior is correct for constructing URLs. It is the wrong primitive for constructing a Basic Auth header.

## Why the difference matters

For a plain ASCII password containing none of the escaped characters, the two paths produce identical bytes. For any password containing `:`, `@`, `/`, `%`, or space, the current code sends percent-encoded bytes on the wire where RFC 7617 expects raw bytes.

| password | raw (RFC 7617) | current code (via `url.Userinfo.String`) |
|----------|----------------|------------------------------------------|
| `password` | `admin:password` | `admin:password` ✅ |
| `p:w0rd` | `admin:p:w0rd` | `admin:p%3Aw0rd` ❌ |
| `p@ssw0rd` | `admin:p@ssw0rd` | `admin:p%40ssw0rd` ❌ |
| `p/w0rd` | `admin:p/w0rd` | `admin:p%2Fw0rd` ❌ |
| `p%w0rd` | `admin:p%w0rd` | `admin:p%25w0rd` ❌ |
| `pw 0rd` | `admin:pw 0rd` | `admin:pw%200rd` ❌ |

Whether the Grafana instance behind the proxied MCP server happens to be lenient and decode percent-escaped userinfo is not verified in this report. Basic Auth implementations vary, and this particular path runs only when mcp-grafana is proxying to another MCP server (e.g. Tempo) via `NewProxiedClient` using the forwarded user credentials. Either way, the bytes sent deviate from RFC 7617, and a strict parser (or a future hardened implementation) will reject them.

Scope of the bug: it only affects the **proxied MCP transport path**, not direct Grafana API calls. Users who never enable proxied tools never hit it. Users who do enable proxied tools AND use Basic Auth (not a service account token) AND have any of `:` `@` `/` `%` space in their password hit it silently.

## Precedent in this codebase

Every other Basic-Auth construction site in mcp-grafana already uses the raw-bytes form:

| File | Line | Pattern |
|------|------|---------|
| `mcpgrafana.go` | 669 | `req.SetBasicAuth(auth.Username(), password)` |
| `k8s_client.go` | 237-238 | `req.SetBasicAuth(cfg.BasicAuth.Username(), password)` |
| `tools/oncall.go` | 34-35 | `req.SetBasicAuth(cfg.BasicAuth.Username(), password)` |
| `tools/rendering.go` | 121-122 | `req.SetBasicAuth(config.BasicAuth.Username(), password)` |
| `proxied_tools.go` | 146-147 | `req.SetBasicAuth(config.BasicAuth.Username(), password)` |
| `tools/alerting_client.go` | 92-93 | `req.SetBasicAuth(c.basicAuth.Username(), password)` |

Only `proxied_client.go:35-36` uses `(*url.Userinfo).String()`. This looks like an accidental outlier rather than an intentional choice. The difference is easy to miss when you're focused on building a `map[string]string` of headers rather than calling into `http.Request`'s helper.

## Reproduction (standalone Go test snippet)

Save as any `_test.go` file and run `go test -v`:

```go
package imp1_repro_test

import (
	"encoding/base64"
	"net/url"
	"testing"
)

// buggyBasicAuth replicates proxied_client.go:35-36:
//     auth := config.BasicAuth.String()
//     headers["Authorization"] = "Basic " + base64.StdEncoding.EncodeToString([]byte(auth))
func buggyBasicAuth(username, password string) string {
	auth := url.UserPassword(username, password).String()
	return "Basic " + base64.StdEncoding.EncodeToString([]byte(auth))
}

// correctBasicAuth encodes per RFC 7617, identical to
// (*http.Request).SetBasicAuth() in the standard library.
func correctBasicAuth(username, password string) string {
	return "Basic " + base64.StdEncoding.EncodeToString([]byte(username+":"+password))
}

func TestBasicAuthByteLevelSpecDeviation(t *testing.T) {
	cases := []struct {
		name, username, password string
	}{
		{"plain ASCII", "admin", "password"},
		{"colon in password", "admin", "p:w0rd"},
		{"at sign in password", "admin", "p@ssw0rd"},
		{"percent in password", "admin", "p%w0rd"},
		{"space in password", "admin", "pw 0rd"},
		{"slash in password", "admin", "p/w0rd"},
	}
	for _, tc := range cases {
		t.Run(tc.name, func(t *testing.T) {
			got := buggyBasicAuth(tc.username, tc.password)
			want := correctBasicAuth(tc.username, tc.password)
			if got != want {
				t.Errorf("byte-level deviation from RFC 7617 Basic Auth:\n  current : %s\n  expected: %s", got, want)
			}
		})
	}
}
```

Expected result: plain-ASCII passes, all five special-character cases fail with a diagnostic showing the byte-level deviation.

## Suggested fix shape

A natural fit is to extract a small helper that mirrors `(*http.Request).SetBasicAuth`'s encoding, something like `basicAuthHeader(u *url.Userinfo) string`, and have `NewProxiedClient` call it. That keeps the call site small and makes the encoding behavior directly testable with the same table-driven test as the repro above. `proxied_client_test.go` doesn't exist yet in the repo, so the test lands cleanly as a new file.

## Question

Is this framing (byte-level RFC 7617 deviation, observable for passwords with the listed characters, fix via a helper that mirrors `SetBasicAuth`) one you'd accept as a PR? Happy to open it against your preferred shape. I have a working branch (`fix/proxied-client-basic-auth-rfc7617`) with the helper extracted, the test written, lint clean, and `make test-unit` passing.

boinger:
## Summary

Implements the fix discussed in #757: `proxied_client.go` was building the Basic Auth header via `(*url.Userinfo).String()`, which percent-escapes reserved characters per RFC 3986. RFC 7617 wants raw `username ":" password` bytes. See #757 for the full walkthrough of the spec deviation, observable impact, and precedent elsewhere in the codebase.

## What changed

- Extracted `basicAuthHeader(u *url.Userinfo) string` helper that mirrors `(*http.Request).SetBasicAuth`'s encoding.
- `NewProxiedClient` now calls the helper instead of the inline `config.BasicAuth.String()` path.
- New `proxied_client_test.go` with `TestBasicAuthHeaderEncoding`: 7 subtests covering plain ASCII, colon/at/percent/space/slash in password, empty password. Each asserts an exact expected base64 byte string so regressions fail loudly in either direction.

## Test plan

- [x] `make test-unit` passes (including the 7 new subtests)
- [x] `make lint` shows 0 issues (golangci-lint + jsonschema linter)
- [ ] Manual verification of the byte-level change against a Tempo proxied datasource with a password containing `:` (out of scope for a stdlib-pure unit test, but welcome suggestions if maintainers want integration coverage)

Closes #757

---

> [!NOTE]
> **Medium Risk**
> Changes how Basic Auth headers are generated for proxied MCP connections, which could affect connectivity to servers that previously (incorrectly) tolerated percent-encoded credentials. Scope is small and covered by new unit tests for special-character passwords.
> 
> **Overview**
> Fixes Basic Auth header construction in `NewProxiedClient` by replacing `url.Userinfo.String()`-based encoding with a dedicated `basicAuthHeader` helper that base64-encodes raw `username:password` bytes per RFC 7617.
> 
> Adds `proxied_client_test.go` with regression cases for reserved characters (e.g., `:`, `@`, `%`, space, `/`) and empty passwords to ensure the Authorization header bytes match the standard library behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5af28e4d6e5984f162525a9401953f983c1da60c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/proxied_client.go
+++ b/proxied_client.go
@@ -5,6 +5,7 @@ import (
 	"encoding/base64"
 	"fmt"
 	"log/slog"
+	"net/url"
 	"sync"
 
 	mcp_client "github.com/mark3labs/mcp-go/client"
@@ -22,6 +23,17 @@ type ProxiedClient struct {
 	mutex          sync.RWMutex
 }
 
+// basicAuthHeader encodes credentials per RFC 7617:
+// base64(username ":" password) with raw bytes, matching
+// (*http.Request).SetBasicAuth in the standard library. url.Userinfo.String()
+// percent-escapes reserved characters per RFC 3986, which is wrong here —
+// passwords containing ':', '@', '/', '%', or space would be double-encoded
+// on the wire and rejected by spec-compliant Basic Auth parsers.
+func basicAuthHeader(u *url.Userinfo) string {
+	password, _ := u.Password()
+	return "Basic " + base64.StdEncoding.EncodeToString([]byte(u.Username()+":"+password))
+}
+
 // NewProxiedClient creates a new connection to a remote MCP server
 func NewProxiedClient(ctx context.Context, datasourceUID, datasourceName, datasourceType, mcpEndpoint string) (*ProxiedClient, error) {
 	// Get Grafana config for authentication
@@ -32,8 +44,7 @@ func NewProxiedClient(ctx context.Context, datasourceUID, datasourceName, dataso
 	if config.APIKey != "" {
 		headers["Authorization"] = "Bearer " + config.APIKey
 	} else if config.BasicAuth != nil {
-		auth := config.BasicAuth.String()
-		headers["Authorization"] = "Basic " + base64.StdEncoding.EncodeToString([]byte(auth))
+		headers["Authorization"] = basicAuthHeader(config.BasicAuth)
 	}
 
 	// Add org ID header if configured
--- a/proxied_client_test.go
+++ b/proxied_client_test.go
@@ -1,0 +1,71 @@
+package mcpgrafana
+
+import (
+	"net/url"
+	"testing"
+
+	"github.com/stretchr/testify/assert"
+)
+
+// TestBasicAuthHeaderEncoding verifies that basicAuthHeader produces the
+// same bytes as (*http.Request).SetBasicAuth — base64(username ":" password)
+// with raw bytes per RFC 7617. Regression test for the bug where
+// url.Userinfo.String() was used, which percent-escapes reserved characters
+// per RFC 3986 and breaks Basic Auth for passwords containing ':', '@',
+// '/', '%', or space.
+func TestBasicAuthHeaderEncoding(t *testing.T) {
+	cases := []struct {
+		name, username, password, want string
+	}{
+		{
+			name:     "plain ASCII",
+			username: "admin",
+			password: "password",
+			want:     "Basic YWRtaW46cGFzc3dvcmQ=",
+		},
+		{
+			name:     "colon in password",
+			username: "admin",
+			password: "p:w0rd",
+			want:     "Basic YWRtaW46cDp3MHJk",
+		},
+		{
+			name:     "at sign in password",
+			username: "admin",
+			password: "p@ssw0rd",
+			want:     "Basic YWRtaW46cEBzc3cwcmQ=",
+		},
+		{
+			name:     "percent in password",
+			username: "admin",
+			password: "p%w0rd",
+			want:     "Basic YWRtaW46cCV3MHJk",
+		},
+		{
+			name:     "space in password",
+			username: "admin",
+			password: "pw 0rd",
+			want:     "Basic YWRtaW46cHcgMHJk",
+		},
+		{
+			name:     "slash in password",
+			username: "admin",
+			password: "p/w0rd",
+			want:     "Basic YWRtaW46cC93MHJk",
+		},
+		{
+			name:     "empty password",
+			username: "admin",
+			password: "",
+			want:     "Basic YWRtaW46",
+		},
+	}
+	for _, tc := range cases {
+		t.Run(tc.name, func(t *testing.T) {
+			u := url.UserPassword(tc.username, tc.password)
+			got := basicAuthHeader(u)
+			assert.Equal(t, tc.want, got,
+				"basicAuthHeader must produce raw base64(user:password) per RFC 7617, not percent-encoded userinfo")
+		})
+	}
+}


sd2k (approved):
Thanks!

PR #762, opened 2026-04-20
fix: reject malformed X-Grafana-URL header instead of panicking

boinger:
Supersedes the sentinel-client approach. Per @sd2k's review: validates `X-Grafana-URL` at the HTTP middleware layer and returns 400 Bad Request on malformed input.

## Summary

- New exported `mcpgrafana.ValidateGrafanaURLMiddleware(next http.Handler) http.Handler`. Checks `X-Grafana-URL` on every request. Present-but-invalid: 400 Bad Request with error text. Absent or valid: passes through.
- Wired into both HTTP transports in `cmd/mcp-grafana/main.go` (SSE + streamable-http). stdio unaffected.
- Force-pushes the branch, superseding the prior sentinel-client revision. That earlier revision added ~300 LoC of sentinel HTTP-client machinery (`errorClientTransport`, `errorRoundTripper`, `newSentinelGrafanaClient`, `newSentinelIncidentClient`, `validateHeaderURL`, and their tests). None of that is needed with middleware at the transport boundary, and none of it is carried forward in this revision. The diff vs `main` shows the new approach directly, not deletions. The prior revision was only on the PR branch, never merged.
- Introduces `ValidateGrafanaURL` (consolidates trim internally) and `ErrInvalidGrafanaURL`. Both are used by the middleware and the nav guard.
- Adds a guard for the `config.URL` path in `tools/navigation.go:generateDeeplink` to match the middleware's guarantees on the `X-Grafana-URL` path. The guard protects `gc.PublicURL` (populated from Grafana's `/api/frontend/settings` `appUrl` response, not the caller's header), which middleware at the inbound boundary does not cover. Second commit on this branch.

## Behavioral matrix

| X-Grafana-URL | HTTP response |
|---|---|
| Absent | 200 (env fallback) |
| Valid http/https | 200 |
| Present but invalid | 400 Bad Request |
| Present but trims to empty (`/`) | 400 Bad Request (fail closed) |

## Library contract (README.md Programmatic Usage section updated)

- HTTP transports (SSE, streamable-http): middleware handles validation transparently.
- stdio transport: config from env; operators must set a valid `GRAFANA_URL`.
- Library consumers wiring mcp-grafana context-funcs into their own `http.Server`: install `ValidateGrafanaURLMiddleware` to match the binary's behavior.
- Library consumers using `NewGrafanaClient` directly (stdio or programmatic construction): pre-validate with `ValidateGrafanaURL` (the existing `NewGrafanaClient` docstring already notes this).
- Both patterns are shown in README "Programmatic Usage."

## Errors.Is contract change

`errors.Is(err, mcpgrafana.ErrInvalidGrafanaURL)` no longer triggers on HTTP-transport paths. Requests are rejected before extractors run. The sentinel is still returned by `ValidateGrafanaURL` itself (wrapped), so direct validation callers continue to detect it.

## Tests

- `TestValidateGrafanaURLMiddleware`: 12 cases (pass-through, rejection shapes, duplicate-header, CR-LF injection, 20-goroutine concurrency under `-race`).
- `TestValidateGrafanaURL`: 20 cases (scheme allow-list, host requirement, trim behavior).
- `TestExtractGrafanaClientFromHeaders_ValidURL` + `_NoHeader`: smoke coverage for client construction, with an httptest request counter proving the extractor wires a client at the header-supplied URL.
- `TestGenerateDeeplink_RejectsMalformedBaseURL_ClientWithEmptyPublicURL` + `_NoClient`: guard regression tests.

## Test plan

- [x] `make test-unit` passes
- [x] `make lint` reports 0 issues
- [x] Manual: valid/malformed/`/`-collapse/absent header behavior on both transports

## Follow-ups

- #776: `ValidateGrafanaURL` should reject embedded credentials (`user:pass@host`). Separate PR to follow this one.

## Closes

- #763 (extract-boundary, subsumed by middleware)

---

> [!NOTE]
> **Medium Risk**
> Changes request handling for SSE/streamable HTTP by rejecting malformed `X-Grafana-URL` with `400`, which can affect clients that previously relied on permissive parsing; also adds URL validation to deeplink generation to prevent bad URLs from propagating.
> 
> **Overview**
> Prevents malformed `X-Grafana-URL` values from reaching Grafana client construction by introducing `ValidateGrafanaURL` (with `ErrInvalidGrafanaURL`) and an `ValidateGrafanaURLMiddleware` that returns **400 Bad Request** when the header is present but invalid.
> 
> Wires this middleware into the binary’s SSE and streamable-http handlers, and adds a separate base-URL validation guard in `tools/navigation.go` so deeplink generation errors out (instead of emitting malformed links) when `config.URL`/`PublicURL` is invalid. Updates README with guidance for library consumers and adds comprehensive unit tests for the validator, middleware, extractor smoke paths, and deeplink regression cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9e97ee03015171d74e800a9d355bdff243cea400. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/README.md
+++ b/README.md
@@ -854,6 +854,26 @@ grafanaConfig := mcpgrafana.GrafanaConfig{
 contextFunc := mcpgrafana.ComposedStdioContextFunc(grafanaConfig)
 ```
 
+**URL validation when wiring your own HTTP server:**
+
+When library consumers wire mcp-grafana's context functions into their own `http.Server`, install `ValidateGrafanaURLMiddleware` to reject malformed `X-Grafana-URL` headers with 400 Bad Request (matching the binary's behavior):
+
+```go
+mux.Handle(path, mcpgrafana.ValidateGrafanaURLMiddleware(yourMCPHandler))
+```
+
+When calling `NewGrafanaClient` directly (stdio or programmatic construction), pre-validate untrusted URLs to avoid a reachable panic:
+
+```go
+if err := mcpgrafana.ValidateGrafanaURL(urlFromHeader); err != nil {
+    http.Error(w, err.Error(), http.StatusBadRequest)
+    return
+}
+client := mcpgrafana.NewGrafanaClient(ctx, urlFromHeader, apiKey, nil)
+```
+
+Both patterns share `ValidateGrafanaURL` as the single validator.
+
 ### Server TLS Configuration (Streamable HTTP Transport Only)
 
 When using the streamable HTTP transport (`-t streamable-http`), you can configure the MCP server to serve HTTPS instead of HTTP. This is useful when you need to secure the connection between your MCP client and the server itself.
--- a/cmd/mcp-grafana/main.go
+++ b/cmd/mcp-grafana/main.go
@@ -385,7 +385,10 @@ func run(transport, addr, basePath, endpointPath string, logLevel slog.Level, dt
 		if basePath == "" {
 			basePath = "/"
 		}
-		mux.Handle(basePath, observability.WrapHandler(srv, basePath))
+		mux.Handle(basePath, observability.WrapHandler(
+			mcpgrafana.ValidateGrafanaURLMiddleware(srv),
+			basePath,
+		))
 		mux.HandleFunc("/healthz", handleHealthz)
 		if obs.MetricsEnabled {
 			if obs.MetricsAddress == "" {
@@ -411,7 +414,10 @@ func run(transport, addr, basePath, endpointPath string, logLevel slog.Level, dt
 		}
 		srv := server.NewStreamableHTTPServer(s, opts...)
 		mux := http.NewServeMux()
-		mux.Handle(endpointPath, observability.WrapHandler(srv, endpointPath))
+		mux.Handle(endpointPath, observability.WrapHandler(
+			mcpgrafana.ValidateGrafanaURLMiddleware(srv),
+			endpointPath,
+		))
 		mux.HandleFunc("/healthz", handleHealthz)
 		if obs.MetricsEnabled {
 			if obs.MetricsAddress == "" {
--- a/tools/navigation.go
+++ b/tools/navigation.go
@@ -43,6 +43,16 @@ func generateDeeplink(ctx context.Context, args GenerateDeeplinkParams) (string,
 		return "", fmt.Errorf("grafana url not configured. Please set GRAFANA_URL environment variable or X-Grafana-URL header")
 	}
 
+	// Validate baseURL separately from the inbound X-Grafana-URL middleware:
+	// gc.PublicURL is populated by fetchPublicURL from Grafana's
+	// /api/frontend/settings appUrl response, which is not covered by the
+	// middleware at the HTTP transport boundary. A misconfigured Grafana can
+	// therefore return a malformed appUrl that flows into deeplink construction
+	// (e.g. http://%gg/d/<uid>) unless checked here.
+	if err := mcpgrafana.ValidateGrafanaURL(baseURL); err != nil {
+		return "", fmt.Errorf("grafana url is invalid: %w. Please set GRAFANA_URL environment variable or X-Grafana-URL header", err)
+	}
+
 	var deeplink string
 
 	switch strings.ToLower(args.ResourceType) {
--- a/tools/navigation_test.go
+++ b/tools/navigation_test.go
@@ -2,6 +2,7 @@ package tools
 
 import (
 	"context"
+	"errors"
 	"net/url"
 	"testing"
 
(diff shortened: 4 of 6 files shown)


boinger:
The behavioral shift this addresses was called out in the original PR body under "Behavioral shifts to note", with a note that I was happy to bundle the fix here if reviewers preferred. Bugbot's independent finding is strong signal to land it in this PR rather than as a follow-up.

`af9b7e6` adds a `ValidateGrafanaURL` guard in `tools/navigation.go:generateDeeplink` right after the existing empty-baseURL check. When the sentinel client is attached with an empty `PublicURL`, `generateDeeplink` falls through to `config.URL` (which `ExtractGrafanaInfoFromHeaders` sets unconditionally from the raw header). Without this guard, the malformed URL gets embedded in the returned deeplink (e.g. `http://%gg/d/<uid>`) with no error signal back to the caller.

Two new tests in `tools/navigation_test.go` cover the sentinel-client path (primary, reproducing the finding) and the no-client path. Both assert `errors.Is(err, mcpgrafana.ErrInvalidGrafanaURL)`.

Keeping scope narrow to `navigation.go` since it's the one surface with the silent-corruption shape (plausible-looking garbage URL returned to the LLM). Other tools reading `config.URL` either fail observably at HTTP request time or leak the malformed value through log strings and error messages. Filing a follow-up issue for the broader `config.URL` propagation problem so the fix at `ExtractGrafanaInfoFromHeaders` can be reviewed separately.

sd2k:
Hi @boinger, thanks for the issue report. I agree this is an issue but I think the solution using a sentinel client may be slightly overkill. Could you instead experiment with a middleware that validates the URL (if provided) and returns a 400 if not? I think we could do that in ~50LoC and catch every request without needing to wait until the tool call to determine whether the URL is correct from tools (especially manually, like we had to do in the navigation client).

boinger:
Thanks @sd2k, that makes good sense. The sentinel-client approach was absolutely a workaround for "don't touch the transport layer". With that constraint lifted, middleware is the right shape. I am in for the rework, but I want to surface a few structural questions before coding so I don't have to redo it.

**1. Amend this PR, or close and file fresh?**

The sentinel approach in this branch touches `sentinel_client.go`, `ExtractGrafanaClientFromHeaders` / `ExtractIncidentClientFromHeaders`, both `client_cache.go` extractors, and a dedicated test file. Middleware + 400 is a different layer with most of that deleted. Either path works on my end:
- Force-push this branch with the redesign. Preserves this review thread, but the diff is a near-total rewrite.
- Close this PR, open a new one. Cleaner diff history, loses this thread's comments.

I default to force-push unless you prefer fresh.

**2. Middleware as library export, or binary-only?**

`ExtractGrafanaInfoFromHeaders` is an exported package symbol. A library user who wires it into their own `http.Server` without the middleware would lose header validation entirely. Two directions:
- Export something like `ValidateGrafanaURLMiddleware` as a library primitive, document in README as the recommended wiring. Extractor stays a simple data-extraction function.
- Keep middleware binary-only in `cmd/mcp-grafana/main.go`. Library consumers who skip it are on their own (closer to Go stdlib idiom).

I lean toward exporting. That's the one moving piece that ties the library surface to the same defensive model as the binary, for ~5 extra lines.

**3. #763 (the stacked PR)**

#763 validates the header at `ExtractGrafanaInfoFromHeaders`. Once middleware rejects with 400 before the context-func chain runs, that inline validation becomes dead code on the binary's path. Plan is to close #763 when this redesign lands, unless you want its inline validation retained as a library-level safety net for consumers who bypass the middleware. Let me know.

**4. Scope note: `af9b7e6` (the nav.go guard in this PR) stays regardless.**

That guard is not about `X-Grafana-URL`. `tools/navigation.go:generateDeeplink` reads `gc.PublicURL`, which is populated by `fetchPublicURL` from Grafana's `/api/frontend/settings` `appUrl` response. That response is unvalidated, and the guard is the only defender against a misconfigured Grafana returning a garbage `appUrl`. Middleware on inbound `X-Grafana-URL` doesn't touch that data flow. Flagging in case it read as "part of what middleware replaces" in your review. It isn't.

Happy to start coding as soon as you weigh in on (1), (2), (3).

sd2k:
Thanks @boinger. To your points:

1. Happy with a force push, I'm not fussy!
2. I think it's worth exporting, yes.
3. Indeed, I think we can close it.
4. Valid point, thanks for the correction.

sd2k (approved):
Looks great, thanks for addressing feedback and for the contribution!

boinger:
Of course!  Thanks, @sd2k for providing the feedback to steer me toward what the project prefers!

PR #763, opened 2026-04-21
fix(mcpgrafana): validate X-Grafana-URL at extract boundary

boinger:
> **Depends on PR #762.** This PR builds on `ValidateGrafanaURL` from #762 and will not compile/merge cleanly on its own. Please land #762 first, then merge this against the rebased base. If you want to squash-merge both in sequence, this PR is fine to merge second; if you want to rebase-merge, this PR's branch will auto-rebase onto #762's merged state.

## Summary

`ExtractGrafanaInfoFromHeaders` previously wrote `config.URL` unconditionally from the raw `X-Grafana-URL` header, even when the URL was malformed. PR #762 closed the client-side failure path via sentinel clients; this PR closes the operator-visibility gap at the config boundary.

## What changed

In `mcpgrafana.go::ExtractGrafanaInfoFromHeaders`, if the raw `X-Grafana-URL` header is present, the trimmed value is validated inline via `ValidateGrafanaURL` (the helper introduced in #762). On error: `slog.Warn` naming the rejection + `config.URL = ""` (fail-closed). On absent header: existing env-fallback behavior unchanged.

Semantic matrix (new):

| X-Grafana-URL | Behavior |
|---------------|----------|
| Valid http/https | `config.URL` set from header |
| Present but invalid (e.g. `http://%gg`) | `config.URL = ""`, WARN emitted |
| Present but trims to empty (e.g. `/`) | `config.URL = ""`, WARN emitted (fail closed, NOT env fallback) |
| Absent | `config.URL` from `GRAFANA_URL` env, else `defaultGrafanaURL` |

## Why inline, not via `validateHeaderURL`

`validateHeaderURL` from #762 has four existing callers (two grafana client extractors + two incident client extractors, cached and uncached variants). Its current contract is "trimmed-empty implies absent," which the client extractors depend on. This PR wants the opposite semantic ("present-but-trim-empty implies invalid") so a literal `/` header is rejected rather than silently falling back to env. Expressing that inline keeps the two call-site semantics separate and legible; modifying `validateHeaderURL` would silently shift behavior across all four callers with no test coverage for that shift.

## Tests

`TestExtractGrafanaInfoFromHeaders_URLValidation` is table-driven with six cases:

- valid header + env
- invalid header (`http://%gg`) + env produces `""`, WARN
- absent header + env produces env
- absent header + absent env produces `defaultGrafanaURL` (`http://localhost:3000`, via existing fallback)
- `/` header + env produces `""`, WARN (the trim-collapse regression guard)
- valid header + absent env

Each sub-case uses `t.Setenv` for hermetic env state. WARN capture via a test-local `capturingHandler` installed as `slog.Default()` with `t.Cleanup` restore. No `t.Parallel`.

## Test plan

- [x] `make test-unit` passes
- [x] `make lint` reports 0 issues
- [x] Manual verification: valid / malformed / `/` / absent-header cases all produce expected behavior + stderr WARN on the two fail-closed paths

---

> [!NOTE]
> **Medium Risk**
> Changes request-header handling for Grafana/Incident client/config extraction to reject malformed `X-Grafana-URL` values and return sentinel errors instead of panicking or generating bad links, which may alter behavior for misconfigured callers.
> 
> **Overview**
> Adds strict validation for caller-supplied `X-Grafana-URL` and makes header-based flows **fail-closed** when the header is present but invalid (including trim-to-empty cases like `/`), emitting `slog.Warn` and leaving `config.URL` empty rather than silently falling back to env.
> 
> Client extractors (cached and uncached, Grafana and Incident) now attach **non-cached sentinel clients** on invalid headers so downstream tool calls consistently surface `ErrInvalidGrafanaURL` instead of triggering `NewGrafanaClient` panics. Deeplink generation now validates its chosen base URL to avoid returning malformed links.
> 
> Adds comprehensive unit tests for URL validation, sentinel client behavior, cache non-poisoning, concurrent mixed valid/invalid requests, and WARN-log emission on rejected headers.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5814a8745b9b0ac7d299bb9152109c2788950cb0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/client_cache.go
+++ b/client_cache.go
@@ -244,11 +244,21 @@ func hashAPIKey(key string) string {
 }
 
 // extractGrafanaClientCached creates an httpContextFunc that uses the cache.
+// If X-Grafana-URL is present but malformed or uses a non-HTTP(S) scheme, a
+// sentinel client is attached to the context instead of invoking
+// NewGrafanaClient (which would panic on url.Parse failure). The sentinel is
+// not cached: caching it would let a single bad caller poison a cache slot
+// keyed by their malformed URL.
 func extractGrafanaClientCached(cache *ClientCache) httpContextFunc {
 	return func(ctx context.Context, req *http.Request) context.Context {
 		config := GrafanaConfigFromContext(ctx)
 		if config.OrgID == 0 {
 			slog.Warn("No org ID found in request headers or environment variables, using default org. Set GRAFANA_ORG_ID or pass X-Grafana-Org-Id header to target a specific org.")
+		}
+
+		if err := validateHeaderURL(req); err != nil {
+			slog.Warn("rejecting malformed X-Grafana-URL header, attaching sentinel Grafana client", "error", err)
+			return WithGrafanaClient(ctx, newSentinelGrafanaClient(err))
 		}
 
 		u, apiKey, basicAuth, _ := extractKeyGrafanaInfoFromReq(req)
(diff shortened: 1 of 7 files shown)


boinger:
Closing in favor of #762's redesigned middleware approach. The force-push (2026-04-22) replaces the prior sentinel-client revision with an HTTP middleware that returns 400 on malformed X-Grafana-URL at the transport boundary, subsuming this PR's extract-boundary validation. Per @sd2k's review of #762.

PR #767, opened 2026-04-21
feat(tools): support PromQL queries against VictoriaMetrics datasources

### LINKED ISSUE #766: query_prometheus fails against VictoriaMetrics datasource (victoriametrics-metrics-datasource plugin)
## Summary

`query_prometheus` (and `query_prometheus_histogram`) fail when targeting a datasource of type `victoriametrics-metrics-datasource` (the official VictoriaMetrics Grafana plugin). Discovery tools (`list_prometheus_metric_names`, `list_prometheus_label_names`, `list_prometheus_label_values`) work fine against the same datasource.

This mirrors the gap that was fixed for Cloud Monitoring / Stackdriver in #581 / #647 — a non-native datasource that exposes a PromQL-compatible API but is not of `type: prometheus`.

## Reproduction

Datasource configured in Grafana:
- `type: victoriametrics-metrics-datasource`
- `name: VictoriaMetrics`

Call:
```
query_prometheus(
  datasourceUid="<uid>",
  expr="up",
  queryType="instant",
  endTime="now",
)
```

## Actual

Instant query:
```
querying Prometheus instant: Skip: do not know how to skip: 0,
error found in #0 byte of ...|||..., bigger context ...|||...
```

Range query (more revealing):
```
querying Prometheus range: bad_response: readObjectStart: expect { or n,
but found H, error found in #1 byte of ...|Hello from |...,
bigger context ...|Hello from VM data source!|...
```

The `Hello from VM data source!` response is what the VictoriaMetrics plugin returns from its root/health endpoint, suggesting the MCP query path is being routed to the wrong endpoint on the plugin (it's not using the plugin's PromQL-compatible `/api/v1/query` route).

## Expected

`query_prometheus` and `query_prometheus_histogram` should work against `victoriametrics-metrics-datasource` the same way they do against native Prometheus, since VictoriaMetrics exposes a fully PromQL-compatible API.

## What works

- `list_prometheus_metric_names` ✅
- `list_prometheus_label_names` ✅
- `list_prometheus_label_values` ✅

## What doesn't

- `query_prometheus` ❌
- `query_prometheus_histogram` ❌ (presumably, same code path)

## Related

- #581 / #647 — same shape of problem, fixed for Cloud Monitoring / Stackdriver
- #634 — adds VictoriaLogs (logs), not VictoriaMetrics
- #585 — `get_dashboard_panel_queries` also drops non-Prometheus panels

DanielRamosAcosta:
## Summary

Adds support for `query_prometheus` (and `query_prometheus_histogram`) against datasources of type `victoriametrics-metrics-datasource` (the official VictoriaMetrics Grafana plugin).

Closes #766.

## Why

The native Prometheus backend talks to datasources via Grafana's `/api/datasources/uid/{uid}/resources/api/v1/query` proxy. The VictoriaMetrics plugin does **not** expose `/api/v1/query` on that path — it returns its plugin landing response (`Hello from VM data source!`), which trips the Prometheus client's JSON parser:

```
querying Prometheus range: bad_response: readObjectStart: expect { or n,
but found H, error found in #1 byte of ...|Hello from |...,
bigger context ...|Hello from VM data source!|...
```

Discovery endpoints (`list_prometheus_metric_names`, `list_prometheus_label_names`, `list_prometheus_label_values`) already work because the plugin serves them under different paths through the same proxy.

## Approach

This mirrors the fix used for Cloud Monitoring in #647: route queries through Grafana's generic `/api/ds/query` endpoint (the same path the Grafana frontend uses to talk to plugin-backed datasources). The new backend embeds `prometheusBackend` so label/metadata lookups continue to use the existing implementation; only `Query` is overridden.

The query payload matches the [`Query` struct in the VictoriaMetrics datasource plugin](https://github.com/VictoriaMetrics/victoriametrics-datasource/blob/main/pkg/plugin/query.go), with `instant` / `range` booleans rather than separate URL paths. All frame-to-Prometheus-value conversion (`framesToVector`, `framesToMatrix`) is reused unchanged from the Cloud Monitoring backend.

## What changed

- `tools/prom_backend.go` — add `victoriametrics-metrics-datasource` case to the backend selector.
- `tools/prom_backend_victoriametrics.go` — new backend (~150 LOC), embeds `prometheusBackend`, overrides only `Query`.
- `tools/prom_backend_victoriametrics_test.go` — unit tests covering payload shape (instant/range, step → interval/intervalMs), instant response decoding, upstream error propagation, and HTTP error handling.

## Test plan

- [x] `make lint-jsonschema` passes
- [x] `make test-unit` passes
- [x] Manually verified against a real Grafana + VictoriaMetrics datasource (instant, range, and `query_prometheus_histogram` all return correct results)
- [ ] Integration tests against a docker-composed VictoriaMetrics — happy to add if you want; would mirror the existing prometheus integration setup

## Notes

- VictoriaLogs (PR #634) is a different datasource type (`victoriametrics-logs-datasource`) and a different code path (Loki-equivalent tools), so this is independent.
- Could also add a Cursor-style migration test that hits the real plugin if you have a VM container in CI; otherwise the unit tests cover the contract.

---

> [!NOTE]
> **Medium Risk**
> Introduces a new query execution path via `/api/ds/query` and refactors shared query plumbing, which could affect query behavior/error handling for Cloud Monitoring and the new VictoriaMetrics datasource type.
> 
> **Overview**
> Adds first-class support for datasources of type `victoriametrics-metrics-datasource` by introducing a new `victoriaMetricsBackend` that reuses the existing Prometheus backend for discovery but routes `Query` calls through Grafana’s generic `/api/ds/query` (building the plugin-specific payload and normalizing time/step defaults).
> 
> Refactors the Cloud Monitoring backend to reuse a new shared `doDSQuery` helper (moved to `prom_backend.go`) and adds unit tests covering VictoriaMetrics query payload shape, frame decoding to Prometheus `model.Value`, and error handling.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit dc169a352486821d23188138a77d7ab037a4259a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tools/prom_backend_victoriametrics.go
+++ b/tools/prom_backend_victoriametrics.go
@@ -1,0 +1,149 @@
+package tools
+
+import (
+	"bytes"
+	"context"
+	"encoding/json"
+	"fmt"
+	"io"
+	"net/http"
+	"strconv"
+	"time"
+
+	"github.com/grafana/grafana-openapi-client-go/models"
+	mcpgrafana "github.com/grafana/mcp-grafana"
+	"github.com/prometheus/common/model"
+)
+
+const victoriaMetricsDatasourceType = "victoriametrics-metrics-datasource"
+
+// victoriaMetricsBackend implements promBackend for the VictoriaMetrics
+// Grafana plugin (type: victoriametrics-metrics-datasource).
+//
+// Discovery endpoints (label names, label values, metric metadata) are served
+// correctly by the plugin's resource proxy and reuse the native Prometheus
+// backend. Only Query is overridden to go through Grafana's /api/ds/query
+// endpoint, because the plugin's resource proxy does not expose
+// /api/v1/query — hitting it returns the plugin's "Hello from VM data source!"
+// landing response and trips the Prometheus client's JSON parser.
+type victoriaMetricsBackend struct {
+	*prometheusBackend
+	httpClient    *http.Client
+	baseURL       string
+	datasourceUID string
+}
+
+func newVictoriaMetricsBackend(ctx context.Context, uid string, ds *models.DataSource) (*victoriaMetricsBackend, error) {
+	promBackend, err := newPrometheusBackend(ctx, uid, ds)
+	if err != nil {
+		return nil, err
+	}
+
+	cfg := mcpgrafana.GrafanaConfigFromContext(ctx)
+	baseURL := trimTrailingSlash(cfg.URL)
+
+	transport, err := mcpgrafana.BuildTransport(&cfg, nil)
+	if err != nil {
+		return nil, fmt.Errorf("failed to create custom transport: %w", err)
+	}
+	transport = NewAuthRoundTripper(transport, cfg.AccessToken, cfg.IDToken, cfg.APIKey, cfg.BasicAuth)
+	transport = mcpgrafana.NewOrgIDRoundTripper(transport, cfg.OrgID)
+
+	client := &http.Client{
+		Transport: mcpgrafana.NewUserAgentTransport(transport),
+		Timeout:   30 * time.Second,
+	}
+
+	return &victoriaMetricsBackend{
+		prometheusBackend: promBackend,
+		httpClient:        client,
+		baseURL:           baseURL,
+		datasourceUID:     ds.UID,
+	}, nil
+}
+
+// Query executes a PromQL/MetricsQL query via Grafana's /api/ds/query endpoint.
+//
+// The VictoriaMetrics plugin distinguishes instant vs range via two boolean
+// fields on the query payload (instant/range), unlike Prometheus where the
+// distinction is the URL path.
+func (b *victoriaMetricsBackend) Query(ctx context.Context, expr string, queryType string, start, end time.Time, stepSeconds int) (model.Value, error) {
+	if start.IsZero() {
+		start = end
+	}
+	if end.IsZero() || end.Before(start) {
+		end = start
+	}
+
+	step := stepSeconds
+	if step == 0 {
+		step = 60
+	}
+
+	query := map[string]interface{}{
+		"refId": "A",
+		"datasource": map[string]string{
+			"type": victoriaMetricsDatasourceType,
+			"uid":  b.datasourceUID,
+		},
+		"expr":       expr,
+		"instant":    queryType == "instant",
+		"range":      queryType == "range",
+		"interval":   fmt.Sprintf("%ds", step),
+		"intervalMs": int64(step) * 1000,
+	}
+
+	payload := map[string]interface{}{
+		"queries": []interface{}{query},
+		"from":    strconv.FormatInt(start.UnixMilli(), 10),
+		"to":      strconv.FormatInt(end.UnixMilli(), 10),
+	}
+
+	resp, err := b.doDSQuery(ctx, payload)
+	if err != nil {
+		if queryType == "range" {
+			return nil, fmt.Errorf("querying VictoriaMetrics range: %w", err)
+		}
+		return nil, fmt.Errorf("querying VictoriaMetrics instant: %w", err)
+	}
+
+	return framesToPrometheusValue(resp, queryType)
+}
+
+// doDSQuery executes a request against Grafana's /api/ds/query endpoint.
+// Mirrors the Cloud Monitoring backend's helper of the same name; kept local
+// to avoid coupling the two backends.
+func (b *victoriaMetricsBackend) doDSQuery(ctx context.Context, payload map[string]interface{}) (*dsQueryResponse, error) {
+	payloadBytes, err := json.Marshal(payload)
+	if err != nil {
+		return nil, fmt.Errorf("marshaling query payload: %w", err)
+	}
+
+	req, err := http.NewRequestWithContext(ctx, http.MethodPost, b.baseURL+"/api/ds/query", bytes.NewReader(payloadBytes))
+	if err != nil {
+		return nil, fmt.Errorf("creating request: %w", err)
+	}
+	req.Header.Set("Content-Type", "application/json")
+
+	resp, err := b.httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("executing request: %w", err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+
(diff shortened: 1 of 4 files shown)


sd2k:
This LGTM, I think VM is similar enough to Prom that we can skip the integration tests, but I wouldn't say no if you want to add them.

DanielRamosAcosta:
ready @sd2k 😄

tools/prom_backend_victoriametrics.go
```diff
@@ -0,0 +1,149 @@
...
+	if err != nil {
+		return nil, fmt.Errorf("failed to create custom transport: %w", err)
+	}
+	transport = NewAuthRoundTripper(transport, cfg.AccessToken, cfg.IDToken, cfg.APIKey, cfg.BasicAuth)
+	transport = mcpgrafana.NewOrgIDRoundTripper(transport, cfg.OrgID)
+
+	client := &http.Client{
+		Transport: mcpgrafana.NewUserAgentTransport(transport),
+		Timeout:   30 * time.Second,
+	}
```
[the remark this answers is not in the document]
DanielRamosAcosta: This is a false positive — `BuildTransport` does not apply auth/OrgID/UserAgent wrapping.

Looking at `mcpgrafana.go:455-478`, `BuildTransport` only:
1. picks `base` or `http.DefaultTransport`,
2. wraps with TLS if `cfg.TLSConfig != nil`,
3. wraps with `NewExtraHeadersRoundTripper` if `cfg.ExtraHeaders` is non-empty,
4. returns.

It does not touch `NewAuthRoundTripper`, `NewOrgIDRoundTripper`, `NewUserAgentTransport`, or `otelhttp`. The chain at `mcpgrafana.go:832-849` that does apply those is a separate path inside `NewGrafanaClient` (set via reflection on the openapi client's internal Transport field) — not part of `BuildTransport`.

The Cloud Monitoring backend this code mirrors does the same thing: `prom_backend_cloudmonitoring.go:41-50` calls `BuildTransport` and then explicitly wraps with `NewAuthRoundTripper` → `NewOrgIDRoundTripper` → `NewUserAgentTransport`. The Incident client (`mcpgrafana.go:929-934`, `954-959`) also wraps `BuildTransport` with OrgID + UserAgent. If `BuildTransport` already did this, all of those would be duplicated in production today — and they aren't.

Removing the wrappers here would drop auth headers, `X-Grafana-Org-Id`, and the User-Agent from outgoing requests. Leaving as-is.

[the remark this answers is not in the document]
sd2k: @DanielRamosAcosta I think since #771 `BuildTransport` does actually do those things, so Cursor might be picking up on stuff since the merge of main, and we might need to adjust slightly. Fortunately it should just be deletion of a few lines of code!

DanielRamosAcosta:
@sd2k you were right — after rebasing onto `main` and pulling in #771, I dropped the manual `NewAuthRoundTripper` / `NewOrgIDRoundTripper` / `NewUserAgentTransport` wrapping in `tools/prom_backend_victoriametrics.go`. `BuildTransport` handles all of that now, same as the Cloud Monitoring backend.

While I was in there I also flattened the error branch in `Query` and trimmed a few narrative comments. Ready for another look whenever! 🙌

tools/prom_backend_victoriametrics.go
```diff
@@ -0,0 +1,133 @@
...
+		return nil, fmt.Errorf("query returned status %d: %s", resp.StatusCode, string(body[:min(len(body), 1024)]))
+	}
+
+	var queryResp dsQueryResponse
+	if err := json.Unmarshal(body, &queryResp); err != nil {
+		return nil, fmt.Errorf("unmarshaling response: %w", err)
+	}
+
+	return &queryResp, nil
+}
```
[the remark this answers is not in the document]
DanielRamosAcosta: Good catch — extracted `doDSQuery` to a free function in `prom_backend.go` and switched both Cloud Monitoring and VictoriaMetrics to use it. Net `-35` lines of duplication. Pushed in 51cb744.

tools/prom_backend_victoriametrics.go
```diff
@@ -0,0 +1,95 @@
...
+
+// Query executes a PromQL/MetricsQL query via Grafana's /api/ds/query endpoint.
+// The VM plugin signals instant vs range with two boolean fields on the payload.
+func (b *victoriaMetricsBackend) Query(ctx context.Context, expr string, queryType string, start, end time.Time, stepSeconds int) (model.Value, error) {
+	if start.IsZero() {
+		start = end
+	}
+	if end.IsZero() || end.Before(start) {
+		end = start
+	}
```
sd2k, line 70: if start and end are zero here I think we end up with a zero time query - worth adding a test and maybe defaulting to `time.Now()`.

tools/prom_backend_victoriametrics.go
```diff
@@ -0,0 +1,95 @@
...
+		prometheusBackend: promBackend,
+		httpClient:        client,
+		baseURL:           baseURL,
+		datasourceUID:     ds.UID,
+	}, nil
+}
+
+// Query executes a PromQL/MetricsQL query via Grafana's /api/ds/query endpoint.
+// The VM plugin signals instant vs range with two boolean fields on the payload.
+func (b *victoriaMetricsBackend) Query(ctx context.Context, expr string, queryType string, start, end time.Time, stepSeconds int) (model.Value, error) {
```
sd2k, line 57: We should validate that `queryType` is one of "instant" or "range" here.

sd2k (commented):
This looks great, thank you @DanielRamosAcosta! I have a couple of improvements to request and I can merge once those are fixed.

tools/prom_backend_victoriametrics.go
DanielRamosAcosta: Good point — both start and end zero now falls back to `time.Now()`, with a test covering it. Pushed in dc169a3.

tools/prom_backend_victoriametrics.go
DanielRamosAcosta: Done — added an early `invalid query type` error mirroring the native Prometheus backend, plus a test. Same commit (dc169a3).

sd2k (approved):
Thanks again!

PR #784, opened 2026-04-24
tool: adds support to create datasource via MCP server

jcolladokuri:
## Summary

This PR introduces two new write tools for Grafana datasource management, gated behind the existing enableWrite flag, along with a credential guard layer that enforces a security policy: secrets and authentication credentials can never be sent through MCP — they must always be configured in the Grafana UI.

## New tools
create_datasource
Creates a new Grafana datasource via the API (POST /api/datasources). Accepts standard datasource fields (name, type, url, access, jsonData, etc.) but actively blocks any attempt to pass credentials or authentication configuration through the tool — those are intercepted by the credential guard and redirected to the Grafana UI.

add_authentication_to_datasource
A purpose-built tool for authentication-intent requests. Instead of accepting credentials, it always redirects the agent and user to the Grafana datasource settings page (edit page for an existing UID, new-datasource page otherwise). This is the explicitly supported path for adding passwords, API tokens, basic auth, bearer tokens, and TLS secrets to datasources.

Both tools are registered only when grafana:write scope is granted, matching the existing pattern used by incident tools.

## Credential guard (tools/credentialguard.go)
A new guard layer runs on every create_datasource call and inspects all string input fields (name, type, URL, database, user, and all string values in jsonData) for two categories of policy violation:

Auth intent: regex patterns that detect phrases like "add authentication", "enable basic auth", "username and password", "configure credentials with token", etc. Requires datasource/Grafana context for some patterns to reduce false positives.
Secret-like values: patterns that detect high-confidence secrets — RSA/EC/OPENSSH private keys, AWS AKIA keys, GitHub PATs (ghp_/ghs_), GitLab PATs (glpat-), Slack tokens (xox*), Bearer tokens, and key=value pairs with credential-named keys.
Additionally, basicAuth: true, basicAuthUser, and any secureJsonData fields are blocked unconditionally.

On a violation the tool returns a structured credential_policy_redirect result (marked isError: true) with:

A JSON payload containing outcome, reason, credential_policy, message, and open_config_page_url
An MCP ResourceLink pointing to the Grafana configuration page
A best-effort browser open of that URL (works when server and client are on the same machine)
jsonData schema hook for future dynamic schema
jsonData is typed as a named datasourceJSONData type (underlying map[string]interface{}) that implements jsonschema.JSONSchemaer. The schema comes from a single function, datasourceJSONDataSchema(), which currently returns a static object schema with a description including concrete examples for common plugins (Prometheus, Loki, Elasticsearch, CloudWatch, InfluxDB).

When the Grafana datasource settings API becomes available, replacing the body of datasourceJSONDataSchema() is the only change needed to make schemas dynamic and plugin-specific — no struct tag edits, no changes to tool registration.

## Other changes
- AddDatasourceTools gains an enableWrite bool parameter (matching the AddIncidentTools pattern); main.go is updated to pass the flag.
- Unit tests cover: auth intent matching (blocked and allowed cases), secret detection (blocked and allowed cases), checkDatasourceCredentials policy logic, datasourceConfigPageURL URL construction (edit vs. new, public URL preference, path escaping, scheme validation), credentialViolationResult content shape, and addAuthenticationToDatasource routing logic.
Integration tests cover: successful datasource creation, and credential violations for basicAuth, and secureJsonData.

## Screenshots

Create a datasource:
image

Block input when requested to add authentication:
Screenshot 2026-04-28 at 6 14 31 PM

## Future work

This will be a series of PRs going into `jck/manage-datasources-tool`, then eventually into `main`. As part of the upcoming work is the following:

- Tool to edit existing datasources
- Tool to perform health checks (on demand and when new datasources are created or edited)
- Schema integration into `jsonData` for proper validation 
- Tool to install plugins whenever the requested one is not yet available.

---

> [!NOTE]
> **Medium Risk**
> Adds a new write-capable MCP tool that creates Grafana datasources via the HTTP API and changes tool registration to depend on the write flag; mistakes could allow unintended datasource creation or incorrect URL/link generation.
> 
> **Overview**
> Adds a new `create_datasource` MCP tool that calls Grafana’s datasource create API and returns a structured result including `nextSteps` plus a `resource_link` to the datasource configuration page.
> 
> Updates datasource tool registration (`AddDatasourceTools`) to accept an `enableWrite` flag and only register the new write tool when write operations are enabled; the main server and TLS example are updated to pass this flag.
> 
> Extends unit and integration tests to cover successful datasource creation and validation of the returned JSON payload and configuration-page link behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2c979e5d7b77e197f23fcb29f110360aea3407cd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

Depends on https://github.com/grafana/grafana-assistant-app/pull/6203

### DIFF
--- a/tools/datasources.go
+++ b/tools/datasources.go
@@ -163,6 +163,10 @@ func createDatasource(ctx context.Context, args CreateDatasourceParams) (*mcp.Ca
 	result := &CreateDatasourceResult{
 		Datasource: p.Datasource,
 	}
+
+	if p.Message != nil {
+		result.Message = *p.Message
+	}
 	if p.ID != nil {
 		result.ID = *p.ID
 	}
--- a/tools/datasources.go
+++ b/tools/datasources.go
@@ -38,7 +38,6 @@ type ListDatasourcesResult struct {
 	Datasources []dataSourceSummary `json:"datasources"`
 	Total       int                 `json:"total"`   // Total count before pagination
 	HasMore     bool                `json:"hasMore"` // Whether more results exist
-	Message     string              `json:"message,omitempty"`
 }
 
 func listDatasources(ctx context.Context, args ListDatasourcesParams) (*ListDatasourcesResult, error) {
--- a/tools/datasources.go
+++ b/tools/datasources.go
@@ -2,9 +2,11 @@ package tools
 
 import (
 	"context"
+	"encoding/json"
 	"fmt"
 	"strings"
 
+	"github.com/invopop/jsonschema"
 	"github.com/mark3labs/mcp-go/mcp"
 	"github.com/mark3labs/mcp-go/server"
 
@@ -36,6 +38,7 @@ type ListDatasourcesResult struct {
 	Datasources []dataSourceSummary `json:"datasources"`
 	Total       int                 `json:"total"`   // Total count before pagination
 	HasMore     bool                `json:"hasMore"` // Whether more results exist
+	Message     string              `json:"message,omitempty"`
 }
 
 func listDatasources(ctx context.Context, args ListDatasourcesParams) (*ListDatasourcesResult, error) {
@@ -84,7 +87,118 @@ func listDatasources(ctx context.Context, args ListDatasourcesParams) (*ListData
 		Datasources: summarizeDatasources(paginated),
 		Total:       total,
 		HasMore:     hasMore,
+		Message:     "Data sources retrieved successfully. Please remember to never enter sensitive data via this chat.",
 	}, nil
+}
+
+// datasourceJSONData holds plugin-specific non-secret settings for a datasource.
+// Its schema description is sourced from datasourceJSONDataSchema — swap that
+// function body for an API fetch when the datasource settings API is available.
+type datasourceJSONData map[string]interface{}
+
+func (datasourceJSONData) JSONSchema() *jsonschema.Schema {
+	return datasourceJSONDataSchema()
+}
+
+// datasourceJSONDataSchema returns the JSON schema for the jsonData field.
+// TODO: Replace with a fetch from the Grafana datasource settings API when available.
+func datasourceJSONDataSchema() *jsonschema.Schema {
+	return &jsonschema.Schema{
+		Type: "object",
+		Description: "Datasource-specific non-secret settings. Keys and values vary by plugin type; " +
+			"ask the user what to set or consult the plugin documentation. " +
+			"Examples — prometheus: {\"httpMethod\":\"GET\",\"timeInterval\":\"15s\"}; " +
+			"loki: {\"maxLines\":1000}; elasticsearch: {\"index\":\"my-index\",\"timeField\":\"@timestamp\"}; " +
+			"cloudwatch: {\"defaultRegion\":\"us-east-1\",\"authType\":\"default\"}; " +
+			"influxdb: {\"version\":\"Flux\",\"product\":\"InfluxDB Enterprise 3.x\"}.",
+		Extras: map[string]any{},
+	}
+}
+
+type CreateDatasourceParams struct {
+	Name            string             `json:"name" jsonschema:"required,description=Datasource display name"`
+	Type            string             `json:"type" jsonschema:"required,description=Grafana datasource plugin type\\, for example prometheus"`
+	URL             string             `json:"url,omitempty" jsonschema:"description=Datasource base URL when required by the plugin"`
+	Access          string             `json:"access,omitempty" jsonschema:"description=How Grafana should access the datasource (proxy or direct)"`
+	Database        string             `json:"database,omitempty" jsonschema:"description=Optional database name"`
+	User            string             `json:"user,omitempty" jsonschema:"description=Optional username"`
+	BasicAuth       bool               `json:"basicAuth,omitempty" jsonschema:"description=Whether Grafana should use basic auth"`
+	WithCredentials bool               `json:"withCredentials,omitempty" jsonschema:"description=Whether Grafana should forward credentials such as cookies"`
+	IsDefault       bool               `json:"isDefault,omitempty" jsonschema:"description=Whether this should become the default datasource"`
+	JSONData        datasourceJSONData `json:"jsonData,omitempty"`
+}
+
+type CreateDatasourceResult struct {
+	Message    string             `json:"message"`
+	ID         int64              `json:"id"`
+	UID        string             `json:"uid"`
+	Name       string             `json:"name"`
+	Datasource *models.DataSource `json:"datasource,omitempty"`
+	NextSteps  string             `json:"nextSteps,omitempty"`
+}
+
+func createDatasource(ctx context.Context, args CreateDatasourceParams) (*mcp.CallToolResult, error) {
+	dsAccess := args.Access
+	if dsAccess == "" {
+		// use grafana default
+		dsAccess = "proxy"
+	}
+	c := mcpgrafana.GrafanaClientFromContext(ctx)
+	body := &models.AddDataSourceCommand{
+		Name:            args.Name,
+		Type:            args.Type,
+		URL:             args.URL,
+		Access:          models.DsAccess(dsAccess),
+		Database:        args.Database,
+		BasicAuth:       args.BasicAuth,
+		IsDefault:       args.IsDefault,
+		JSONData:        models.JSON(args.JSONData),
+		WithCredentials: args.WithCredentials,
+	}
+	resp, err := c.Datasources.AddDataSourceWithParams(
+		datasources.NewAddDataSourceParamsWithContext(ctx).WithBody(body),
+	)
+	if err != nil {
+		return nil, fmt.Errorf("create datasource: %w", err)
+	}
+	p := resp.Payload
+	result := &CreateDatasourceResult{
+		Datasource: p.Datasource,
+	}
+	if p.Message != nil {
+		result.Message = fmt.Sprintf("%s. Please remember to never enter sensitive data via this chat.", *p.Message)
+	}
+	if p.ID != nil {
+		result.ID = *p.ID
+	}
+	if p.Name != nil {
+		result.Name = *p.Name
+	}
+	if p.Datasource != nil {
+		result.UID = p.Datasource.UID
+	}
+	if result.UID != "" {
+		grafanaURL := mcpgrafana.GrafanaConfigFromContext(ctx).URL
+		configPageURL := fmt.Sprintf("%s/connections/datasources/edit/%s", grafanaURL, result.UID)
+		result.NextSteps = fmt.Sprintf("Visit the datasource configuration page to finish setting it up: %s", configPageURL)
+		b, err := json.Marshal(result)
+		if err != nil {
+			return nil, fmt.Errorf("marshal result: %w", err)
+		}
+		toolResult := mcp.NewToolResultText(string(b))
+		toolResult.Content = append(toolResult.Content, mcp.ResourceLink{
+			Type:        "resource_link",
+			URI:         configPageURL,
+			Name:        result.Name,
+			Description: "Datasource configuration page",
+		})
+		return toolResult, nil
+	}
+	b, err := json.Marshal(result)
+	if err != nil {
+		return nil, fmt.Errorf("marshal result: %w", err)
+	}
+	return mcp.NewToolResultText(string(b)), nil
 }
 
 // filterDatasources returns only datasources of the specified type `t`. If `t`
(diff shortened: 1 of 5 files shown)


jcolladokuri:
@cursor push fe31e46dc1

jcolladokuri:
@cursor push 67ac59e66b

jcolladokuri:
@cursor push 0096693052

tools/credentialguard.go
```diff
@@ -0,0 +1,222 @@
...
+		cmd = exec.Command("xdg-open", rawURL)
+	}
+	if err := cmd.Start(); err != nil {
+		return err
+	}
+	go func() {
+		_ = cmd.Wait()
+	}()
+	return nil
+}
```
sd2k: This won't work as expected in cases where the MCP server is running remotely.

Rather than having two tools, I think a reasonable flow would be for the 'create datasource' tool to create the datasource under the hood, then return a URL to the datasource edit page (built from the Grafana URL) and hint that the model should forward that URL to the user with instructions to add the credentials manually. That would mean we only need 1 tool and this should work even for remote/hosted instances of the MCP server.

tools/datasources.go
```diff
@@ -123,6 +222,24 @@ var ListDatasources = mcpgrafana.MustTool(
...
+)
+
+var AddAuthenticationToDatasource = mcpgrafana.MustTool(
+	"add_authentication_to_datasource",
+	"Use this when the user asks to add, set, configure, or rotate authentication for a Grafana datasource—passwords, API tokens, secrets, basic auth, bearer tokens, or TLS secrets—including Prometheus, Loki, or any other plugin type. Opens the Grafana UI to the datasource settings page (pass uid from list_datasources or get_datasource for an existing datasource; omit uid to open the new datasource page). Does not accept credential values through MCP. Do not use create_datasource for authentication or secrets; this server blocks those fields on create and this tool is the supported path for that intent. If credentials or PII is entered, remind the user to rotate and revoke them to keep them safe.",
+	addAuthenticationToDatasource,
+	mcp.WithTitleAnnotation("Add authentication to datasource"),
+	mcp.WithIdempotentHintAnnotation(false),
+	mcp.WithReadOnlyHintAnnotation(false),
+)
```
sd2k: Mentioned elsewhere but I'm not sure this is needed 🤔

sd2k (changes requested):

tools/credentialguard.go
jcolladokuri: So we have a combination of both, this openBrowser is best effort only for when the server is running in the same environment. But in addition to this, the configuration URL is returned as a resource link that the client can surface. I was debating on keeping this `openBrowser` since I don't actually know how helpful it actually is given that I imagine most of the time the server will be ran as remote.

You do make a good call out on still creating the datasource and then directing the user to the config page, I will make this change so that creation goes through without any creds.

tools/datasources.go
jcolladokuri: I added this tool was because this was the only way that I could more consistently get MCP clients to block auth attempts right away when the prompt was not part of a "create" flow (i.e. straight up asking `Add authentication to my prometheus datasource`), this is more relevant to the edit flow.

Without this, I saw that the client would attempt to gather information (username + password), then call the tool for editing the datasource, at which point, blocking felt more irrelevant given that I had already been asked to give my credentials.

I guess it is more of a precaution than anything else to try to cover all scenarios - perhaps there are better ways to handle this and I am open to ideas, what do you think?

tools/credentialguard.go
jcolladokuri: Okay I have changed it so that whenever create + creds are detected, we still go ahead and create the datasource but then output the URL so that they can go and add the creds themselves:

image


Let me know what you think!

tools/datasources.go
```diff
@@ -87,6 +89,101 @@ func listDatasources(ctx context.Context, args ListDatasourcesParams) (*ListData
...
+		dsAccess = "proxy"
+	}
+	c := mcpgrafana.GrafanaClientFromContext(ctx)
+	body := &models.AddDataSourceCommand{
+		Name:      args.Name,
+		Type:      args.Type,
+		URL:       args.URL,
+		Access:    models.DsAccess(dsAccess),
+		Database:  args.Database,
+		BasicAuth: args.BasicAuth,
```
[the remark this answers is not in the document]
jcolladokuri: @cursor This is fine and expected as it only sets authentication to Basic authentication but it does not actually set any secure fields.

jcolladokuri:
@cursor push bd93073069

tools/credentialguard.go
```diff
@@ -0,0 +1,251 @@
...
+		}
+		if matchesAuthIntent(s) {
+			return "auth_credential_instructions"
+		}
+		if matchesSecretLike(s) {
+			return "embedded_secret_or_token"
+		}
+	}
+	return ""
+}
```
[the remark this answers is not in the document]
jcolladokuri: This is expected and okay, we can set `basicAuth` flag without exposing sensitive data or PII

RafaelPaulovic:
@codex review

tools/credentialguard.go
```diff
@@ -0,0 +1,251 @@
...
+
+// checkDatasourceCredentials checks CreateDatasourceParams for credential policy violations.
+// It mirrors DatasourceInputCredentialViolation from mcp-manage-datasources.
+// Returns a reason code or "" if no violation.
+func checkDatasourceCredentials(args CreateDatasourceParams) string {
+	if args.BasicAuthUser != "" {
+		return "basic_auth_user_via_mcp_disallowed"
+	}
+	if len(args.SecureJSONData) > 0 {
+		return "secure_json_data_found"
```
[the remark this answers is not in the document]
jcolladokuri: This is okay and expected, we still block on setting actual credentials, but if it is just the flag that is being set, we allow this and then prompt the user to set them in the created datasource UI via URL

tools/datasources.go
```diff
@@ -87,6 +89,107 @@ func listDatasources(ctx context.Context, args ListDatasourcesParams) (*ListData
...
+		IsDefault: args.IsDefault,
+		JSONData:  models.JSON(args.JSONData),
```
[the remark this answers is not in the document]
jcolladokuri: since we are not configuring anything that could be PII or sensitive data, it is okay to ignore args.User. However, since withCredentials is just a flag similar to args.basicAuth, it is okay to set this.

jcolladokuri:
@cursor push cbed96e7b0

jcolladokuri:
@cursor push 365f2591e4

tools/credentialguard.go
```diff
@@ -0,0 +1,251 @@
...
+// checkDatasourceCredentials checks CreateDatasourceParams for credential policy violations.
+// It mirrors DatasourceInputCredentialViolation from mcp-manage-datasources.
+// Returns a reason code or "" if no violation.
+func checkDatasourceCredentials(args CreateDatasourceParams) string {
+	credentialFieldReason := ""
+	if args.BasicAuthUser != "" {
+		credentialFieldReason = "basic_auth_user_via_mcp_disallowed"
+	} else if len(args.SecureJSONData) > 0 {
+		credentialFieldReason = "secure_json_data_found"
+	}
```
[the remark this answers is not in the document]
jcolladokuri: I believe this should be okay, in general the messaging stays consistent that credentials and sensitive info is omitted, while the rest is not.

sd2k (commented):
Thanks for adjusting!

My concern with the credential guard part is that by the time it kicks in, it's already too late - in the flow we're looking at, the user has already sent a message including the credentials to the LLM asking it to set up the datasource, so the creds are already part of the conversation, and just denying the MCP tool call won't affect that.

I guess we can suggest that the user rotates the password but realistically I'm not sure how likely they are to do that if they are already relying on AI to set up their datasources 😬 

Part of me is a bit wary of adding this capability at all since it might even mislead/encourage users into posting creds into the chat as part of a setup flow. What do you think about that concern, is it valid or am I just being overly cautious?

jcolladokuri:
@sd2k not being overly cautious! When it comes to things like these, I think it is better.

I share your concern and cautiousness, and that is part of the reason why I had added the add_authentication tool separately so that it could help handle more cases and catch auth attempts earlier when possible.

That being said, I don't think we should block on adding this altogether based on that, and I think we can try and guard as much as possible around it. In the end we can't fully control what users will end up entering in chat when using this server:

- Document that credentials are not supported in this MCP server and should never be entered via chat
- As part of a later PR we were thinking of adding some sort of warning or messaging to all tool results to try and remind user of this. I have added this, claude has been consistently ignoring the part about credentials :,)

Screenshot 2026-05-05 at 8 52 48 AM

- Like you mentioned, also suggest to the user that they rotate the credentials they have entered in chat - I have added this as well.

Screenshot 2026-05-05 at 8 54 21 AM

- As part of what we are already doing, continue to ignore secure fields so that they are never actually sent to Grafana.

Let me know what you think!

jcolladokuri:
@cursor push 6564c1cd12

tools/datasources.go
```diff
@@ -84,9 +87,112 @@ func listDatasources(ctx context.Context, args ListDatasourcesParams) (*ListData
...
+		return credentialViolationResult(credentialViolationReason, datasourceConfigPageURL(ctx, "")), nil
+	}
+	dsAccess := args.Access
+	if dsAccess == "" {
+		// use grafana default
+		dsAccess = "proxy"
+	}
+	c := mcpgrafana.GrafanaClientFromContext(ctx)
+	body := &models.AddDataSourceCommand{
+		Name:            args.Name,
```
yesoreyeram, line 146: We should check for invalid/empty Name and Type

jcolladokuri: For this case I don't think we need to check for empty Name or Type . Both fields are declared as required in the jsonSchema which essentially tells the MCP client to enforce these fields. If an empty value does slip through, the Grafana API would catch it and return an error that we would expose in line 168.

I personally don't think it is necessary to add more code on our side to handle this, but happy to discuss if you feel strongly about it.

As for a valid Type, I think it makes more sense for it to be part of later work when we connect this tool with the plugin management tools so that we can leverage getting plugin info and checking the type directly there.

What do you think?

jcolladokuri:
@cursor push a6bbcc6e1e

tools/datasources.go
yesoreyeram: not a big blocker but having such validations may save some roundtrips/tokens. :)

yesoreyeram (approved):
LGTM. Good work @jcolladokuri

tools/datasources.go
```diff
@@ -36,6 +38,7 @@ type ListDatasourcesResult struct {
 	Datasources []dataSourceSummary `json:"datasources"`
 	Total       int                 `json:"total"`   // Total count before pagination
 	HasMore     bool                `json:"hasMore"` // Whether more results exist
+	Message     string              `json:"message,omitempty"`
```
sd2k:
```suggestion
```

tools/datasources.go
```diff
@@ -84,9 +87,123 @@ func listDatasources(ctx context.Context, args ListDatasourcesParams) (*ListData
 		Datasources: summarizeDatasources(paginated),
 		Total:       total,
 		HasMore:     hasMore,
+		Message:     "Data sources retrieved successfully. Please remember to never enter sensitive data via this chat.",
```
sd2k:
```suggestion
```

tools/datasources.go
```diff
@@ -84,9 +87,123 @@ func listDatasources(ctx context.Context, args ListDatasourcesParams) (*ListData
...
+	if err != nil {
+		return nil, fmt.Errorf("create datasource: %w", err)
+	}
+	p := resp.Payload
+	result := &CreateDatasourceResult{
+		Datasource: p.Datasource,
+	}
+	if p.Message != nil {
+		result.Message = fmt.Sprintf("%s. Please remember to never enter sensitive data via this chat.", *p.Message)
+	}
```
sd2k, line 169:
```suggestion
```

sd2k (commented):
Generally LGTM!

I think we should get rid of the various 'sensitive data' warnings and tackle them in a separate PR if we're worried about them, IMO they're orthogonal to this tool and I'm not convinced they would help much given that they're only returned _after_ a tool call.

tools/datasources.go
```diff
@@ -87,6 +89,116 @@ func listDatasources(ctx context.Context, args ListDatasourcesParams) (*ListData
...
+	if p.ID != nil {
+		result.ID = *p.ID
+	}
+	if p.Name != nil {
+		result.Name = *p.Name
+	}
+	if p.Datasource != nil {
+		result.UID = p.Datasource.UID
+	}
```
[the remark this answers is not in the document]
jcolladokuri: already added.

tools/datasources_unit_test.go
```diff
@@ -227,3 +229,65 @@ func TestGetDatasource_ErrorWhenNeitherProvided(t *testing.T) {
...
+	require.Len(t, toolResult.Content, 2)
+
+	text, ok := toolResult.Content[0].(mcp.TextContent)
+	require.True(t, ok)
+
+	var got CreateDatasourceResult
+	require.NoError(t, json.Unmarshal([]byte(text.Text), &got))
+	assert.Equal(t, uid, got.UID)
+	assert.Equal(t, name, got.Name)
+	assert.Equal(t, msg+". Please remember to never enter sensitive data via this chat.", got.Message)
```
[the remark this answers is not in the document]
jcolladokuri: fixed

sd2k (approved):
Thanks!

PR #804, opened 2026-04-27
fix(loki): unquote timestamp in query_loki_logs result (#803)

### LINKED ISSUE #803: query_loki_logs returns malformed JSON in content[0].text when log lines contain quotes
## Bug: `query_loki_logs` returns malformed JSON in `content[0].text` when log lines contain quotes

When `query_loki_logs` returns log lines that themselves contain JSON (i.e. structured logs emitted as a JSON string by the application), the `text` payload inside the tool result becomes invalid JSON — values like `timestamp` are wrapped in extra escaped quotes, and inner string values are not properly escaped, so a standard `json.loads()` on the content fails.

This is silent: most clients (including JSON-RPC parsers in language SDKs) just see the outer envelope as valid and the inner content as a string, but the moment a downstream caller tries to parse that string as the documented array shape, it errors out and **the matches look like "0 results"**.

## Versions

- Helm chart: `grafana/grafana-mcp` **0.3.1**
- App image: `grafana/mcp-grafana` **0.8.2**
- Transport: `streamable-http`
- Loki: `6.55.0` chart, schema v13, store TSDB+S3
- Tool args: `--disable-write --enabled-tools=datasource,loki,prometheus,dashboard,search,navigation,tempo,pyroscope`

## Steps to reproduce

1. Have a Loki tenant where log lines are JSON-formatted strings (typical for structured loggers like zap/logrus/zerolog), e.g.:
   ```
   {"@timestamp":"2025-01-01T00:00:00Z","level":"INFO","app_name":"myapp","verb":"PUT","resultStatus":200,"message":"PUT /api/foo"}
   ```
2. Call `query_loki_logs` with a `|=` line filter that matches at least one such line:
   ```jsonc
   {
     "name": "query_loki_logs",
     "arguments": {
       "datasourceUid": "loki",
       "logql": "{service_name=\"myapp\"} |= \"some-unique-substring\"",
       "startRfc3339": "2025-01-01T00:00:00Z",
       "endRfc3339":   "2025-01-02T00:00:00Z",
       "limit": 5,
       "direction": "backward"
     }
   }
   ```
3. Read `result.content[0].text` and try `JSON.parse(...)` / `json.loads(...)`.

## Expected

`result.content[0].text` is a JSON-encoded array, e.g.:

```json
[
  {
    "timestamp": "1234567890123456789",
    "labels": {"service_name": "myapp", "...": "..."},
    "line": "{\"@timestamp\":\"2025-01-01T00:00:00Z\",\"level\":\"INFO\",\"verb\":\"PUT\",\"resultStatus\":200}"
  }
]
```

## Actual

A near-identical-looking payload, but with the timestamp value wrapped in extra escaped quotes and inner string values broken:

```text
[{\"timestamp\":\"\\"1234567890123456789\\"\",\"line\":\"{\\"@timestamp\\":\\"2025-01-01T00:00:00Z\\",\\"level\\":\\"INFO\\",\\"app_name\\":\\"myapp\\",\\"verb\\":\\"PUT\\",\\"resultStatus\\":200, ... \\"hostname\\":\\"app-myapp-abc-y
```

After unescaping the JSON-RPC envelope once (i.e. doing `\"` → `"`), you get this for the inner string:

```text
[{"timestamp":"\"1234567890123456789\"","line":"{\"@timestamp\":\"2025-01-01T00:00:00Z\",...
```

Note:
- `"timestamp":"\"1234567890123456789\""` — the value is a JSON string that itself starts with `\"` and ends with `\""`, so a parser sees `"\"1234567890123456789\""` as `"\""` (string containing one `"`) followed by garbage, and fails at column 90 with `Expecting ',' delimiter`.
- Inner JSON in `line` reuses single-level escaping (`\"@timestamp\":\"...\"`) but the surrounding string was already escaped for the envelope, so the result is one level short.

## Result for users

Downstream tooling that reasonably does `json.loads(content[0].text)` reports **0 matches** — except matches *do* exist; they are just unparseable. We hit this in production while bisecting a 30-day search for a specific structured-metadata `trace_req_id`: every 4-day slice returned `parse error` instead of the actual hits, until we extracted timestamps and key fields with regex from the malformed payload.

## Likely root cause (educated guess from the output)

Looks like the timestamp is being `Marshal`'d a second time, and the `line` value is being concatenated as raw bytes instead of being decoded → re-encoded. So instead of:

```go
content := mustJSON([]Result{{Timestamp: "1234...", Line: rawJSONLine}})
```

something like:

```go
content := mustJSON([]Result{{Timestamp: mustJSON("1234..."), Line: rawJSONLine}}) // double-encoded
```

is happening for `Timestamp`, and `Line` is being injected without escaping its quotes for the outer string context.

## Workaround

Extract structured fields from `result.content[0].text` with regex (e.g. `"@timestamp":"([^"]+)"`, `"trace_url":"([^"]+)"`) instead of relying on `json.loads`. Not ideal — drops some lines and doesn't scale.

## Impact

High signal-vs-noise problem: tool calls return ✅ HTTP 200, the envelope parses fine, the outer schema looks correct — but agents downstream silently treat real matches as zero. For "needle in a haystack" investigations (rare error correlation, single trace lookup, audit), this leads agents to wrong conclusions.

Happy to test a fix on our end if helpful.

jeffersonlmartins:
## Summary

Fixes #803.

`value[0]` from a Loki streams response is a `json.RawMessage` containing a JSON-encoded string (`"1234567890"` — quotes included). The current code does `string(value[0])` which preserves those literal quotes, so `LogEntry.Timestamp` ends up holding `"1234567890"` instead of `1234567890`. When the entry array is marshalled into the tool result's `content[0].text`, `encoding/json` faithfully escapes the embedded quotes, producing:

```text
"timestamp":"\"1234567890\""
```

After the outer JSON-RPC envelope is unescaped once, the resulting string is no longer parseable JSON, and clients that follow the documented `JSON.parse(content[0].text)` pattern silently get a parse error. Real matches end up looking like 0 results.

The same code path already does the right thing for `value[1]` (the log line) with `json.Unmarshal(value[1], &logLine)`. This change applies the same unmarshalling to `value[0]`, stripping the surrounding quotes before storing the timestamp.

## Diff at the call site

```go
// before
entry := LogEntry{
    Timestamp: string(value[0]), // keeps the literal quotes
    Line:      logLine,
    Labels:    stream.Stream,
}

// after
var ts string
if err := json.Unmarshal(value[0], &ts); err != nil {
    continue
}
entry := LogEntry{
    Timestamp: ts,           // unquoted nanoseconds-as-string
    Line:      logLine,
    Labels:    stream.Stream,
}
```

## Test plan

- [x] `TestLogEntryTimestampNotDoubleEncoded` (new) — asserts (a) the timestamp is plain digits, no embedded `"`, (b) the serialised array contains `"timestamp":"<digits>"` and never `"timestamp":"\"`, and (c) `json.Unmarshal` round-trips the marshalled output back into `[]LogEntry` without any further unescaping.
- [x] `TestEnforceLogLimit`, `TestHasCategorizeLabelsFlag`, `TestCategorizedLabelsParsing`, `TestCategorizedLabelsBackwardCompat` all still pass — change is fully behaviour-preserving for the rest of the parsing pipeline.
- Manually verified against a live Loki tenant (chart 6.55.0, schema v13) with structured-metadata-bearing log lines.

## Notes

The fix is intentionally minimal: only the timestamp parse changes, the rest of the function is untouched. No public API changes.

---

> [!NOTE]
> **Low Risk**
> Low risk: a small parsing change isolated to Loki stream results plus a regression test; main risk is skipping entries if a timestamp is malformed.
> 
> **Overview**
> Fixes `query_loki_logs` stream parsing so the timestamp (`values[0]`) is `json.Unmarshal`’d into a plain string instead of using `string(raw)`, preventing *double-encoded quotes* in `LogEntry.Timestamp` that break downstream JSON parsing.
> 
> Adds a regression unit test (`TestLogEntryTimestampNotDoubleEncoded`) that asserts the marshalled `[]LogEntry` output contains a normal `"timestamp":"<digits>"` value and round-trips through `json.Unmarshal`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0e8698a83c321344e32cd09d10489cd9e1293bad. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tools/loki.go
+++ b/tools/loki.go
@@ -592,6 +592,15 @@ func queryLokiLogs(ctx context.Context, args QueryLokiLogsParams) (*QueryLokiLog
 		for _, stream := range streams {
 			for _, value := range stream.Values {
 				if len(value) >= 2 {
+					// Parse timestamp. value[0] is a json.RawMessage containing
+					// a JSON-encoded string like `"1234567890"`; using string()
+					// directly would keep the surrounding quotes and double-encode
+					// when the LogEntry is marshalled (see issue #803).
+					var ts string
+					if err := json.Unmarshal(value[0], &ts); err != nil {
+						continue // Skip invalid timestamps
+					}
+
 					// Parse log line
 					var logLine string
 					if err := json.Unmarshal(value[1], &logLine); err != nil {
@@ -599,7 +608,7 @@ func queryLokiLogs(ctx context.Context, args QueryLokiLogsParams) (*QueryLokiLog
 					}
 
 					entry := LogEntry{
-						Timestamp: string(value[0]), // Nanoseconds as string
+						Timestamp: ts, // Nanoseconds as string (unquoted)
 						Line:      logLine,
 						Labels:    stream.Stream,
 					}
--- a/tools/loki_unit_test.go
+++ b/tools/loki_unit_test.go
@@ -158,6 +158,69 @@ func TestCategorizedLabelsParsing(t *testing.T) {
 	assert.Equal(t, map[string]string{"level": "error"}, cats2.Parsed)
 }
 
+// Regression test for #803: timestamps in `value[0]` arrive as JSON-encoded
+// strings (e.g. `"1234567890"`) and must be unquoted before being placed in
+// LogEntry.Timestamp; otherwise the surrounding quotes are double-encoded
+// when the entry is marshalled, producing `"timestamp":"\"1234567890\""` and
+// breaking any downstream `json.Unmarshal` of the tool result.
+func TestLogEntryTimestampNotDoubleEncoded(t *testing.T) {
+	rawResponse := `{
+		"status": "success",
+		"data": {
+			"resultType": "streams",
+			"result": [
+				{
+					"stream": {"app": "frontend"},
+					"values": [
+						["1693996529000222496", "some log line"]
+					]
+				}
+			]
+		}
+	}`
+
+	var response lokiQueryResponse
+	require.NoError(t, json.Unmarshal([]byte(rawResponse), &response))
+
+	var streams []LokiLogStream
+	require.NoError(t, json.Unmarshal(response.Data.Result, &streams))
+	require.Len(t, streams, 1)
+	require.Len(t, streams[0].Values, 1)
+
+	// Mirror the parsing the production code does in queryLokiLogs.
+	var ts string
+	require.NoError(t, json.Unmarshal(streams[0].Values[0][0], &ts))
+
+	var line string
+	require.NoError(t, json.Unmarshal(streams[0].Values[0][1], &line))
+
+	entry := LogEntry{
+		Timestamp: ts,
+		Line:      line,
+		Labels:    streams[0].Stream,
+	}
+
+	// Timestamp must be the plain numeric string, with no surrounding quotes.
+	assert.Equal(t, "1693996529000222496", entry.Timestamp)
+	assert.NotContains(t, entry.Timestamp, `"`,
+		"timestamp must not contain literal quotes — see #803")
+
+	// Marshalling the entry must produce a JSON object that can be
+	// re-parsed with json.Unmarshal without further unescaping.
+	out, err := json.Marshal([]LogEntry{entry})
+	require.NoError(t, err)
+	assert.Contains(t, string(out), `"timestamp":"1693996529000222496"`,
+		"timestamp must serialise as a plain string, not as a JSON-encoded JSON string")
+	assert.NotContains(t, string(out), `"timestamp":"\"`,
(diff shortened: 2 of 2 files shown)


sd2k (approved):
LGTM, thanks! I'll get this merged once the CLA is signed.

gerboland:
Thank you for the contribution, however as we've not heard from you in some time and the CLA was not signed, I'll close this in favour of https://github.com/grafana/mcp-grafana/pull/982

Issue #795, opened 2026-04-26
First-class tools for Mimir + Loki recording rules (separate from alerting rules)


QuentinBisson:
## Summary                                  
               
Add four MCP tools covering the Prometheus/Mimir and Loki ruler APIs for
recording rules: `list_mimir_recording_rules`, `get_mimir_recording_rule`,                                                                                                                                                                                                    
`list_loki_recording_rules`, `get_loki_recording_rule`.
                                                                                                                                                                                                                                                                                
## Motivation                                                 
                                                                    
Recording rules are a distinct shape from alerting rules:     
- No `for` / `keep_firing_for` semantics.          
- No Alertmanager plumbing or annotations.                                                                                                                                                                                                                                    
- The dashboard / debugging questions are different ("what does this
  pre-aggregated metric come from?" rather than "is this firing?").                                                                                                                                                                                                           
                                                                    
Today the existing rule tools mix both shapes, which makes the LLM's job                                                                                                                                                                                                      
harder than it needs to be — it has to filter every rule list to find
recording rules, and the response payloads carry alerting-only fields                                                                                                                                                                                                         
that are noise.                                                   
                                                                                                                                                                                                                                                                                
## Proposal                                                                                                                                                                                                                                                                   
                                                                                                                                                                                                                                                                                
Four read-only tools, mirroring the existing alerting-rule shape:                                                                                                                                                                                                             
                                                                                                                                                                                                                                                                                
| Tool | Maps to |                                                                                                                                                                                                                                                            
|---|---|                                                                                                                                                                                                                                                                     
| `list_mimir_recording_rules` | Mimir ruler `/rules` filtered to type=recording |                                                                                                                                                                                            
| `get_mimir_recording_rule` | one rule by group + name |                                                                                                                                                                                                                     
| `list_loki_recording_rules` | Loki ruler `/rules` filtered to type=recording |                                                                                                                                                                                              
| `get_loki_recording_rule` | one rule by group + name |
                                                                                                                                                                                                                                                                                
Each accepts a datasource selector (uid or name match) and a tenant
header passthrough that the existing tools already plumb.

PR #819, opened 2026-04-29
fix(alerting): include recording rules in datasource ruler listings

QuentinBisson:
## Summary

`alerting_manage_rules` with `datasource_uid` set silently drops every recording rule from the response, even when the caller explicitly passes `rule_type=recording`. The drop happens in `convertPrometheusRulesToSummary` (`tools/alerting_manage_rules_datasource.go`), which only emits `v1.AlertingRule` cases and `continue`s past `v1.RecordingRule`.

## Changes

- Emit recording rules with the fields that apply to them: `query`, `labels`, `health`, `last_evaluation`. Alerting-only fields (`state`, `for`, `annotations`) stay empty.
- Tag every datasource-managed rule with `type` (`"alerting"` | `"recording"`) so the LLM can branch without re-deriving from missing fields.
- Surface `query` on alerting rules from the datasource path too — previously only `get` exposed the expression.
- Mark `state` `omitempty` (recording rules never have one).

Touches the same code path that already serves Loki and Mimir rule listings via `/api/prometheus/{uid}/api/v1/rules`; no new tools, no new endpoints.

Refs #795.

---

> [!NOTE]
> **Medium Risk**
> Changes the shape and contents of the `list` response for datasource-managed rules (new `type`/`query`, recording rules included), which may affect downstream consumers expecting only alerting rules.
> 
> **Overview**
> Fixes `alerting_manage_rules` *datasource-managed* `list` results to include **recording rules** (previously dropped), and tags each returned rule with a `type` of `alerting` or `recording`.
> 
> Datasource rule summaries now also expose the rule `query`, normalize/omit empty label/annotation maps, and omit `state`/`last_evaluation` when not present. Because Grafana’s datasource proxy may not honor `rule_type`, the PR enforces `rule_type` filtering client-side for consistent behavior, tightens the `rule_type` JSON schema to an enum, and adds unit coverage for the new rule-type filtering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 32a1b14667dcd6f776ebac9358dd0cf6f58faf74. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


sd2k (approved):
LGTM, thanks!

PR #821, opened 2026-04-29
fix(alerting): query datasource rules via proxy with per-type path

QuentinBisson:
## Summary

`GetDatasourceRules` currently hits Grafana's legacy unified endpoint `/api/prometheus/{uid}/api/v1/rules`. That endpoint applies an internal path template that does not compose with datasource URLs already carrying a backend prefix — e.g. a Mimir gateway DS configured with `http://mimir-gateway/prometheus` produces 404s, even though hitting the same datasource through the generic DS proxy at `/api/datasources/proxy/uid/{uid}/api/v1/rules` returns rules correctly.

The inline comment in `alerting_client.go` already flagged this: *"use the Grafana unified endpoint - maybe we need to use the datasource proxy endpoint in the future as this is an api for internal use"*. This PR makes that switch, mirroring how `GetAlertmanagerConfig` already operates.

The path appended through the proxy depends on the datasource type, because the two Grafana DS plugins handle prefixes differently:

| DS type | Path appended | Reason |
|---|---|---|
| prometheus / mimir / cortex | `/api/v1/rules` | Prom plugin doesn't add a prefix; any `/prometheus` is in the DS URL itself |
| loki | `/prometheus/api/v1/rules` | Loki plugin hardcodes `/loki/...` so the DS URL is the gateway root; `/prometheus/api/v1/rules` is Loki's Prom-compat ruler endpoint that returns the `{status, data: {groups: [...]}}` shape the decoder expects (the native `/loki/api/v1/rules` returns YAML — incompatible) |

The caller `listDatasourceAlertRules` already fetches the datasource record (for `isRulerDatasource`) and now threads `ds.Type` through.

Independent of #819 (recording-rules emission); the two PRs touch different files.

### DIFF
--- a/tools/alerting_client.go
+++ b/tools/alerting_client.go
@@ -219,11 +219,24 @@ type alert struct {
 	Value       string        `json:"value"`
 }
 
-// GetDatasourceRules queries a datasource's Prometheus ruler API
-func (c *alertingClient) GetDatasourceRules(ctx context.Context, datasourceUID string, opts *GetRulesOpts) (*v1.RulesResult, error) {
-	// use the Grafana unified endpoint - maybe we need to use the datasource proxy endpoint in the future as this
-	// is an api for internal use
-	path := fmt.Sprintf("/api/prometheus/%s/api/v1/rules", datasourceUID)
+// GetDatasourceRules queries a datasource's Prometheus ruler API via the
+// generic Grafana datasource proxy. The proxy appends the request path to
+// the configured datasource URL, which already carries any backend-specific
+// prefix (e.g. Mimir gateway URLs that end in /prometheus). The legacy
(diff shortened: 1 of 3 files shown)


QuentinBisson:
Closing — after reading Grafana's lotex source (`pkg/services/ngalert/api/lotex_prom.go`, `util.go`), the unified endpoint and the generic DS proxy emit identical upstream URLs for Prometheus-compatible and Loki datasources. The path-doubling I described in the original framing doesn't actually happen inside Grafana, so this PR isn't a bug fix — it would be a consistency refactor at best, and would also drop the editor-role elevation that the lotex proxy applies (potential regression). Will revisit if a real root cause for the original 404 surfaces.

PR #828, opened 2026-05-01
Expose opt-in last alert payload on alert groups

### LINKED ISSUE #814: Feature Request: Expose last_alert payload in get_alert_group response
## Summary

The Grafana OnCall API already returns a `last_alert` object (including full payload) inside the alert group response, but the Go client's `AlertGroup` struct doesn't map it — so `get_alert_group` silently drops it today.

## The Fix

1. Add `last_alert` to the `AlertGroup` struct in `amixr-api-go-client`:

```go
type AlertGroup struct {
    // existing fields ...
    LastAlert *Alert `json:"last_alert,omitempty"`
}
```

2. Expose it in `get_alert_group` as opt-in to avoid polluting MCP context by default:

```go
type GetAlertGroupParams struct {
    AlertGroupID           string `json:"alertGroupId" jsonschema:"required,description=The ID of the alert group to retrieve"`
    IncludeLastAlertPayload bool  `json:"includeLastAlertPayload,omitempty" jsonschema:"description=If true, includes the full payload of the last alert in the group"`
}
```

## Motivation

Alert payloads contain stable fingerprints that an AI agent can use to correlate recurring alerts across different alert group IDs:

| Source | Field | Example |
|---|---|---|
| Sentry | `payload.data.event.hashes[0]` | `"66b46acbdeae7d18599d803d44d7c10f"` |
| groundcover / Grafana Alertmanager | `payload.alerts[0].fingerprint` | `"912a47a2bcc2c22e"` |

Today the agent sees only `alerts_count: 1` with no way to reach the payload in a single call. With this fix, one `get_alert_group` call returns the fingerprint needed for correlation — no additional tool required.

The `includeLastAlertPayload` flag is intentional: a Sentry alert payload can exceed 10KB (full stacktrace, 50+ module versions, request headers). Opt-in avoids polluting the MCP context when the payload isn't needed.

## Related

- #748 — update_incident
- #813 — update_alert_group
- Follow-up: a `list_alerts` tool (with field selection) will be proposed separately for cases where the full alert history of a group is needed

Genmin:
Fixes #814.

## Summary
- add `includeLastAlertPayload` to `get_alert_group` so the full `last_alert` payload can be requested only when needed
- keep the default response compact by omitting `last_alert` unless explicitly requested
- decode `last_alert.payload` into an arbitrary JSON map so source-specific fingerprints and nested fields are preserved

## Validation
- `gofmt -w tools/oncall.go tools/oncall_unit_test.go`
- `git diff --check`
- `go test ./tools -run 'TestGetAlertGroup'`\n- `go test ./tools`\n- `go test ./...`\n- `make lint-jsonschema lint-openapi`\n\nNote: `golangci-lint` is not installed in my local environment, so I could not run `golangci-lint run` locally.

---

> [!NOTE]
> **Medium Risk**
> Changes the `get_alert_group` tool response shape by wrapping results in `GetAlertGroupResult` (with optional `last_alert`), which may affect downstream consumers expecting the prior `OnCallAlertGroup` payload. Uses a lower-level HTTP request/JSON decode path for the public API, so correctness depends on the OnCall API response format.
> 
> **Overview**
> `get_alert_group` now returns a new `GetAlertGroupResult` that embeds the existing alert group fields and, when using the public OnCall API fallback, includes `last_alert` with its raw `payload` decoded into `map[string]any`.
> 
> The public API implementation switches from the typed `amixr` alert group service call to a direct `GET alert_groups/{id}/` request (with `url.PathEscape`) to capture fields not exposed by the client library, while the proxy path preserves existing behavior but wraps the result. A new `tools/oncall_unit_test.go` verifies the settings lookup + alert group request sequence and that nested `last_alert.payload` data is preserved.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e6fcdb1340e2b2a7a1a862622ad6063ab091822d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tools/oncall.go
+++ b/tools/oncall.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
 	"reflect"
 	"strings"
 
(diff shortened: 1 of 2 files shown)


sd2k (commented):
Thanks for this. I wonder if we should just always return the last alert payload? The extra included context on the off chance that this tool is called is probably worth it vs the need to add the extra `includeLastAlertPayload` param and description for _every_ conversation and user, especially since if the user is debugging an alert they probably want to see the payload anyway.

Genmin:
Updated for the design feedback and rebased onto current main. The tool no longer exposes an includeLastAlertPayload parameter; get_alert_group now includes the last alert payload by default when the public OnCall API returns it. The IRM proxy path remains compatible and returns the regular alert-group shape if that internal endpoint does not expose last_alert.\n\nLocal validation:\n\n- go test ./tools -tags unit -run 'TestGetAlertGroupIncludesLastAlertPayload|TestProxyGetAlertGroup'\n- go test ./tools -tags unit\n\nThe remaining blocker is the CLA status.

OshriBay:
@sd2k  any update ?

sd2k:
Yeah this looks good but we still need the CLA to be signed before it's mergeable.

sd2k:
Superseded by #1081, thanks for the contribution!

PR #579, opened 2026-02-15
Make server instructions reflect disabled categories

drtootsie:
Addresses #552

## Problem
The MCP server instructions were hardcoded and always listed all capabilities, even when categories were disabled. This caused AI clients to repeatedly attempt using disabled features.

## Solution
Made instructions dynamic - they now reflect only enabled categories.

## Changes
- Added `buildInstructions()` method to dynamically build instructions
- Added `isEnabled()` helper to check category status
- Prometheus & Loki line now adapts based on what's enabled
- Disabled categories are completely omitted

## Examples

**With `--disable-loki`:**
- Before: "Prometheus & Loki: Run PromQL and LogQL queries..."
- After: "Prometheus: Run PromQL queries..."

**With `--disable-loki --disable-oncall --disable-incident`:**
Only enabled categories appear in the instructions.

## Impact
AI clients now receive accurate capability information upfront, preventing wasted attempts to use disabled features.

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it refactors tool-category registration into a centralized map and changes how server `WithInstructions` is generated, which could inadvertently omit/incorrectly describe capabilities if the mapping drifts from flags.
> 
> **Overview**
> Server instructions are no longer hardcoded; they’re now built dynamically from the configured `enabled-tools` list and `--disable-*` flags via `disabledTools.buildInstructions()`, omitting disabled categories and adapting the Prometheus/Loki capability line based on what’s enabled.
> 
> Tool registration is refactored to use a single `toolCategories()` source-of-truth (including category metadata/instruction text) and `addTools()` now iterates that list instead of calling `maybeAddTools` per-category.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 9d1d0cdfb16fdac5e6587a22721364d5c8f33fa0. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #829, opened 2026-05-01
feat: build server instructions from enabled tool categories

### LINKED ISSUE #552: Server instructions should reflect disabled categories
## Problem

The MCP server `instructions` field (sent in the `initialize` response) is hardcoded and always lists all capabilities, including disabled ones. This causes AI clients like Claude Code to repeatedly attempt using disabled features.

For example, with this configuration:

```
--disable-loki --disable-oncall --disable-incident --disable-sift --disable-admin --disable-pyroscope
```

The instructions still announce:

> - Prometheus & Loki: Run PromQL and LogQL queries, retrieve metric/log metadata, and explore label names/values.

Even though the Loki tools are correctly not registered, the AI client reads the instructions and keeps trying to use Loki for log queries. The generic disclaimer *"Note that some of these capabilities may be disabled"* is not strong enough to prevent this behavior.

## Expected Behavior

The `instructions` text should dynamically reflect which categories are actually enabled. Disabled categories should be omitted from the capabilities list entirely.

## Suggested Implementation

The `newServer()` function in `cmd/mcp-grafana/main.go` already has access to the disabled tools configuration. The instructions string could be conditionally built based on which categories are enabled. For example:

- When Loki is disabled, the line should read "Prometheus: Run PromQL queries..." instead of "Prometheus & Loki: Run PromQL and LogQL queries..."
- Fully disabled categories (oncall, incident, sift, etc.) should be omitted from the list entirely

This would be a small, low-risk change that significantly improves the client experience.

## Current Workaround

Adding explicit instructions in the client configuration (e.g., CLAUDE.md) to override the MCP server's announcements. This works but is fragile and shouldn't be necessary.

## Version

Tested on v0.7.8, confirmed still present in v0.9.0.

blackwell-systems:
## Summary

- Server instructions now list only capabilities that are actually enabled, instead of hardcoding all categories regardless of `--disable-*` flags or `--enabled-tools` configuration
- Introduced `categoryDescription` map and `toolEntries()` as the single source of truth for category-to-description mapping, per the design `sd2k` suggested in the [review of PR #579](https://github.com/grafana/mcp-grafana/pull/579#pullrequestreview-2667155804)
- Removed the "Note that some of these capabilities may be disabled" disclaimer since the instructions now only list what's available

## Motivation

When categories are disabled (e.g. `--disable-loki` or `--enabled-tools=search,dashboard`), the instructions still advertised Loki, Prometheus, Elasticsearch, etc. Agents then attempted to call tools that don't exist, wasting tokens and producing errors. This particularly affects users running multiple MCP servers where tooling conflicts (see [#552 comment](https://github.com/grafana/mcp-grafana/issues/552#issuecomment-4330340000)).

## Design

The previous PR (#579) was closed because it introduced a separate `isEnabled` function with untyped string comparisons. This PR follows `sd2k`'s suggested architecture:

1. `toolEntries()` returns the ordered list of `{adder, disabled, category}` entries: the single place mapping flags to categories
2. `processTools()` iterates entries to register tools on the server
3. `buildInstructions()` iterates the same entries to build the instruction string

No new `isEnabled` function. No untyped string matching outside the `categoryDescription` map.

## Test plan

- [x] `TestBuildInstructions_ReflectsEnabledCategories` (4 subtests): all defaults, disabled flag exclusion, enabled-list exclusion, empty enabled list
- [x] `TestBuildInstructions_TimestampNote`: timestamp note always present regardless of categories
- [x] Existing `TestNewServer_SessionIdleTimeout*` tests still pass
- [x] `go build ./cmd/mcp-grafana/` passes
- [x] `go test ./cmd/mcp-grafana/` passes (4/4)

Fixes #552

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are limited to how server instructions are generated and how tool registration is iterated, with no changes to underlying tool implementations or auth/data paths.
> 
> **Overview**
> Server startup now **builds the MCP instruction text dynamically** from the actually enabled tool categories (respecting `--enabled-tools` and per-category `--disable-*` flags), instead of using a hardcoded capabilities list.
> 
> Tool registration is refactored to use a single ordered `toolEntries()` list shared by both `processTools()` (registration) and `buildInstructions()` (instructions), and new unit tests validate instruction content for enabled/disabled/empty configurations and that the timestamp note is always present.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 60d96e7e04c7eb7dd44477eed8f779d769dfa4c4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/cmd/mcp-grafana/main.go
+++ b/cmd/mcp-grafana/main.go
@@ -23,17 +23,47 @@ import (
 	"go.opentelemetry.io/otel/semconv/v1.40.0/mcpconv"
 )
 
-func maybeAddTools(s *server.MCPServer, tf func(*server.MCPServer), enabledTools []string, disable bool, category string) {
+// maybeAddTools registers a tool category and returns true if the category was enabled.
+func maybeAddTools(s *server.MCPServer, tf func(*server.MCPServer), enabledTools []string, disable bool, category string) bool {
 	if !slices.Contains(enabledTools, category) {
 		slog.Debug("Not enabling tools", "category", category)
-		return
+		return false
 	}
 	if disable {
 		slog.Info("Disabling tools", "category", category)
-		return
+		return false
 	}
 	slog.Debug("Enabling tools", "category", category)
 	tf(s)
+	return true
+}
+
+// categoryDescription maps a tool category to the description shown in server instructions.
+var categoryDescription = map[string]string{
+	"search":        "Search: Find dashboards, folders, and other Grafana resources.",
+	"datasource":    "Datasources: List and fetch details for datasources.",
+	"incident":      "Incidents: Search, create, update, and resolve incidents in Grafana Incident.",
+	"prometheus":    "Prometheus: Run PromQL queries, retrieve metric metadata, and explore label names/values.",
+	"loki":          "Loki: Run LogQL queries, retrieve log metadata, and explore label names/values.",
+	"elasticsearch": "Elasticsearch and OpenSearch: Query Elasticsearch and OpenSearch datasources using Lucene syntax or Query DSL for logs and metrics.",
+	"influxdb":      "InfluxDB: Query InfluxDB datasources.",
+	"alerting":      "Alerting: List and fetch alert rules and notification contact points.",
+	"dashboard":     "Dashboards: Search, retrieve, update, and create dashboards. Extract panel queries and datasource information.",
+	"folder":        "Folders: Manage dashboard folders.",
+	"oncall":        "OnCall: View and manage on-call schedules, shifts, teams, and users.",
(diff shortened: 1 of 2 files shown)


sd2k (dismissed):
Nice thanks. I am a bit unsure about how we're going to keep the category descriptions map up to date with the slice of tool entries. It'd be nice to have some kind of static assertion that each contains the same entries 🤔 

Still, this is definitely an improvement, I'll look at that other issue separately.

sd2k:
(ah, needs a `go fmt`)

sd2k (approved):

PR #835, opened 2026-05-05
tool: add new tool to install and search plugins

jcolladokuri:
## Summary
Extends the plugin management toolset with two new tools and improves discoverability guidance..

`search_plugin_information` — searches the Grafana.com plugin catalog by keyword (name, slug, description, tags). Results are sorted by trust tier: official Grafana Labs plugins first, then commercial plugins, then community. Attaches structured warnings for enterprise-only, private-signature, and Angular-deprecated plugins. Capped at 10 results with a note when truncated. This tool is intended to be called before get_plugin or install_plugin when the user describes a plugin by purpose rather than exact ID.

`install_plugin` — installs a plugin by ID via POST `/api/plugins/{id}/install`. If no version is provided, the tool resolves the latest version from the catalog and returns it for user confirmation before proceeding. Write-gated behind enableWrite alongside the existing tools.

`get_plugin` — updated tool description and Suggestion field on the not-found path to direct the model toward install_plugin.

## Future work

On the plugin management side of things, this should be the final PR. However, next steps will be to connect this tooling with datasource management tooling so that:

1. If a user has a generic `create a new azure datasource`, the client can use plugin discoverability help the user choose the specific plugin they are looking for.
2. If a user attempts to create a datasource for a plugin that has not yet been installed, they are able to go through the discover > install plugin flow before creating it.

## Screenshots

- (`search_plugin_information`) Generic prompt results in client listing relevant datasources in order of priority (1. grafana, 2. commercial, 3. community)
Screenshot 2026-05-05 at 3 38 19 PM

- After selecting the specific plugin, tools check whether it is installed, and if not, ask for user confirmation to install:

Screenshot 2026-05-05 at 3 40 42 PM

- After confirming, plugin is successfully installed, and user is prompted to configure a datasource as next steps (eventually this will be linked with create datasource tool so that the user know they can create a new datasource via chat)

Screenshot 2026-05-05 at 3 41 57 PM

- Attempting to install / connect to a non existing plugin, returns nothing

Screenshot 2026-05-05 at 3 42 33 PM

---

> [!NOTE]
> **Medium Risk**
> Introduces a new destructive `install_plugin` write tool that can modify a Grafana instance and adds network calls to the public Grafana.com catalog; failures/misuse could impact plugin state or rely on external availability.
> 
> **Overview**
> Adds plugin discoverability and installation capabilities to the MCP toolset. A new read-only `search_plugin_information` tool queries the Grafana.com plugin catalog, returns up to 10 matches sorted by signature trust/popularity, and surfaces warnings (enterprise-only, private, Angular).
> 
> Adds a write-gated `install_plugin` tool that POSTs to `/api/plugins/{id}/install`; if no version is provided it first resolves the latest catalog version and returns a confirmation-required response. `get_plugin` now validates empty IDs, shares a generic request helper for GET/POST, and returns a not-installed suggestion pointing users to `install_plugin`.
> 
> Server tool registration is updated so plugin tools respect the global write enable/disable flag, and tests are expanded with extensive unit coverage plus integration tests for catalog search and install confirmation flow.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6fc1ca740d8ce38fc5231cce4f7c6e779c5db18f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tools/plugins.go
```diff
@@ -104,13 +121,275 @@ func getPlugin(ctx context.Context, args GetPluginParams) (*GetPluginResult, err
...
+		return nil, fmt.Errorf("decode catalog: %w", err)
+	}
+
+	query := strings.ToLower(strings.TrimSpace(args.Query))
+	var matched []catalogPlugin
+	for _, p := range catalog.Items {
+		if pluginMatchesQuery(p, query) {
+			matched = append(matched, p)
+		}
+	}
```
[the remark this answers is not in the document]
jcolladokuri: I have not found evidence that `https://grafana.com/api/plugins` has pagination. From my testing, it seems like all items are returned in a single call response.

tools/plugins.go
```diff
@@ -104,13 +121,275 @@ func getPlugin(ctx context.Context, args GetPluginParams) (*GetPluginResult, err
...
+	return result.Version, nil
+}
+
+func installPlugin(ctx context.Context, args InstallPluginParams) (*InstallPluginResult, error) {
+	cfg := mcpgrafana.GrafanaConfigFromContext(ctx)
+	if cfg.URL == "" {
+		return nil, fmt.Errorf("grafana URL is not configured")
+	}
+
+	pluginID := strings.TrimSpace(args.PluginID)
```
yesoreyeram, line 191: if pluginID == "" {
    return nil, fmt.Errorf("pluginId is required")
}

can we add a validation for empty plugin Id?

tools/plugins.go
```diff
@@ -104,13 +121,275 @@ func getPlugin(ctx context.Context, args GetPluginParams) (*GetPluginResult, err
...
+
+	if args.Version == "" {
```
yesoreyeram: We need to default to `latest`. Later if required, we can add support for installing specific versions

tools/plugins.go
```diff
@@ -74,13 +91,13 @@ func getPlugin(ctx context.Context, args GetPluginParams) (*GetPluginResult, err
 	}
 
 	pluginID := strings.TrimSpace(args.PluginID)
```
yesoreyeram, line 93: let's add validation for empty plugin id

tools/plugins.go
```diff
@@ -104,13 +121,275 @@ func getPlugin(ctx context.Context, args GetPluginParams) (*GetPluginResult, err
...
+		Message:    "Plugin installed successfully. Grafana may need to be restarted for the plugin to become active.",
+		Suggestion: "Configure a new data source for the plugin.", // For now keeping this static to a single suggestion, down the line we may end up with a list
+	}, nil
+}
+
+var InstallPlugin = mcpgrafana.MustTool(
+	"install_plugin",
+	"Install a Grafana plugin by its plugin ID. If the version is not already confirmed with the user, omit it — the tool will look up the latest version and return it for confirmation before installing.",
+	installPlugin,
+	mcp.WithTitleAnnotation("Install plugin"),
```
yesoreyeram, line 233: can we add other annotations such as similar to other places we have

```
mcp.WithReadOnlyHintAnnotation(false)
mcp.WithIdempotentHintAnnotation(false)
mcp.WithDestructiveHintAnnotation(true) // if available
```

yesoreyeram (dismissed):

tools/plugins.go
jcolladokuri: We do default to `latest` if the user does not provide a specific version they want to install for a plugin in the prompt. 

Do you mean we should not support users being able to specify the version via chat at all?

yesoreyeram: > Do you mean we should not support users being able to specify the version via chat at all?

yes. We don't need as we are moving towards managed plugins allowing users to install the specific version will be problematic in long term. But if we are defaulting to `latest`, I am fine as well.

tools/plugins.go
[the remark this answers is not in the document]
yesoreyeram: yes. Unfortunately the API doesn't have pagination/search functionality yet.

tools/plugins.go
jcolladokuri: done :)

tools/plugins.go
jcolladokuri: Done :)

tools/plugins.go
jcolladokuri: Done!

tools/plugins.go
```diff
@@ -104,13 +125,281 @@ func getPlugin(ctx context.Context, args GetPluginParams) (*GetPluginResult, err
...
+		return result, nil
+	}
+
+	_, status, err := grafanaPluginRequest(ctx, cfg, http.MethodPost, "/api/plugins/"+url.PathEscape(pluginID)+"/install", map[string]string{"version": args.Version})
+	if err != nil {
+		return nil, fmt.Errorf("install plugin: %w", err)
+	}
+	if status != http.StatusOK {
+		return nil, fmt.Errorf("install plugin: unexpected status %d", status)
+	}
```
sd2k, line 220: I imagine a common case being a 403 when the user doesn't have permissions, perhaps it would be nice to include the response body here though, think it might be more actionable?

sd2k (dismissed):
LGTM. I added a comment on the install plugin tool but otherwise this seems good! Thanks.

tools/plugins.go
jcolladokuri: yes! good idea. I have included this as part of the result :)

jcolladokuri:
@cursor push d124a140e9

sd2k (approved):

PR #878, opened 2026-05-13
tool: single datasource provisioning support

jcolladokuri:
## Summary

This PR adds support for being able to generate a YAML file for a specific datasource type based on a structured schema.

### Process
This tool works by asking the client to be called twice:

1. The first time it is called, it should be called with the datasource type that the user wants to provision (i.e. prometheus), the tool will recognize that no additional params have been shared, and go and fetch the right schema for the datasource so tha the client knows what information to ask for. 

2. Once the client has gathered the right information from the user, another call is made into the tool with the right parameters from which we can now generate the YAML file.

### Output:

For now, the tool returns two things:
1. the generated YAML as text - mainly for clients that do not support ZIP file contents
2. the generated YAML file as a zip content - happy path for clients that do support ZIP file contents. Because we won't be able to write to disk for any user when they are running the server remotely, this is is a way to surface content that they can just ask their client of choice to extract into whatever directory they would like to use.

One thing to note, is that some clients like Claude will not gracefully handle types that they do not support and will simply result in an error where Claude will try to generate the YAML itself from whatever information it can gather

It seems the tool retumed an unsupported formit  Here&#39;s the standard provisioning YAMIL vou

## Future work
1. Right now provisioning works for a single datasource, next steps would be to support it for multiple, both within the same file and accross files.
2. Due to the complexity of the schemas, for now we are including them as static files, and they are pulled from this repository: https://github.com/grafana/dsconfig/tree/main/src. Eventually the idea is to bundle these files into the plugin and access the file via DNS similar to plugin JSON
3. Suport terraform provisioning for cloud customers

## Screenshots and screen recordings

https://github.com/user-attachments/assets/6f4541c4-a970-4265-a535-d40c2a203c99

HTTP method PUT vs POST

---

> [!NOTE]
> **Medium Risk**
> Adds a new tool that generates provisioning YAML from embedded datasource schemas and returns zip blobs; correctness and schema filtering/routing bugs could produce invalid configs or omit required fields. Also significantly increases embedded static assets, impacting binary size and maintenance.
> 
> **Overview**
> Adds a new `datasource-provisioning` tool category (enabled by default and toggleable via `--disable-datasource-provisioning`) and wires it into server capability instructions.
> 
> Introduces a new `provision_datasource` tool that uses an embedded per-datasource JSON schema and a **two-phase** call flow (first returns field guidance; second generates provisioning YAML and a base64 zip attachment), including logic to route inputs to root vs `jsonData`, ignore `secureJsonData`, and handle section-scoped keys.
> 
> Bundles a large set of datasource schema JSON files under `tools/datasource_schemas/` and adds comprehensive unit tests covering schema filtering, field routing, YAML output, and zip generation.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5f3d994948ae9be0e5401c9326f89f5da18bf041. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jcolladokuri:
@cursor push 0d1218f1c3

jcolladokuri:
@cursor push 168720ca71

jcolladokuri:
@cursor push c272e4e1d9

jcolladokuri:
@cursor push 06355df37c

PR #799, opened 2026-04-27
feat(tools): add Amazon Athena datasource support

imankurpatel000:
## Summary

Adds 5 MCP tools for querying Amazon Athena via Grafana's `/api/ds/query` and datasource resource API, following the ClickHouse pattern. Athena is disabled by default — users opt in with `--enabled-tools ...,athena`.

Closes #277.

### What's included

- **4 discovery tools** — `list_athena_catalogs`, `list_athena_databases`, `list_athena_tables`, `describe_athena_table` — call the Athena plugin's resource endpoints to let LLMs explore schema before querying.
- **`query_athena`** — executes SQL via `/api/ds/query` with macro substitution (`$__timeFilter`, `$__dateFilter`, `$__unixEpochFilter`, `$__timeFrom`, `$__timeTo`, `$__from`, `$__to`, `$__interval`), template variable substitution, limit enforcement (default 100, max 1000), and result reuse support (Athena engine v3).
- **SHOW/DESCRIBE passthrough** — `enforceAthenaLimit` skips LIMIT enforcement for `SHOW` and `DESCRIBE` statements, since Athena SQL does not accept LIMIT on these (unlike `SELECT`). This allows LLMs to run `DESCRIBE db.table` for column types and Iceberg partition structure.
- **Empty-result hints** — Athena-specific suggestions when a query returns no rows, including guidance to use `DESCRIBE` for partition-aware debugging.
- **Docs** — README updated with Athena section, 5 tool rows in the reference table, `--disable-athena` in the flags reference, `athena` added to the Makefile's `ENABLED_TOOLS`.

### Scope

- **In:** Schema discovery (catalogs, databases, tables, columns), SQL query with macros/variables/limits, result reuse, empty-result hints, per-query overrides for region/catalog/database, SHOW/DESCRIBE passthrough.
- **Out:** `list_athena_workgroups` — workgroup is not overridable per-query in the Athena plugin (comes from datasource config only), so listing them would be informational only. Can be added later if there's demand.
- **Out:** `$__timeGroup` macros — no datasource in this repo implements these; the plugin handles them server-side.

### Unit tests (22 test cases)

- [x] `go build ./...` — ok
- [x] `go test -tags unit ./...` — all pass (adds 22 new unit test cases covering macro substitution, limit enforcement, SHOW/DESCRIBE bypass, HTTP client methods, and hints)
- [x] jsonschema comma linter — clean
- [x] gofmt — clean

[...]

PR #884, opened 2026-05-15
fix: add readResponseBody helper to limit and detect oversized responses

imankurpatel000:
## Summary

Introduces a `readResponseBody(r, limit)` helper that reads `limit+1` bytes and returns a hard error on overflow, replacing silent truncation across all HTTP clients. This ensures callers get a clear "response body exceeds maximum size" error instead of confusing downstream JSON/YAML parse failures from truncated data.

- Adds limits to **6 previously unbounded** success-path reads (alerting, asserts, oncall, plugins, run_panel_query)
- Consolidates **8 duplicate 10MB constants** into a single `defaultResponseLimitBytes`
- Removes `unmarshalJSONWithLimitMsg` (now redundant — overflow is caught before unmarshal)
- Preserves intentionally different limits: cloudwatch resource-listing (1MB), opensearch (48MB)
- Renames `json_utils.go` → `response_utils.go` (the helper handles YAML, NDJSON, DOT graphs, not just JSON)

Follow-up from @sd2k's [request in #799](https://github.com/grafana/mcp-grafana/pull/799#discussion_r3207744076).

### Note

This ended up as a larger refactor than I initially anticipated, what started as adding `io.LimitReader` to a few call sites grew into consolidating duplicated constants and replacing silent truncation with proper overflow detection. Let me know if this isn't what you had in mind and I can split it up.

### Out of scope

- `rendering.go` (image binary) — left unbounded pending discussion on appropriate image limit
- `alerting_client.go` `json.NewDecoder` path — self-limiting (stops after one JSON value)

## Test plan

- [x] `go build ./...` passes
- [x] `make test-unit` passes (all existing + new `TestReadResponseBody` covering: within limit, at limit, exceeds by 1 byte, reader error, empty body)
- [x] `gofmt -l .` clean
- [x] `make lint` passes (jsonschema + openapi linters; pre-existing golangci-lint `reflect.Ptr` warnings unrelated)
- [x] `grep -rn "io.ReadAll(resp.Body)" tools/ | grep -v _test.go | grep -v rendering.go` returns nothing
- [x] `grep -rn "unmarshalJSONWithLimitMsg" tools/` returns nothing

---

> [!NOTE]
> <sup>[Cursor Bugbot](https://cursor.com/bugbot) is generating a summary for commit 3aafb2be4c6e72bfe0a6dd8a4d3b8632117e247a. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tools/oncall_proxy.go
sd2k, line 1: Claude had a bug report about the two functions here:

  oncall_proxy.go: readResponseBody error swallowed on non-200 responses

  In both fetchSinglePage and fetchPaginated, the err from readResponseBody is checked after the status code check:

```
  body, err := readResponseBody(resp.Body, defaultResponseLimitBytes)
  _ = resp.Body.Close()
  if resp.StatusCode != http.StatusOK {
      if len(body) > 0 {           // body is nil when readResponseBody errors
          return nil, fmt.Errorf(...)
      }
      return nil, fmt.Errorf("request failed with status %d", resp.StatusCode) // ← returned here
  }
  if err != nil {                   // ← never reached
      return nil, fmt.Errorf(...)
  }
```

  If readResponseBody returns an error (oversized body or I/O failure) and the status is non-200, body is nil, so the code falls into the len(body) == 0 branch and returns a generic
  status-code error. The actual read error is silently discarded. This is a pre-existing pattern (the err-after-status ordering was there before), but the behavioral change is that
  previously io.ReadAll would return partial data with the error, so len(body) > 0 could still include the error body. Now readResponseBody returns nil on overflow, so non-200 oversized
  responses always lose their body content. Low severity in practice (10MB error bodies are unlikely from OnCall), but worth noting.

sd2k (commented):
Thanks a lot for this, it looks great. One very small issue with the Oncall proxy but otherwise this looks OK to merge!

tools/oncall_proxy.go
imankurpatel000: Yeah, this came up with my own review with claude as well but I thought that 500 status would be more informational than response body exceeds error. But I fixed it to include both. so error will be 
```
request failed with status 500 (response body exceeds maximum size of 10485760 bytes; try narrowing your query)
```

sd2k (approved):
LGTM, thanks!

PR #887, opened 2026-05-19
feat(proxied): support full proxy of resources, templates, and prompts for tempo

HackToHell:
## Summary
  This PR implements full proxy capabilities for upstream MCP servers discovered dynamically via Grafana datasources. It expands the existing "proxied tools only" behavior into a
  complete proxy layer that re-exposes resources, templates, and prompts.
  ## Key Changes
  - **Resources & Templates Namespacing**:
    - Rewrites upstream static resource URIs into unique, namespaced URNs (`urn:mcp-grafana:<datasourceType>:<datasourceUid>:<encoded-original-uri>`).
    - Correctly maintains raw URI templates (per RFC 6570 compliance) to allow clients to perform substitution before sending back URN read requests.
  - **Prompts Proxying**:
    - Rewrites prompt names to include the datasource type as a prefix and injects a required `datasourceUid` argument to dynamically route prompts/get calls.
  - **Improved Lifecycle Management**:
    - Prevents redundant registration passes and log spam in sessions where no upstream MCP servers are configured by tracking registration status in the session state.
  - **Testing & Quality**:
    - Full suite of namespacing roundtrip and edge-case unit tests.
    - Integration and manual end-to-end client verification validating full tools, resources, and templates discovery and routing against active dual-Tempo containers.

---

> [!NOTE]
> **Medium Risk**
> Expands the proxied MCP surface area beyond tools into resources/templates/prompts and changes session/initialize-time capability advertisement, which could affect client discovery flows and multi-tenant session scoping if incorrect.
> 
> **Overview**
> Enables **full proxying of upstream MCP servers discovered via Grafana datasources**, expanding beyond tools to also forward *resources*, *resource templates*, and *prompts* (with Tempo TraceQL docs exposed under namespaced URNs).
> 
> Adds namespacing + routing for proxied resources (`urn:mcp-grafana:<type>:<uid>:<escaped-uri>`), resource templates (verbatim template embedding), and prompts (namespaced names plus injected required `datasourceUid`), along with new handlers to forward `resources/read` and `prompts/get` to the correct upstream client.
> 
> Updates server startup/session hooks to lazily register proxied capabilities on any relevant MCP entrypoint (tools/resources/prompts) and to advertise resource/prompt capabilities at `initialize` time; introduces session state to prevent redundant registration, hardens upstream capability probing/error handling, and adds unit tests plus new deps (`uritemplate`, promotes `gojq` to direct).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd69e7a78795c3b91f1d968455157bffcb592aed. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


proxied_tools.go
```diff
@@ -208,6 +210,122 @@ func parseProxiedToolName(toolName string) (string, string, error) {
...
+	parts := strings.SplitN(rest, ":", 3)
+	if len(parts) != 3 {
+		return "", "", "", fmt.Errorf("invalid proxied resource URI format: %s", uri)
+	}
+	original, err := url.QueryUnescape(parts[2])
+	if err != nil {
+		return "", "", "", fmt.Errorf("failed to decode original URI in %s: %w", uri, err)
+	}
+	return parts[0], parts[1], original, nil
+}
```
sd2k: `url.QueryUnescape` will corrupt + characters in resource template URIs

`namespaceResourceURI` uses `url.QueryEscape`, and `parseNamespacedResourceURI` uses `url.QueryUnescape` — this roundtrips correctly for static resources. But for resource templates, `namespaceResourceTemplate` embeds the upstream URI verbatim (no encoding). When a client expands the template and sends the result to resources/read, `parseNamespacedResourceURI` still runs `url.QueryUnescape` on the original-URI portion, which
  decodes + → space.

Example: upstream template `docs://a+b/{section}` → client expands to `urn:mcp-grafana:tempo:abc-123:docs://a+b/foo` → `QueryUnescape("docs://a+b/foo")` → `docs://a b/foo` (broken).

Fix: Use url.PathEscape/url.PathUnescape instead of url.QueryEscape/url.QueryUnescape. PathUnescape treats + as literal.

sd2k (commented):
Nice thanks for this, I didn't even know the Tempo MCP server had resources or templates!

Claude noticed this one inline finding, could you write a test for it and fix if it's correct? Cheers.

proxied_tools.go
HackToHell: thanks, fixed this.

cmd/mcp-grafana/main.go
```diff
@@ -309,10 +309,23 @@ func newServer(transport string, dt disabledTools, obs *observability.Observabil
 	// then register tools.
 	instructions := dt.buildInstructions()
 
-	s = server.NewMCPServer("mcp-grafana", mcpgrafana.Version(),
+	serverOpts := []server.ServerOption{
```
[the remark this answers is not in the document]
HackToHell: Fixed this as well, I think only tools had a default hook.

sd2k:
LGTM now, sorry about the go.sum conflicts :disappointed: will get this in as soon as that's resolved.

sd2k:
Thanks for your patience on this one, @HackToHell — `main` moved a long way underneath the branch, so I merged it and pushed the resolution (`2f56b0d`). CI is green: 18 pass, 0 fail (the 8 skips are cloud-credential-gated jobs that fork PRs don't get).

Most of the merge was mechanical, but I made one behavioural change around prompts that I want to flag explicitly, since it narrows what the PR advertises.

### What I changed

Your call to skip per-session prompts was right, and I kept it. What I removed is the capability declaration that went alongside it:

```diff
 	if !dt.proxied {
-		serverOpts = append(serverOpts,
-			server.WithResourceCapabilities(false, true),
-			server.WithPromptCapabilities(true),
-		)
+		serverOpts = append(serverOpts, server.WithResourceCapabilities(false, true))
 	}
```

The two pieces were inconsistent with each other. `InitializeAndRegisterProxiedCapabilities` never registers a prompt on HTTP/SSE, but `WithPromptCapabilities(true)` applied to every transport — so on SSE and streamable-http the server told clients at `initialize` that it had prompts, and then answered `prompts/list` with an empty list forever. That's a wasted round trip for every client that believes the handshake, and a confusing surface for anyone debugging why the Tempo prompts they were promised never appear.

Dropping it doesn't cost anything on stdio: `AddPrompts` calls `implicitlyRegisterPromptCapabilities` internally, and the stdio path registers before `srv.Listen`, so the capability is still advertised at `initialize` — but only when a prompt actually exists.

Net effect: prompts are proxied on **stdio only**, and only stdio advertises them.

### Why the session-side constraint still holds

Worth confirming since the merge also took mcp-go 0.46 → 0.55: your original reasoning is still accurate. 0.55 added session-scoped resources and templates, and both HTTP transports assert them —

```go
_ SessionWithTools             = (*streamableHttpSession)(nil)
_ SessionWithResources         = (*streamableHttpSession)(nil)
_ SessionWithResourceTemplates = (*streamableHttpSession)(nil)
```

— but there is still no `AddSessionPrompts` and no `SessionWithPrompts`. So a proxied prompt on an HTTP transport could only be registered server-wide, where one caller's prompts would be visible to every other session. That's a tenant-isolation problem, not a cosmetic one, so skipping remains the correct behaviour rather than something to work around.

### Making the gap visible

The one thing I didn't want was for this to be a silent drop, so it's now observable and documented:

- `mcp_proxied_capabilities_skipped_total{mcp_proxied_capability="prompts",mcp_proxied_skip_reason="no_session_prompt_support"}` counts every dropped prompt.
- The warning log is routed through the context-aware logger (matching how the rest of the proxied subsystem logs post-#1001) instead of the bare `tm.logger`.
- README, CHANGELOG, and `docs/sources/configure/proxied-tools.md` all state the stdio-only limitation rather than implying prompts work everywhere.

Tempo exposes no prompts today, so no current deployment is affected either way — this only matters for the next upstream MCP server that does. If `AddSessionPrompts` ever lands upstream, the prompts are already built and namespaced on the shared set (`proxiedToolSet.prompts`), so enabling them should be a registration call plus putting that capability declaration back.

Happy to revisit if you disagree with the reasoning on the capability flag.

sd2k:
I think we are likely to get rid of proxied tools/etc soon because it's a high operational burden and blocks us going completely stateless. #1149 tracks that. We can add resources/templates/prompts once that's in!

PR #889, opened 2026-05-21
feat(dashboard): add get_dashboard_intent + list_dashboard_intent tools

tonypowa:
## Summary

Surface author-declared dashboard intent (purpose, owner, expected behavior, failure modes, related SLOs, runbooks, per-field provenance) to MCP clients.

The Grafana Assistant **Dashboard Intent** feature stores intent inside the dashboard JSON — a top-level \`intent\` block and per-panel \`panels[].intent\` blocks. These tools project that on-disk camelCase shape into a normalized, snake_case shape so external agents (Cursor, incident bots, status pages, CI pipelines) don't have to re-walk the dashboard JSON to read it.

- \`get_dashboard_intent(uid, panel_id?)\` — returns the intent block for a dashboard (\`panel_id\` omitted) or one of its panels. Errors when the addressed block is absent so callers can treat that the same as \"no authored intent yet\".
- \`list_dashboard_intent(uid)\` — returns the full bundle in one call: the dashboard-level block (when present) plus every panel-level block discovered while walking the dashboard's panels (including row-nested panels). Returns an empty bundle for dashboards with no authored intent.

Both tools are **read-only** and built on the existing \`GetDashboardByUID\` path, so they work against any Grafana the MCP client is authenticated to — they do **not** require the Grafana Assistant plugin to be installed on the target Grafana.

## Why now

The Dashboard Intent feature recently switched from an assistant-side DB to dashboard-JSON-native storage (canonical source of truth lives in the dashboard JSON itself). With the JSON shape stable, external MCP clients are the next natural surface to plug in.

## Test plan

- [x] \`go test ./tools/ -run TestProjectIntent -count=1\` — projection helper (full block, schema-version default, panel addressing, malformed-entry skip)
- [x] \`go test ./tools/ -run TestListDashboardIntent -count=1\` — bundle walk (dashboard + row-nested panels)
- [x] \`go test ./tools/ -short -count=1\` — full short-test suite still green
- [ ] Manual: connect the MCP server to a Grafana running the Assistant app with intent authored on at least one dashboard, and call \`get_dashboard_intent\` / \`list_dashboard_intent\` from an MCP client (Cursor / Claude Desktop).

## Related

- Grafana Assistant App: dashboard-intent JSON-native storage (PR [#6793](https://github.com/grafana/grafana-assistant-app/pull/6793))
- grafana/grafana: UI surfaces (PR [#125200](https://github.com/grafana/grafana/pull/125200))

Made with [Cursor](https://cursor.com)

### DIFF
--- a/tools/dashboard.go
+++ b/tools/dashboard.go
@@ -797,4 +797,5 @@ func AddDashboardTools(mcp *server.MCPServer, enableWriteTools bool) {
 	GetDashboardPanelQueries.Register(mcp)
 	GetDashboardProperty.Register(mcp)
 	GetDashboardSummary.Register(mcp)
+	AddDashboardIntentTools(mcp)
 }
--- a/tools/dashboard_intent.go
+++ b/tools/dashboard_intent.go
@@ -1,0 +1,289 @@
+package tools
+
+import (
+	"context"
+	"fmt"
+	"strings"
+	"time"
+
+	"github.com/mark3labs/mcp-go/mcp"
+	"github.com/mark3labs/mcp-go/server"
+
+	mcpgrafana "github.com/grafana/mcp-grafana"
+)
+
+// Dashboard intent (https://grafana.com — Grafana Assistant Dashboard
+// Intent feature) lives inside the dashboard JSON itself as a top-level
+// `intent` block and as a per-panel `panels[].intent` block. These
+// tools project that block into a normalized, snake_case shape so MCP
+// clients (Cursor, incident bots, CI pipelines, etc.) don't have to
+// re-walk the dashboard JSON to read it.
+//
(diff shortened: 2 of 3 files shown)


sd2k:
Hey Tony! I can see this is in draft now but my gut feeling is that these fields could just be added to the main 'get/list dashboard' tools rather than adding new tools (since those tools add to the token count for all users, which means $$$ spent on token cost). WDYT about that?

sd2k:
Closing for now as part of a cleanup, please re-open if you still think this is necessary after addressing [my comment above](https://github.com/grafana/mcp-grafana/pull/889#issuecomment-5341090438). Cheers!

Issue #721, opened 2026-04-06
Loki query_loki_logs returns 404: path doubled via /resources endpoint


joladev:
## Summary

`query_loki_logs` (and other Loki query tools) return 404 when used against a self-hosted Grafana instance with a self-hosted Loki datasource. The Loki API path is doubled: `/loki/api/v1/loki/api/v1/query_range`.

## Environment

- mcp-grafana: v0.11.4 (latest)
- Grafana: 12.4.2 (self-hosted, Docker)
- Loki: 3.7.0 (self-hosted, Docker)
- Datasource type: `loki`, URL: `http://loki:3100`

## Root cause

The MCP Loki client constructs endpoints with the full Loki API prefix (from `tools/loki.go`):

```go
endpoint = "/loki/api/v1/query_range"
```

This is sent to Grafana via the `/resources` endpoint:

```
GET /api/datasources/uid/<uid>/resources/loki/api/v1/query_range
```

Grafana's built-in Loki plugin `CallResource` handler (in `pkg/tsdb/loki/loki.go`) **always prepends** `/loki/api/v1/` to the resource path:

```go
lokiURL := fmt.Sprintf("/loki/api/v1/%s", url)
```

This results in the doubled path `/loki/api/v1/loki/api/v1/query_range` being sent to Loki, which returns 404.

## Why `/proxy` doesn't save it

The `datasourceFallbackTransport` tries `/proxy` first, which would work (it proxies directly without path manipulation). However, in our case a `/labels` call returned 500 during startup, which cached the fallback to `/resources` for all subsequent requests. Once cached, all Loki queries hit the doubled-path `/resources` endpoint.

Even without this caching issue, `/proxy` returned 400 for `query_range` requests in our testing, suggesting query parameters may not be forwarded correctly through that path.

## Evidence from Grafana server logs

```
logger=tsdb.loki endpoint=callResource uname=sa-1-grafana-mcp level=warn
  msg="Non 200 HTTP status received from loki"
  error="404 page not found\n"
  resourcePath="/loki/api/v1/loki/api/v1/query_range?direction=backward&end=...&query=..."

logger=context uname=sa-1-grafana-mcp
  msg="Request Completed" method=GET
  path=/api/datasources/uid/P8E80F9AEF21F6940/resources/loki/api/v1/query_range
  status=404
```

Note: Grafana dashboard queries to Loki work fine -- they use the `queryData` endpoint, not `callResource`.

## Suggested fix

[...]

PR #897, opened 2026-05-26
fix: scope datasource fallback cache by request path

ba-work:
## Summary
- scope datasource fallback cache entries by request path instead of only the datasource base path
- keep cached fallback behavior for repeated requests to the same endpoint
- add a regression test covering Loki patterns when another Loki endpoint falls back to `/resources`

## Why
Some Grafana datasource endpoints are not interchangeable between `/proxy` and `/resources`. A fallback learned for one endpoint can currently be reused for a different Loki endpoint, causing `query_loki_patterns` to be routed to `/api/datasources/uid/<uid>/resources/loki/api/v1/patterns`, which can return 404 even when `/api/datasources/proxy/uid/<uid>/loki/api/v1/patterns` works.

Minimal repro on a proxied Loki datasource:
- `list_loki_label_names({ datasourceUid: "loki" })`
- `query_loki_patterns({ datasourceUid: "loki", logql: "{}" })`

Related to the same `/resources` path compatibility class as #721.

## Testing
- `GOPROXY=direct go mod download`
- `go test ./tools -run 'TestDatasourceFallbackTransport' -count=1`

---

> [!NOTE]
> **Medium Risk**
> Changes HTTP routing for all datasource proxy calls; low blast radius but can alter which Grafana path is used per endpoint.
> 
> **Overview**
> **Scopes datasource proxy fallback caching per request path** instead of per datasource base URL, so a successful fallback on one Grafana/Loki API route is not reused for unrelated routes where `/proxy` and `/resources` behave differently.
> 
> `datasourceFallbackTransport` now builds a per-path `fallbackCacheKey`, still caches only after a **2xx** fallback, and sets **`ContentLength`** when replaying buffered POST bodies. Tests cover per-path cache hits, cross-endpoint isolation, and a **Loki patterns** regression where `/resources` 404s but `/proxy` works.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bed6574fa21163df19a87f9d9273cf99447dbe95. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/tools/fallback_transport.go
+++ b/tools/fallback_transport.go
@@ -23,7 +23,8 @@ type datasourceFallbackTransport struct {
 }
 
 // fallbackEndpoints caches which datasource proxy paths need the fallback
-// endpoint. Key is the primary base path, value is true if fallback is needed.
+// endpoint. Key is the primary base path plus request path suffix, value is
+// true if fallback is needed.
 var fallbackEndpoints sync.Map
 
 func newDatasourceFallbackTransport(wrapped http.RoundTripper, primaryBase, fallbackBase string) http.RoundTripper {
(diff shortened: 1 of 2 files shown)


sd2k (approved):
LGTM, thank you!

PR #898, opened 2026-05-26
tools: add support for multi datasource creation from mcp

jcolladokuri:
## Summary

This PR modifies `create_datasource` so that it is able to create datasources in bulk for a single datasource type.

The tool still works the same way as before where it is 2-phased:

1. Call the tool with just the type to get the schema and infer what fields are required and need user input
2. Call the tool again with all the filled out fields properly 

Main difference is now it can handle a prompt like:

```
Help me create some prometheus datasources on my grafana instance. I have 10 prometheus regions, I need to create a datasource per region where each datasource has basic auth enabled pointing to URL: https://localhost:9090/<region>. The name for each datasource should be prom-<region> and the available regions are: 1,2,3,4,5,6,7,8,9,10.
```

## Screenshots & screen recordings
https://github.com/user-attachments/assets/d8c86101-073e-48ce-89d7-f9bc6704fd77

---

> [!NOTE]
> **Medium Risk**
> This is a breaking change to the MCP tool input/output contract for datasource provisioning, with concurrent Grafana API writes and partial-failure semantics; coverage is strong in unit/integration tests.
> 
> **Overview**
> **`create_datasource`** now accepts **`type`** plus an optional **`datasources`** array of per-instance specs (`CreateDatasourceSpec`) instead of a single flat payload. The two-phase flow is unchanged in spirit: call with **type only** (empty/missing array) for schema guidance; call again with one or more specs to create them.
> 
> Creation runs **concurrently** (cap of 5) through shared **`executeDatasourceCreate`**, and the tool response is always a **`BulkCreateDatasourceResult`** with per-item **`results`**, **`succeeded`/`failed`/`total`**, and **`IsError`** when any item fails. Per-datasource outcomes add **`error`**, **`datasourceConfigPage`**, and **`nextSteps`** as a string list; the separate MCP **resource link** for the config page is removed.
> 
> Phase-2 validation blocks the whole batch if any spec is invalid (e.g. schema type with missing **`fields`**, or no-schema type with missing **`name`**) and returns guidance instead of partial creates.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d800cf6ba408102d114e970bb771e0247d47ce8b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


jcolladokuri:
@cursor push d434ad5ba4

sd2k (commented):
Hm, I'm not sure whether this is really worth it 🤔 Most MCP hosts have the ability to make multiple tool calls in parallel so they should be able to issue `create_datasource` multiple times if requested by the user, and this increases the complexity of the tools (more tokens consumed by the tool parameters and descriptions, and slightly more work for LLMs to get the parameters correct).

Have you noticed a real world issue with the single 'create datasource' tool being used multiple times?

jcolladokuri:
@sd2k yeah you are right, good call out, I was not aware that MCP hosts could do that without explicitly defining it in the tool. I tested the existing `create_datasource` and it was able to create multiple datasources. Will close this PR for now! thanks

PR #927, opened 2026-06-04
feat(prometheus): support optional time range for list_prometheus_metric_names

madaraszg-tulip:
Add optional startRfc3339/endRfc3339 parameters to list_prometheus_metric_names, mirroring the existing label name/value tools. Previously the __name__ label values were always fetched with unbounded times; the new params let callers restrict results to metrics active within a window. Empty values preserve the prior unbounded behavior.

---

> [!NOTE]
> **Low Risk**
> Read-only MCP tool extension with backward-compatible defaults; only changes how optional time bounds are forwarded to Prometheus label value APIs.
> 
> **Overview**
> **`list_prometheus_metric_names`** now accepts optional **`startRfc3339`** and **`endRfc3339`**, aligned with the existing label name/value discovery tools.
> 
> When set, those bounds are parsed as RFC3339 and passed into **`LabelValues`** for **`__name__`**, so callers can limit discovery to metrics that had series in that window instead of always using an unbounded range. Omitting either parameter keeps the previous zero-time (unbounded) behavior. The tool description and changelog are updated accordingly, and an integration test covers listing with a one-hour window.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 0c195b50df5d9661468b685a5a99c58bfaf8df12. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Added
+
+- Optional `startRfc3339`/`endRfc3339` time range parameters for `list_prometheus_metric_names` to restrict results to metrics active within a window
+
 ## [0.15.2] - 2026-06-04
 
 ### Fixed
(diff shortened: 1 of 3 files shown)


sd2k (dismissed):
Good idea, thanks!

sd2k (approved):

Issue #682, opened 2026-03-25
Memory leak in streamable-http mode causes OOMKilled pod restarts every ~1-2 days


tehranian:
## Description

When running mcp-grafana in `streamable-http` mode as a long-lived Kubernetes deployment, the process exhibits a slow memory leak (~4-5 MiB/hour) that eventually causes OOMKilled pod restarts. In our deployment with a 256Mi memory limit, the pod restarts every ~1-2 days.

## Environment

- **mcp-grafana version**: 0.11.3
- **Transport**: `streamable-http`
- **Flags**: `-t streamable-http --metrics --enabled-tools alerting,annotations,clickhouse,cloudwatch,dashboard,datasource,examples,folder,loki,navigation,prometheus,proxied,pyroscope,rendering,runpanelquery,search,searchlogs,sift`
- **Runtime**: Kubernetes (single replica), memory limit 256Mi
- **Grafana**: Grafana Cloud (twilio.grafana.net)

## Observed behavior

```
$ kubectl get pods -n claude-mcp-grafana -o jsonpath='{range .items[*]}{.metadata.name}{"\n"}{range .status.containerStatuses[*]}  restarts={.restartCount} lastState={.lastState}{"\n"}{end}{end}'

claude-mcp-grafana-grafana-mcp-79c88d46b4-rcfhv
  restarts=2 lastState={"terminated":{"exitCode":137,"reason":"OOMKilled","startedAt":"2026-03-23T21:38:04Z","finishedAt":"2026-03-25T19:13:01Z"}}
```

The pod starts at ~50 MiB after boot, grows steadily at ~4-5 MiB/hour, and is OOMKilled after ~45 hours when it exceeds the 256Mi limit. This cycle repeats indefinitely.

```
$ kubectl top pod -n claude-mcp-grafana
NAME                                              CPU(cores)   MEMORY(bytes)
claude-mcp-grafana-grafana-mcp-79c88d46b4-rcfhv   2m           50Mi     # 35 min after restart
```

## Root cause analysis

I've identified two sources of memory growth in the code:

### 1. New HTTP client + transport created per request (major)

`ComposedHTTPContextFunc` is registered as the `HTTPContextFunc` for streamable-http and is called on **every incoming HTTP request**:

https://github.com/grafana/mcp-grafana/blob/main/mcpgrafana.go#L934-L943

```go
func ComposedHTTPContextFunc(config GrafanaConfig) server.HTTPContextFunc {
[...]

PR #937, opened 2026-06-11
feat(dashboard): support native dashboard schema v2

madaraszg-tulip:
## Summary

Every dashboard tool currently fetches via the legacy `/api/dashboards/uid/{uid}` endpoint, which **lossily down-converts** dashboards stored in schema v2 (`elements`/`layout`/`variables`) back to v1 (`panels[]`/`templating.list`). Agents never see the real v2 dashboard.

This PR makes the dashboard tool group read and write the **native** schema using the `dashboard.grafana.app` Kubernetes-style API, transparently falling back to the legacy API on older Grafana.

## How it works

- **Capability gate (discovered once per connection, cached):** check whether `dashboard.grafana.app` serves `v1beta1` — the marker for Grafana ≥ 12.0.0, where schema v2 can exist. If yes → k8s path; otherwise → legacy API (lossless there, since no v2 dashboards exist on older Grafana).
- **Native fetch:** GET at `v1beta1`; if `status.conversion.storedVersion` is a `v2*` version (it is populated even when the lossy conversion did not error), re-fetch at that version to return the native object. `get_dashboard_by_uid` now also returns `apiVersion` and `isV2`.
- **v2-aware helpers:** `get_dashboard_summary` and `get_dashboard_panel_queries` parse the v2 spec (`elements`, `data.spec.queries[].spec.query` with `group`/`datasource.name`, `variables`, `timeSettings`); `get_dashboard_property` runs JSONPath over the native spec. Classic v1 handling is unchanged.
- **Writes:** `update_dashboard` writes v2 dashboards back through the k8s API (patch and full-JSON modes), preserving the native schema; v1 continues to use the legacy save endpoint.
- **Namespace** resolved from `/api/frontend/settings` (Grafana Cloud `stacks-*` / on-prem `org-N`/`default`), with an OrgID-derived fallback.
- `KubernetesClient` gains `Create`/`Update` and is cached in `ClientCache` (per #682) to avoid per-request transport allocation.

## Relation to #692 / #696

Those explored routing to the k8s API, but request the group's preferred version and convert the payload back to legacy types, so a v2 dashboard still comes back as lossy v1. This PR's focus is **native v2**: stored-version detection + re-fetch + native passthrough + v2-aware parsing. It deliberately avoids a discovery registry / TTL cache in favor of a single cached capability check.

## Test plan

- **Unit:** v2 spec parsing (elements/queries/variables/summary), stored-version detection & re-fetch (mock k8s server), capability gating, `Create`/`Update`, namespace resolution.
- **Integration** (`-tags integration`, two instances in docker-compose):
  - `grafana` on 13.0.2 with `dashboardNewLayouts` — `TestDashboardV2` creates → gets (native v2) → summary → panel-queries → patch; the existing v1 suite still passes.
  - `grafana-legacy` on Grafana 11 (no `v1beta1`) — `TestDashboardLegacyFallback` confirms the tools fall back to the legacy API.
- **Live:** verified against a real native v2 dashboard on a Grafana 13 staging instance — `isV2:true`, `apiVersion:v2`, native `elements`/`variables` returned (not lossy v1); summary/panel-queries/property all parse correctly. (Write path on staging was not exercised — no RW token there; covered by integration tests.)

---

> [!NOTE]
> **High Risk**
> Changes core dashboard read/write paths and optimistic-concurrency k8s updates; wrong namespace or capability fallback could corrupt or hide v2 dashboards, though fail-closed discovery and v1-over-v2 rejection mitigate that.
> 
> **Overview**
> Dashboard MCP tools stop relying solely on the legacy REST API, which **lossily down-converts** v2 dashboards to classic `panels[]` JSON. They now read and write **native** schema when Grafana serves `dashboard.grafana.app/v1beta1`, and fall back to legacy only when that version is definitively absent.
> 
> **Routing and safety:** `fetchDashboard` gates on cached `GroupVersions` discovery; transient discovery errors **fail closed** (no legacy fallback) so v2-capable instances are not read or patched through lossy v1. v2-stored dashboards are re-fetched at `status.conversion.storedVersion` after an initial v1beta1 GET. `get_dashboard_by_uid` adds **`apiVersion`** and **`isV2`**.
> 
> **Writes:** v2 dashboards go through k8s **Create/Update** (patches and full JSON with `elements`/`layout`); v1 still uses the legacy save endpoint. Guards reject in-place v2-over-v1 replacement and honor `overwrite=false`.
> 
> **Infrastructure:** `KubernetesClient` gains Create/Update, per-group version caching, and **`ClientCache`** integration; **`DashboardNamespace`** comes from `/api/frontend/settings` with org-derived fallback. v2-aware parsing for summary, panel queries, and property tools lives in **`dashboard_v2.go`**. Docker-compose adds **Grafana 11 on :3002** for legacy-path integration tests; main Grafana bumps to **13.0.2**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bd2cc3abbdd9458c1ad31d743aa5f45e6a67c99b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tools/dashboard.go
```diff
@@ -20,19 +20,191 @@ import (
...
+// re-fetches at the stored version when the dashboard is actually stored as v2.
+// Callers only reach this after confirming v1beta1 is served, so a 404 here is a
+// genuine "dashboard not found".
+func fetchDashboardViaK8s(ctx context.Context, k8s *mcpgrafana.KubernetesClient, uid string) (*dashboardResult, error) {
+	ns := mcpgrafana.DashboardNamespace(ctx)
+
+	obj, err := k8s.Get(ctx, dashboardDescriptor(dashboardReadVersion), ns, uid)
+	if err != nil {
+		return nil, fmt.Errorf("get dashboard via k8s api %s: %w", uid, err)
+	}
```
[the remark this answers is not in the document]
madaraszg-tulip: Addressed in 1dfdf46, but deliberately **not** by falling back to the legacy API on a k8s 404 — that would conflict with the higher-severity "Discovery failure uses lossy legacy" finding above. The legacy `/api/dashboards/uid` endpoint is org-scoped and would resolve the dashboard regardless of namespace, but for a **v2** dashboard it returns the lossy v1 down-conversion (and a subsequent patch would persist that, corrupting the stored v2). So "on 404, try legacy" reintroduces exactly the silent v2 data loss we're trying to avoid.

A clean "fail closed on a guessed namespace" isn't possible either: `orgNamespace(1)` legitimately returns `"default"` on-prem, which is indistinguishable from "frontend settings failed, defaulted to `default`", and we can't tell Cloud from on-prem *without* frontend settings. Gating on that would break on-prem during transient settings blips.

Instead, the namespace is resolved authoritatively from `/api/frontend/settings` (correct for Cloud `stacks-{id}` and on-prem `org-N`/`default`, cached per URL+org), and `DashboardNamespace` now reports whether it came from settings or the OrgID-derived fallback. A 404 that occurs while using the **fallback** namespace now returns a diagnostic error — e.g. `dashboard "X" not found in namespace "default", which was derived from the org id because /api/frontend/settings was unavailable; the namespace may be wrong (e.g. Grafana Cloud uses "stacks-{id}")` — rather than a bare "not found", so a misresolved namespace is diagnosable without masking real v2 content behind a lossy legacy read.

tools/dashboard.go
```diff
@@ -143,8 +342,90 @@ func updateDashboardWithPatches(ctx context.Context, args UpdateDashboardParams)
...
+	status := "success"
+	body := &models.PostDashboardOKBody{
+		UID:    &uid,
+		Status: &status,
+	}
+	if metadata, ok := obj["metadata"].(map[string]interface{}); ok {
+		if generation, ok := metadata["generation"].(float64); ok {
+			version := int64(generation)
+			body.Version = &version
+		}
```
sd2k, line 383: q: are these definitely the same thing? (legacy 'version' and k8s 'generation')? An agent relying on this for follow-ups could run into surprising behaviour if not, worth a check/comment.

tools/dashboard.go
```diff
@@ -162,6 +443,101 @@ func updateDashboardWithFullJSON(ctx context.Context, args UpdateDashboardParams
...
+	uid := args.UID
+	if uid == "" {
+		if u, ok := spec["uid"].(string); ok {
+			uid = u
+		}
+	}
+	delete(spec, "uid")
+
+	ns, _ := mcpgrafana.DashboardNamespace(ctx)
+	version := "v2beta1"
```
sd2k, line 511: I think it's fine that we're using this for now, but it would be great to negotiate the version we should use somehow. Not a blocker for this PR but it'd be great to have a TODO comment here.

tools/dashboard.go
sd2k, line 458: This field seems to be ignored by `updateDashboardV2` and `createOrUpdateDashboardV2`, is there a way we can support it? Claude suggests a `grafana.app/message` annotation which the [docs](https://grafana.com/docs/grafana/latest/developer-resources/api-reference/http-api/dashboard/#update-dashboard) support.

tools/dashboard.go
```diff
@@ -20,19 +22,209 @@ import (
...
+	ns, nsFromSettings := mcpgrafana.DashboardNamespace(ctx)
+
+	obj, err := k8s.Get(ctx, dashboardDescriptor(dashboardReadVersion), ns, uid)
+	if err != nil {
+		// A 404 when the namespace was only derived from the org id (frontend
+		// settings was unavailable) may be a wrong-namespace miss rather than a
+		// truly missing dashboard — e.g. on Grafana Cloud the real namespace is
+		// "stacks-{id}", not "default"/"org-N". Surface that so it's diagnosable.
+		if isK8sNotFound(err) && !nsFromSettings {
+			return nil, fmt.Errorf("dashboard %q not found in namespace %q, which was derived from the org id because /api/frontend/settings was unavailable; the namespace may be wrong (e.g. Grafana Cloud uses \"stacks-{id}\"): %w", uid, ns, err)
```
sd2k: This diagnostic is quite nice, can we use it in the write (update) path as well?

madaraszg-tulip: Done in a435921. Factored the diagnostic into a shared `k8sDashboardErr(action, uid, ns, nsFromSettings, err)` helper and applied it to the write paths too — `updateDashboardV2` (PUT) and `createOrUpdateDashboardV2` (Create/Update) now produce the same "namespace may be wrong (Grafana Cloud uses stacks-{id})" hint on a 404 with a fallback-derived namespace, not just the read path.

tools/dashboard.go
madaraszg-tulip: Done in a435921. `message` is now written as the `grafana.app/message` annotation (confirmed against the Grafana source: `AnnoKeyMessage = "grafana.app/message"`), applied on both the patch and full-JSON create/update paths alongside `folderUid` via a new `applyV2WriteMetadata` helper.

tools/dashboard.go
madaraszg-tulip: Added a TODO in a435921: negotiate the write version by discovering the served v2 version via GET /apis/dashboard.grafana.app (could be v2/v2beta1/v2alpha1) instead of hardcoding v2beta1. Note this default only applies when creating a brand-new dashboard — for an existing one we already reuse its stored version.

tools/dashboard.go
madaraszg-tulip: Checked — yes, they're the same by construction. Grafana's own unstructured->legacy conversion sets `spec["version"] = obj.GetGeneration()` (pkg/services/dashboards/service/dashboard_service.go), so the k8s `metadata.generation` we surface as the legacy `version` matches what a legacy read reports. Added a comment documenting this in a435921.

sd2k (approved):
Thanks! Looks great, I've been meaning to do this for a long time.

PR #939, opened 2026-06-11
tools: datasource management

jcolladokuri:
## Summary

This PR expands the existing datasources tool set to be able to:

- create and update datasources with schema guidance based on the datasource type and health check to make sure the updates are working as expected
    - if a call is made to create a datasource with an ambiguous type (i.e. azure), the tool relies on `search_plugin_information` to help fetch related datasources and ask user for clarification.
    - if a datasource creation is attempted against a plugin that is not yet installed, the tool will install it first, then proceed with creation.
- run health checks on single ( by UID ) or multiple (by type) or all datasources in a grafana instance

## Notes
- right now the datasource schemas are as static JSONs, we are in the process of getting these schemas to be exposed via the plugin, and once available, we will have a follow up PR to replace these static files with an API call.
- This PR comes from multiple previous PRs that were approved and merged into the feature branch:
    1. https://github.com/grafana/mcp-grafana/pull/784
    2. https://github.com/grafana/mcp-grafana/pull/812
    3. https://github.com/grafana/mcp-grafana/pull/886 

---

> [!NOTE]
> **Medium Risk**
> Adds write paths that mutate Grafana datasource configuration at scale; mitigations include read-only gating, secret field exclusion from guidance, and a mandatory schema review step, but misconfiguration or partial field mapping could still break datasources.
> 
> **Overview**
> Expands MCP **datasource** tooling so agents can **create and update** Grafana datasources with **type-specific schema guidance**, optional **post-create health checks**, and **bulk health checks** (by UID, type, or all).
> 
> **Schema-backed two-step create/update:** For plugin types with embedded `*_schema.json` files, the first `create_datasource` / update call returns **LLM-oriented field guidance** (common fields like `uid` / `isDefault` plus safe `root` / `jsonData` fields). **Secrets, `secureJsonData`, and credential usernames are omitted** from guidance. A second call with **`schemaReviewed=true`** and collected **`fields`** actually provisions the datasource. Types without a schema still prompt for explicit params (at minimum **`name`**) before creation.
> 
> **Infrastructure wiring:** `AddDatasourceTools` now takes an **`enableWriteTools`** flag so create/update register only when writes are allowed (aligned with `--disable-write` / read-only server mode). Examples and `cmd/mcp-grafana` pass this through like other write-gated categories.
> 
> **Coverage and tests:** Adds a large set of static per-plugin schema JSONs plus Go helpers to **embed, load, and filter** them. New Python integration test exercises the **`create_datasource`** two-call flow end-to-end via an LLM tool loop.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit daa71c0f84ab03780c0fe7df3a76b7609b77d527. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


tools/datasources.go
```diff
@@ -87,6 +92,209 @@ func listDatasources(ctx context.Context, args ListDatasourcesParams) (*ListData
...
+				}
+			case "access":
+				if s, ok := v.(string); ok {
+					body.Access = models.DsAccess(s)
+				}
+			case "withCredentials":
+				if b, ok := v.(bool); ok {
+					body.WithCredentials = b
+				}
+			}
```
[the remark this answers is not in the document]
sd2k: @jcolladokuri this seems legit, is it correct?

sd2k (commented):
There are a few Cursor Bugbot comments that seem correct to me, WDYT?

tools/datasources.go
```diff
@@ -87,6 +92,209 @@ func listDatasources(ctx context.Context, args ListDatasourcesParams) (*ListData
...
+		return nil, err
+	}
+	// Phase 1: return field guidance before creation.
+	// With a schema: list schema fields and ask the user to fill them in.
+	// Without a schema: list the explicit params and ask for name + any others
+	// the user wants to set, then call again with those values.
+	if schema != nil && args.Fields == nil {
+		text, _ := json.Marshal(datasourceschemas.BuildSchemaGuidance(schema, "create_datasource"))
+		return mcp.NewToolResultText(string(text)), nil
+	}
```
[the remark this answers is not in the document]
jcolladokuri: added flag to take care of this and distinguish between lack of schema review and an actual valid `fields: {}`

tools/datasources.go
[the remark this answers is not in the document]
jcolladokuri: yes! (and no??) we deliberately excluded `user*` fields to avoid entering any sensitive of PII information in chat, I forgot to exclude those fields from the schemaGuidance too, not just in this function.

I have changed it now so that those fields are excluded in both places.

tools/datasource_schemas/grafana-clickhouse-datasource_schema.json
```diff
@@ -0,0 +1,1232 @@
...
+            "key": "host",
+            "label": "Server address",
+            "description": "ClickHouse host address.",
+            "valueType": "string",
+            "target": "jsonData",
+            "required": true,
+            "storage": {
+                "type": "computed",
+                "read": "jsonData.server"
+            },
```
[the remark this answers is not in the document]
jcolladokuri: this is fine, server is a deprecated field, this is here for legacy reasons.

See: https://github.com/grafana/clickhouse-datasource/blob/main/pkg/plugin/settings.go#L87-L92 for reference

jcolladokuri:
@cursor push fa88598fba

tools/datasources.go
```diff
@@ -185,7 +410,278 @@ var GetDatasource = mcpgrafana.MustTool(
...
+
+	resp, err := client.httpClient.Do(req)
+	if err != nil {
+		return nil, fmt.Errorf("check datasource health %s: %w", args.UID, err)
+	}
+	defer func() { _ = resp.Body.Close() }()
+
+	if resp.StatusCode != http.StatusOK {
+		body, _ := io.ReadAll(resp.Body)
```
ioanarm: Can we bound this? We usually add something like this 
`body, _ := io.ReadAll(io.LimitReader(resp.Body, 1024))`

tools/datasources.go
```diff
@@ -185,7 +410,278 @@ var GetDatasource = mcpgrafana.MustTool(
...
+	mcp.WithIdempotentHintAnnotation(true),
+	mcp.WithReadOnlyHintAnnotation(true),
+)
+
+func AddDatasourceTools(mcp *server.MCPServer, enableWriteTools bool) {
 	ListDatasources.Register(mcp)
 	GetDatasource.Register(mcp)
+	CheckDatasourcesHealth.Register(mcp)
+	if enableWriteTools {
+		// TODO: since these tools are more for set up / admin, when merging to main, we should move them to the admin toolset
```
ioanarm: Should we do this?

tools/datasources.go
```diff
@@ -185,7 +410,278 @@ var GetDatasource = mcpgrafana.MustTool(
...
+	JSONData  map[string]interface{} `json:"jsonData,omitempty" jsonschema:"description=Non-secret plugin settings; replaces existing jsonData when set"`
+}
+
+type UpdateDatasourceResult struct {
+	Message string                  `json:"message,omitempty"`
+	Health  *DatasourceHealthResult `json:"health,omitempty"`
+}
+
+func updateDatasource(ctx context.Context, args UpdateDatasourceParams) (*UpdateDatasourceResult, error) {
+	// pending add credential violation check
```
ioanarm: Is this leftover?

tools/datasources.go
```diff
@@ -185,7 +410,278 @@ var GetDatasource = mcpgrafana.MustTool(
...
+		for _, ds := range resp.Payload {
+			if uidSet[ds.UID] {
+				all = append(all, ds)
+			}
+		}
+	} else {
+		all = filterDatasources(resp.Payload, args.Type)
+	}
+
+	limit := 10
```
ioanarm, line 618: This limit matches `cap(sem)` below, so the semaphore doesn't really throttle anything, is it meant to, or can it go?

ioanarm (commented):
Some comments but lgtm! thanks! 🚀 

Can we also add some doc comments on the exported functions? Oneliners is more than enough. 

eg
```
DsSchemaField
DatasourceSchema
LoadDatasourceSchema
```

tools/datasources.go
jcolladokuri: no! sorry, this was leftover from previous conversations but double checked with @sd2k & we should be good to leave them in the datasources toolset for discoverability purposes :)

tools/datasources.go
jcolladokuri: also a leftover comment sorry, we originally had added credential violation checks but given how late in the flow it was happening we opted for removing it and relying on removing the parameters from the tool.

see this [comment](https://github.com/grafana/mcp-grafana/pull/784#pullrequestreview-4279022020) for more context

tools/datasources.go
jcolladokuri: thank you for the catch on this. I removed the semaphore, and added a timeout to bound concurrency by that instead.

tools/datasources.go
jcolladokuri: done :)

jcolladokuri:
@cursor push 0e30b6944c

jcolladokuri:
@cursor push 9b496540ff

ioanarm (approved):
Thanks for addressing the comments! g2g!

PR #941, opened 2026-06-11
feat(tools): Support Quickwit datasources

RazCrimson:
Adds a new `quickwit` tool category (disabled by default) with a `query_quickwit` tool for querying [Quickwit](https://quickwit.io/) datasources (`quickwit-quickwit-datasource` plugin) via Lucene queries or partial Elasticsearch Query DSL, with time-range and limit support.

## Notes
 - Reusing helpers from the Elasticsearch backend as Quickwit exposes a ES-compatible `_elastic/_msearch` endpoint
 - Newer separate dedicated tool coz of 2 reasons (Can merge into the ES tool if we prefer that): 
      1. Quickwit has its own dedicated API which we might need to support in the future
      2. Easy to enable / disable separately (No extra tokens for existing ES/OS users)
      3. Other stuff like timeField config doesn't exist for Quickwit making it quite diff from ES / OS

## Test Details
- Has quickwit specific unit tests and integration tests.
- docker-compose test infra has a Quickwit service, seed script, and provisioned datasource.

---

> [!NOTE]
> **Medium Risk**
> New read-only query path plus a refactor of shared Elasticsearch `_msearch` code used by existing ES tooling; behavior changes there warrant careful review, though tests and integration coverage reduce risk.
> 
> **Overview**
> Adds an opt-in **quickwit** tool category with **`query_quickwit`**, which searches Quickwit Grafana datasources via Lucene or partial Elasticsearch Query DSL, time filters, and limits. The category is off by default; enable it with `quickwit` in `--enabled-tools` or avoid `--disable-quickwit`.
> 
> Implementation reuses shared Elasticsearch-style helpers: query DSL building moves to **`esSearchQuery`**, **`_msearch`** execution is centralized in **`executeMSearch`**, and URL joining is shared via **`buildURL`** (also wired into Loki and VictoriaLogs). Quickwit resolves the timestamp field from index metadata (including nanosecond sort formats) and calls Quickwit’s **`_elastic/_msearch`** endpoint through the Grafana proxy.
> 
> Docs, CLI flags, and README tables document the new tool. **docker-compose** gains Quickwit, a seed script, the Quickwit Grafana plugin, and a provisioned test datasource; unit and integration tests cover metadata parsing and query behavior. **`make lint`** also runs **`gofmt -l`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 780e73a66b3542363ca493ab1748479974f1c4e3. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


RazCrimson:
@cursor review

tools/es_backend.go
```diff
@@ -242,14 +234,19 @@ func (b *elasticsearchBackend) Search(ctx context.Context, index, query string,
 }
 
 // buildElasticsearchQuery constructs an Elasticsearch query DSL JSON object
-func buildElasticsearchQuery(query string, startTime, endTime *time.Time, size int, timeField string) map[string]interface{} {
+func buildElasticsearchQuery(query string, startTime, endTime *time.Time, size int, timeField string, sortFormat ...string) map[string]interface{} {
```
sd2k: Using `...string` for `sortFormat` here is a bit unidiomatic here, and this function is getting a bit unwieldy, I think we should refactor to a builder struct instead (or simply pass a required but empty string for default sort format).

tools/loki.go
```diff
@@ -91,21 +90,9 @@ func newLokiClient(ctx context.Context, uid string, _ *models.DataSource) (*Clie
...
-	fullURL := c.baseURL
-	if !strings.HasSuffix(fullURL, "/") && !strings.HasPrefix(urlPath, "/") {
-		fullURL += "/"
-	} else if strings.HasSuffix(fullURL, "/") && strings.HasPrefix(urlPath, "/") {
-		// Remove the leading slash from urlPath to avoid double slash
-		urlPath = strings.TrimPrefix(urlPath, "/")
-	}
-	return fullURL + urlPath
-}
-
```
sd2k, line 105: Thanks for the refactor 🙏

tools/url_helpers_test.go
```diff
@@ -0,0 +1,17 @@
...
+
+func TestBuildURL(t *testing.T) {
+	base := "http://proxy/api/datasources/proxy/uid"
+
+	assert.Equal(t, base+"/_msearch", buildURL(base, "/_msearch"))
+	assert.Equal(t, base+"/indexes", buildURL(base, "indexes"))
+	assert.Equal(t, base+"/indexes", buildURL(base, "/indexes"))
+	assert.Equal(t, base+"/indexes", buildURL(base+"/", "indexes"))
+	assert.Equal(t, base+"/indexes", buildURL(base+"/", "/indexes"))
+}
```
sd2k, line 49: It'd be good to test the case when `base` has a trailing slash too. I suspect you could use a table test here!

sd2k (commented):
Generally looks great! A couple of notes inline but happy to get this in soon.

tools/url_helpers_test.go
RazCrimson: We already have the case, but I guess it was missed due to the poor readability. Have make it into table tests with names for each case too

tools/es_backend.go
RazCrimson: @sd2k PTAL when possible
Went with the builder struct as the function was already quite long and splitting it up felt better.

docker-compose.yaml
```diff
@@ -163,6 +164,26 @@ services:
       - ./testdata/opensearch-seed.sh:/seed.sh
     entrypoint: ["sh", "/seed.sh"]
 
+  quickwit:
+    image: quickwit/quickwit:0.8.2
+    command: ["run"]
+    ports:
+      - "7280:7280"
+    healthcheck:
+      test: ["CMD-SHELL", "curl -sf http://localhost:7280/health/livez || curl -sf http://localhost:7280/api/v1/version"]
```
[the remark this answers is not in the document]
RazCrimson: Hmm... weird works locally. Maybe I using some cached image or something. Will check once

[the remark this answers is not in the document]
RazCrimson: The official docker file adds it into the image so for the current version, we should be good.

RazCrimson:
Thanks for fast reviews and clean comments!

sd2k (approved):
Thanks again! 🥳

PR #943, opened 2026-06-12
feat(multi-org): support a per-call orgId across all tools

### LINKED ISSUE #883: `get_panel_image` should support `orgId` when rendering panels
## Description
Grafana MCP supports working with Grafana resources across multiple organizations, but the `get_panel_image` tool does not currently expose or pass an `orgId` parameter when rendering a panel image.

Because of this, panel rendering appears to fall back to the default Grafana organization context, which is usually org `1`. This makes `get_panel_image` unable to render panels from dashboards that belong to other orgs, even when the Grafana token/user has access.

## Current Behavior
`get_panel_image` accepts parameters such as:
- `dashboardUid`
- `panelId`
- `timeRange`
- `variables`
- `width`
- `height`
- `scale`
- `theme`
- `timeout`
However, it does not accept `orgId`, and the render request does not include `orgId`.

## Expected Behavior
`get_panel_image` should support an optional `orgId` parameter and include it in the Grafana render request, for example:
```json
{
  "dashboardUid": "example_uid",
  "panelId": 1,
  "orgId": 2
}

### PR DESCRIPTION
**Pull request by madaraszg-tulip**
## Summary

Adds dynamic multi-organization support so a single connection can operate across Grafana organizations.

Startup-time multi-org already exists (`GRAFANA_ORG_ID` / the `X-Grafana-Org-Id` header fix the org for a whole connection). This PR adds **opt-in per-call selection**: with `--dynamic-multi-org`, tools accept an optional `orgId` argument to target an organization per call.

**Default off**, so tool schemas are identical to `main` (no token-baseline change); the feature is wired in only when the flag is set.

### Per-call `orgId` on native tools
- `--dynamic-multi-org` flag (default off). When on, every native tool that addresses a Grafana organization advertises an optional `orgId`. `OrgIDOverrideMiddleware` reads it, overrides `GrafanaConfig.OrgID` for that call — driving both the `X-Grafana-Org-Id` header (legacy `/api/*`) and the resolved Kubernetes namespace (`/apis/*`) — and strips it from the arguments.
- `orgId` is injected into the schema at registration time via a typed injector (not raw-JSON mutation), so it stays cleanly gated by the flag. Tools that reach no Grafana instance — `search_docs`, `get_doc`, `suggest_loki_alloy_label_config`, `get_query_examples` — opt out via `Tool.NotOrgScoped()`: the argument would be inert there, and advertising it suggests the answer is org-scoped when it cannot be.

### `get_panel_image`
- **Behaviour change against v1.2.0:** the tool no longer declares its own `orgId` argument, using the injected per-call one instead so it behaves like every other org-scoped tool. The argument is therefore only advertised with `--dynamic-multi-org`, and without that flag a call passing `orgId` is rejected as an unknown argument rather than ignored, since undeclared arguments now fail the strict argument check. Rendering an organization other than the connection's now requires the flag.
- The org reaches the tool through `GrafanaConfig`, so one fixed by `GRAFANA_ORG_ID` or the `X-Grafana-Org-Id` header now also sets `targetOrgId`, which the argument-only version did not do.
- Renders honor `orgId` via the `targetOrgId` URL query param. The render endpoint ignores the `X-Grafana-Org-Id` header, and the `orgId` URL param would persist a user-level org switch (via the frontend's `/api/user/using`), whereas `targetOrgId` scopes the org to the single render. **Fixes #883.**
- The dashboard deeplink returned alongside the PNG carries no org, so it is emitted only when it would open in the org the image was rendered from. `?orgId=N` on a UI link is intercepted by Grafana's `OrgRedirect` middleware, which persists the switch onto the viewer's user record, and neither `targetOrgId` (render backend only) nor `X-Grafana-Org-Id` (not sendable by a browser) substitutes — so for a cross-org render the image is returned without a deeplink rather than with one that resolves to a different dashboard (UIDs are unique only within an org). This also means no MCP Apps panel-viewer fallback for that case. The check compares against `/api/user`, which reports the org stored on the user record and ignores `GRAFANA_ORG_ID` and `X-Grafana-Org-Id` — so unlike a check on the argument alone, it also withholds the link when the organization is fixed on the connection rather than passed per call.

### Proxied datasource tools
- When the flag is on, datasource discovery runs across every org the credential can access (`/api/user/orgs`; service-account tokens stay single-org), in parallel, tagging each datasource with the org it was found in. Tools remain flat (`{type}_{tool}`, deduped across orgs) with an optional `orgId`; clients are keyed by `(org, type, uid)` so the same UID in two orgs maps to distinct connections, each with the org baked into its transport. This covers datasource types that exist only in a non-default org.
- A call that omits `orgId` targets the connection's default org, resolved once during discovery.
- Per-org discovery errors are reported only when *every* org fails, so a total failure is not mistaken for "found nothing" and one unreachable org does not discard the orgs that did resolve.

### `user_info` tool
- New default-on `user` category. Reports the current identity, whether it is a Grafana admin, the org this connection's calls target (not the org persisted on the user record, which ignores `GRAFANA_ORG_ID` and `X-Grafana-Org-Id`), and the organizations the credential can access (with roles), so clients can discover valid `orgId` values. Includes an adaptive usage note only when more than one org is accessible.

### Namespace resolution
- `DashboardNamespace` → `GrafanaNamespace` (the namespace is instance/org-wide, not dashboard-specific). It now errors when `/api/frontend/settings` is unreachable instead of guessing the org-derived namespace, which would silently misroute on Grafana Cloud (`stacks-{id}`).

## Testing

- **Unit**: override middleware (incl. arg stripping), namespace resolution, render URL, org enumeration (user/SA fallback, and the default org absent from the membership list), org-keyed proxied client lookup with default-org normalization, org-keyed `GetServerClient`, schema gating, and `user_info`.
- **Integration**: `orgId` across the namespaced (`/apis/*`) dashboard & provisioning paths, the legacy (`/api/*`) dashboard path, image rendering (with a guard that rendering does not change the user's active org), proxied cross-org discovery, and `user_info` — backed by a new `orgs-seed` docker-compose job.
- Also checked behaviourally: with the flag off the advertised tool schemas are unchanged and `orgId` is rejected as an unknown argument; with it on, a Tempo datasource that exists only in org 2 resolves with `orgId: 2` and returns data, while the same call without `orgId` falls back to the default org and fails.

## Docs

README and `docs/sources` updated: the `user` category and `user_info` tool, the `--dynamic-multi-org` and `--disable-user` flags, and a "select an organization per tool call" section.

## Issues

Fixes #883. Supersedes the tool-specific `get_panel_image` orgId argument merged in #903: this replaces it with the injected per-call argument so every org-scoped tool behaves the same (see the behaviour change noted above).

### CODE REVIEW

**Comment by sd2k** [comment]
This is a nice idea, but I am concerned about the amount of tokens it will add. We have a lot of tools, and the description alone is 50 tokens; after adding on the schema it will be even larger. Some ideas for improving it:

- add a config flag to enable/disable multi-org
- only add for tools where it actually makes sense (I don't think it's needed for OnCall, Incident, proxied tools, for example)
- reduce the size of the description

(You can use `make token-baseline` on `main` and `make token-check` to compare the sizes locally, if you have https://github.com/sd2k/mcp-tokens installed. When I ran it I got this:

```
Baseline Comparison
============================================================

Baseline: 18858 tokens
Current:  22263 tokens
Change:   +3405 tokens (+18.1%)
```
)

**Comment by madaraszg-tulip** [comment]
Thanks @sd2k — and thanks for the reminder to keep an eye on the token budget. I made per-call `orgId` **opt-in** rather than always-on:

- **Config flag (default off):** `--dynamic-multi-org`. With it off, tool schemas are byte-identical to `main`, so the token baseline is unchanged — the `orgId` argument and the override middleware are only wired in when it's enabled. No cost for the common single-org case.
- **Shorter description:** trimmed the `orgId` argument description to a single line.
- **"Only where it makes sense":** `proxied` tools never picked up `orgId` automatically anyway — they don't go through the tool-conversion path where the param is injected (they're registered by the proxy manager from the upstream schema), and they're discovered per-org and keyed by `datasourceUid`. So I handled them explicitly: cross-org discovery plus an `orgId` that routes to the right org's datasource. For OnCall/Incident/Sift/Asserts the override *does* thread through (they build their client via `BuildTransport` with the org header, through Grafana's per-org plugin proxy), so whether it's *useful* there is a product question about whether those backends partition by org — I couldn't justify excluding them from the code alone. Since it's opt-in, the cost only applies when you turn it on.

I also added a `user_info` tool so clients can discover which `orgId` values are valid before passing one.

**Comment by sd2k** [comment]
@madaraszg-tulip Hi again, sorry for the radio silence, I've been on parental leave for a while but I'm back now. Your latest changes seem sensible to me, I'm much happier merging now that they don't affect users unless they need multi-org. Unfortunately this has become quite stale now, but I'd be happy to re-review with the aim of getting this merged if you have time.

File: tools.go [hunk 1]
```diff
@@ -44,13 +67,35 @@ func (e *HardError) Unwrap() error {
...
+	// Clone so the retained schema stays pristine if Register runs again.
+	properties := maps.Clone(t.inputSchemaProperties)
+	injectOrgIDProperty(properties)
+	raw, err := buildInputSchema(t.Tool.Name, t.inputSchemaType, properties, t.inputSchemaRequired)
+	if err != nil {
+		return t.Tool
+	}
+	resolved := t.Tool
+	resolved.RawInputSchema = raw
+	return resolved
```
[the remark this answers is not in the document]
madaraszg-tulip: Real for the un-shortened link, and the distinction drawn at the end is the sharp
part: `get_panel_image` can withhold the deeplink and still return the image,
whereas here the link is the entire result, so withholding is not available as a
fix.

Two corrections though, both narrowing it.

`orgId` is not inert: with `shorten=true` the tool POSTs to `/api/short-urls`,
which creates the short URL in the request's org. And the link can carry an
organization — Grafana's own API returns it that way:

```
POST /api/short-urls   (X-Grafana-Org-Id: 2)
{"uid":"ffwfv2y6hh2iob","url":"http://localhost:3000/goto/ffwfv2y6hh2iob?orgId=2"}
```

`normalizeShortURLWithPublicBase` keeps the path, query and fragment and only
swaps scheme/host, so that `orgId` survives into what the tool returns. Checked
against a two-org instance:

```
shorten=true,  orgId=2  ->  http://localhost:3000/goto/efwfv6rcsqg3ke?orgId=2
shorten=false, orgId=2  ->  http://localhost:3000/d/mcp-orgid-ns
```

So the hazard is the second line only — the default path. Worth noting the
shortened link works cross-org precisely because it carries `?orgId=`, which
trips `OrgRedirect` and persists the viewer's org switch: the side effect
deliberately avoided for the `get_panel_image` deeplink, chosen here by Grafana
itself.

It also predates `--dynamic-multi-org`: a server started with `GRAFANA_ORG_ID=2`,
or a client sending `X-Grafana-Org-Id: 2`, already gets an un-shortened deeplink
with no org in it, which opens in the viewer's persisted org where the same UID
can be a different dashboard. The flag only makes it reachable per call.

orgid.go
```diff
@@ -0,0 +1,255 @@
...
+func orgIDFromArguments(args map[string]any) (int64, bool) {
+	raw, present := args[OrgIDArgument]
+	if !present {
+		return 0, false
+	}
+
+	var orgID int64
+	switch v := raw.(type) {
+	case float64:
+		orgID = int64(v)
```
sd2k, line 110: nit: let's reject fractional inputs here, we shouldn't coerce e.g. 2.9 to 2.

sd2k (dismissed):
This LGTM, thank you for this change and adding the flag!

Just added one nit, I think we can ship this though.

orgid.go
madaraszg-tulip: done, with tests

madaraszg-tulip:
> This LGTM, thank you for this change and adding the flag!
> 
> Just added one nit, I think we can ship this though.

Thanks. The situation with `generate_deeplink` is not ideal, however I don't know what the best solution is, perhaps it could be improved in a followup PR.

sd2k (approved):

sd2k:
Thanks! Yeah the `generate_deeplink` thing is tricky but not worth blocking this on, we can address it later if it becomes an issue for users.

PR #945, opened 2026-06-13
fix: pass cfg.HTTPTransport() to propagate headers to loki downstream calls

AlexMacocian:
> [!NOTE]
> **Low Risk**
> Small behavioral fix in shared HTTP middleware with a focused regression test; low blast radius unless callers depended on ignoring BaseTransport when base was nil.
> 
> **Overview**
> When **`BuildTransport`** is called with a **nil** base round tripper, it now uses **`cfg.HTTPTransport()`** instead of always **`http.DefaultTransport`**. That means a configured **`BaseTransport`** (and the same default fallback as elsewhere) is used as the innermost layer, so auth, extra headers, org ID, and other middleware still wrap the transport the host actually configured.
> 
> This aligns nil-base call sites—such as **frontend settings**, **incident**, and **Loki downstream** HTTP clients—with **`NewGrafanaClient`**, which already relied on **`BaseTransport`** for custom pooling, tracing, or forwarded headers.
> 
> A test confirms that with **`BaseTransport`** set, **`BuildTransport(cfg, nil)`** routes requests through that base and still applies middleware (e.g. **`Authorization`**).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1f120056804642982de4fd970716e95ab8e57d88. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/mcpgrafana.go
+++ b/mcpgrafana.go
@@ -662,7 +662,7 @@ func BuildTransport(cfg *GrafanaConfig, base http.RoundTripper, opts ...Transpor
 	}
 
 	if base == nil {
-		base = http.DefaultTransport
+		base = cfg.HTTPTransport()
 	}
 	transport := base
 
--- a/mcpgrafana_test.go
+++ b/mcpgrafana_test.go
@@ -1071,6 +1071,31 @@ func TestBuildTransport(t *testing.T) {
 		require.NotNil(t, transport)
 	})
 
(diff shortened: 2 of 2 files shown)


sd2k (approved):
Thanks!

PR #946, opened 2026-06-14
feat(prometheus): surface query warnings to callers

bdchatham:
**Problem**

The Prometheus query path discards the client's `v1.Warnings` return value (`result, _, err := b.api.Query(...)`). On a Thanos datasource with partial-response enabled, a cross-cluster blip returns HTTP 200 with partial data and a non-empty `warnings` array.  Today that's dropped, so a partial aggregate reads as complete and callers silently trust it.

**Fix**
`promBackend.Query` now returns `v1.Warnings`; the native Prometheus backend threads through what the client already returns; `query_prometheus` exposes a new `Warnings []string` field on `QueryPrometheusResult` (`json:"warnings,omitempty"`).

**Compatibility**

Additive on the tool output (field omitted when empty). The only signature change is the internal `promBackend.Query` method; all implementers (native Prometheus, Cloud Monitoring, VictoriaMetrics) are updated.

**Tests**

Added `tools/prom_backend_test.go` covering instant and range queries with and without a `warnings` payload.

---

> [!NOTE]
> **Low Risk**
> Additive MCP response field and internal interface change with all backends updated; no change to query semantics when warnings are absent.
> 
> **Overview**
> **`query_prometheus`** now includes an optional **`warnings`** array on its result so callers can see datasource-reported issues (e.g. Thanos partial responses) instead of treating HTTP 200 + partial data as fully complete.
> 
> The internal **`promBackend.Query`** contract returns **`promv1.Warnings`** alongside the series data. The native Prometheus/Thanos path forwards warnings from instant and range API calls; Cloud Monitoring and VictoriaMetrics keep returning empty warnings. **`queryPrometheusWithWarnings`** carries warnings through the stack while **`queryPrometheus`** (used by histogram and other callers) still returns data only.
> 
> Unit tests in **`prom_backend_test.go`** cover instant/range responses with and without a **`warnings`** payload.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a2eecc687fe20aa2c0fc2b456b6658383390f264. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -10,6 +10,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Added
 
 - Optional `startRfc3339`/`endRfc3339` time range parameters for `list_prometheus_metric_names` to restrict results to metrics active within a window
+- `query_prometheus` now surfaces datasource `warnings` (e.g. partial responses from Thanos) in its result
 
 ## [0.15.2] - 2026-06-04
 
--- a/tools/prom_backend.go
+++ b/tools/prom_backend.go
@@ -19,8 +19,10 @@ import (
 // promBackend abstracts the differences between datasource types that support
 // PromQL-compatible queries (native Prometheus, Cloud Monitoring, etc.).
 type promBackend interface {
-	// Query executes a PromQL query (instant or range) and returns the result.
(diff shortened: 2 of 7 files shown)


sd2k (approved):
LGTM, thanks!

PR #967, opened 2026-06-25
Add Gamut to MCP client examples

iddogino:
Adds [Gamut](https://www.gamut.so) as an example MCP client alongside Claude Desktop and Cursor.

Gamut is an AI agent platform with native MCP support that can connect to the Grafana MCP server.

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,7 +12,7 @@ This provides access to your Grafana instance and the surrounding ecosystem.
 
 ## Quick Start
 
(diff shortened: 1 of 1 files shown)


sd2k:
Hi, rather than including in the README I'd accept a PR that adds some client instructions for Gamut to https://github.com/grafana/mcp-grafana/tree/main/docs/sources/clients if you'd like to add one in a separate PR? Thanks!

Issue #963, opened 2026-06-23
Alert rules discovered through MCP have empty/missing UIDs on Grafana 9.4.3


woelfel-tn:
## Summary

When using Grafana MCP with Grafana OSS 9.4.3, alert rules are returned without usable UIDs.

As a result, MCP can discover that alert rules exist, but it cannot reliably retrieve the full configuration of those rules because subsequent operations require a valid rule UID.

This significantly limits alerting-related MCP functionality and prevents AI agents from inspecting alert rule definitions.

## Environment

* Grafana MCP: <version>
* Grafana OSS: 9.4.3
* Unified Alerting enabled

## Expected Behavior

Alert rules returned by MCP should contain valid Grafana rule UIDs.

The UID is required to retrieve the full configuration of a rule and to allow AI agents to inspect, understand, and potentially modify alerting configuration.

## Actual Behavior

Alert rules are returned with empty or missing UIDs.

Because no valid UID is available, MCP cannot retrieve the complete rule configuration for those rules.

## Impact

Without rule UIDs:

* Alert rules cannot be uniquely identified.
* The full alert rule configuration cannot be retrieved.
* AI agents can see that rules exist but cannot inspect their complete definitions.
* Alert management workflows are incomplete or non-functional.

## Reproduction

1. Connect Grafana MCP to a Grafana OSS 9.4.3 instance.
2. Query alert rules through MCP.
3. Observe that returned rules have empty or missing UIDs.
4. Attempt to retrieve additional rule details that require a UID.

## Debug Information

The MCP logs show successful requests to Grafana:

```text id="rq9j1k"
time=2026-06-23T18:40:14.022Z level=DEBUG msg="GET /api/prometheus/grafana/api/v1/rules HTTP/1.1"
time=2026-06-23T18:40:14.117Z level=DEBUG msg="HTTP/1.1 200 OK"
```

The request succeeds and Grafana returns alert rule data.

To verify that this was not caused by MCP itself, I queried the same Grafana endpoint directly and observed the same behavior: the returned rules did not contain usable UIDs.

## Comparison With Newer Grafana Versions

I tested the same MCP workflow against Grafana 11 and the issue does not occur there.

On Grafana 11, alert rules are returned with usable UIDs and MCP can successfully retrieve the associated rule configuration.

This suggests either:

* a compatibility issue with older Grafana versions such as 9.4.x,
* or that MCP currently relies on metadata that is not available in older Grafana releases.

## Suggested Improvements

[...]

PR #968, opened 2026-06-28
Recover Grafana-managed alert rule UIDs from internal labels

jstar0:
## Summary

Recover Grafana-managed alert rule UIDs from internal labels when older Grafana responses omit top-level UID fields.

## Changes

- Request Grafana internal labels when listing Grafana-managed alert rules.
- Use `__alert_rule_uid__` and `__alert_rule_namespace_uid__` as fallbacks when older Grafana responses omit top-level rule and folder UIDs.
- Strip those internal labels from the returned summary labels so they do not appear as normal user labels.

## Why

Older Grafana responses from `/api/prometheus/grafana/api/v1/rules` can omit the top-level alert rule UID. That makes `alerting_manage_rules` return empty UIDs for discovered rules, which prevents callers from using the returned result to fetch a rule by UID.

Grafana exposes the missing identifiers through internal labels when `includeInternalLabels=true` is requested, so this keeps the existing endpoint while recovering the identifiers needed by the MCP tool.

Part of #963.

## Verification

```sh
go test -tags unit ./tools -run 'TestGetRulesOpts_queryValues|TestManageRules_ListRules/list_with_defaults|TestConvertRulesResponseToSummary|TestFindRuleInResponse' -count=1
go test -tags unit ./tools -count=1
make test-unit
```


jstar0:
recheck

sd2k (approved):
LGTM, sorry for the slow review

ioanarm (approved):

PR #973, opened 2026-06-29
feat(datasources): support name filter on list_datasources

### LINKED ISSUE #972: Allow `list_datasources` tool to filter by name
## Problem

On Grafana instances with thousands of datasources (6k+ is realistic in large multi-tenant or multi-environment deployments), `list_datasources` returns a payload that's prohibitively expensive for the LLM context window — even with pagination, the model often has to page through many chunks before finding the datasource it needs.

`type` filtering (added in #52) helps narrow by plugin (e.g. `prometheus`, `loki`), but at this scale a single type can still resolve to hundreds or thousands of entries. When the user already knows part of the datasource name (e.g. `prometheus-prod-eu`), the LLM has no way to express that and falls back to paging or guessing UIDs — both wasteful.

The Grafana API (`GET /api/datasources`) doesn't support a server-side name filter, but — as with `type` in #52 — we can filter inside the tool before pagination and trim the response before it reaches the model.

## Proposal

Add a `name` parameter to `ListDatasourcesParams`:

```go
type ListDatasourcesParams struct {
    Type   string `json:"type,omitempty"   jsonschema:"..."`
    Name   string `json:"name,omitempty"   jsonschema:"description=Case-insensitive substring match on datasource name. For example\\, 'prod' matches 'prometheus-prod-eu' and 'loki-prod-us'."`
    Limit  int    `json:"limit,omitempty"  jsonschema:"..."`
    Offset int    `json:"offset,omitempty" jsonschema:"..."`
}
```

Semantics: case-insensitive substring match on `ds.Name`, applied before pagination, composable with `type` (both must match). Empty `name` is a no-op, matching the existing `type` behavior at `tools/datasources.go:301`.

## Implementation sketch

1. Extend `filterDatasources` in `tools/datasources.go:301` to accept a name pattern alongside `type`, or add a sibling `filterByName` and chain them.
2. Update the `list_datasources` tool description at `tools/datasources.go:331`.
3. Add a unit test next to `TestListDatasources_TypeFilter` in `tools/datasources_unit_test.go:115` covering: empty `name` = no filter, substring match, case-insensitivity, and `name + type` composition.

Callers that reuse `listDatasources` internally (`tools/run_panel_query.go:707`, `tools/datasources.go:615`) keep working unchanged since `Name` is optional and zero-valued.

## Alternatives considered

- **Prefix-only match** — cheaper to reason about, but less useful when the discriminator is an environment/region suffix.
- **Regex / glob** — more powerful but easy to misuse and would diverge from the `type` filter's contains semantics.
- **MCP resources (#51)** — orthogonal: solves the initial discovery problem but doesn't help when the LLM still needs to list/filter at runtime, and 6k+ resources is itself a context-window problem.

PR available in #973

cbcoutinho:
## Summary

- Adds a `name` parameter to the `list_datasources` tool for case-insensitive substring matching on the datasource name, composable with the existing `type` filter (both must match).
- Motivated by Grafana instances with thousands of datasources (6k+ is realistic in large multi-tenant deployments): the `type` filter alone often returns too many entries for an LLM context window. With this change the model can express the partial name it already knows (e.g. `prometheus-prod-eu`) and skip pagination noise.

## Implementation

- Extends `filterDatasources` in `tools/datasources.go` to apply name and type filters together; both are case-insensitive substring matches, and an empty value is treated as a wildcard for that field (matching the existing `type` semantics).
- The Grafana API (`GET /api/datasources`) does not support server-side name filtering, so this is done inline before pagination — mirroring the precedent set in #52 when the `type` filter was added.
- The bulk health caller (`checkDatasourcesHealth`) continues to filter by type only; the signature update is a no-op there.
- Tool description updated to surface the new parameter.

## Tests

New `TestListDatasources_NameFilter` (unit, build-tag `unit`) covering:

- empty name = no filter
- substring match across types
- case-insensitive match
- composition with the `type` filter (AND semantics)
- no-match returns empty
- name filter respects pagination

Existing `TestListDatasources_TypeFilter`, `TestListDatasources_Pagination`, and `TestCheckDatasourcesHealth_*` all still pass.

## Test plan

- [x] `go build ./...`
- [x] `go test -tags=unit ./tools/`
- [x] `gofmt -l` clean on touched files
- [x] `make lint` — JSON schema and OpenAPI linters clean for the new field; only pre-existing govet failures on `mcpgrafana.go` / `tools.go` (unrelated to this change) remain

Fixes #972

---

_This PR was generated with the help of AI, and reviewed by a Human_

### DIFF
--- a/tools/datasources.go
+++ b/tools/datasources.go
@@ -25,6 +25,7 @@ const (
 
 type ListDatasourcesParams struct {
 	Type   string `json:"type,omitempty" jsonschema:"description=The type of datasources to search for. For example\\, 'prometheus'\\, 'loki'\\, 'tempo'\\, etc..."`
+	Name   string `json:"name,omitempty" jsonschema:"description=Case-insensitive substring match on the datasource name. For example\\, 'prod' matches datasources named 'prometheus-prod-eu' and 'loki-prod-us'. Useful on instances with many datasources where filtering by type alone returns too many results."`
 	Limit  int    `json:"limit,omitempty" jsonschema:"default=50,description=Maximum number of datasources to return (max 100)"`
 	Offset int    `json:"offset,omitempty" jsonschema:"default=0,description=Number of datasources to skip for pagination"`
 }
@@ -52,8 +53,8 @@ func listDatasources(ctx context.Context, args ListDatasourcesParams) (*ListData
 		return nil, fmt.Errorf("list datasources: %w", err)
 	}
 
-	// Filter by type if specified
-	datasources := filterDatasources(resp.Payload, args.Type)
+	// Filter by type and/or name if specified
+	datasources := filterDatasources(resp.Payload, args.Type, args.Name)
 	total := len(datasources)
 
 	// Apply default limit if not specified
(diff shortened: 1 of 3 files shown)


ioanarm (commented):
Can we cover this with an integration test as well? There is already one that you can add to `TestDatasourcesTools`. 
Non blocking though.

ioanarm (dismissed):

cbcoutinho:
Hi @ioanarm I've added an integration test for this new feature

ioanarm (approved):
lgtm! :)

PR #978, opened 2026-07-02
feat(loki): enforce label matchers on Loki queries

madaraszg-tulip:
## Summary

Adds `--loki-enforced-matchers`, an operator-configured set of LogQL label matchers that is AND-ed into **every** native-Loki query the server issues, restricting which log streams the MCP can ever read.

## Motivation

On OSS Grafana + OSS Loki there is no per-datasource or per-user label access control (LBAC is Enterprise/Cloud only; Loki's `required_labels` only checks presence, not values). We want to expose Loki to the MCP but guarantee certain streams — e.g. logs that may contain sensitive information — are never readable. Since the MCP must share the org/datasource set with human users, the only place to draw that boundary is inside the MCP itself, which is also the correct trust boundary (the server is operator-controlled; the LLM/user is the restricted party).

## How it works

- The matchers are parsed once at startup (invalid input aborts the server).
- For each query, every stream selector is located and the enforced matchers are appended to it — so multi-selector queries (binary ops, nested aggregations) are all covered. Because Loki AND-s matchers within a selector, a user query can only ever *narrow* within the enforced bounds; a conflicting selector yields an empty result.
- Selectors are found by scanning for `{...}` spans while tracking string, backtick and comment state, and each span is validated with the Prometheus parser this repo already depends on — a LogQL stream selector is Prometheus label-matcher syntax. Outside of strings and comments a `{` only ever opens a stream selector, which is what makes the scan reliable, and it is immune to the line-filter bypass (`{app="x"} |= "namespace=\"foo\""` — the literal lives in string state, so it is never mistaken for a selector).
- Matchers are spliced in as text before the closing brace, so the rest of the query keeps the user's own formatting rather than being re-rendered from an AST.
- **Fails closed**: a query whose selectors cannot be located, or which contains no stream selector at all, is rejected rather than sent unfiltered.

## Flags

- `--loki-enforced-matchers` — e.g. `environment=~"prod|staging"` (allowlist) or `namespace!~"vault"` (exclusion).
- `--loki-label-enumeration-fallback` — `reject` (default) or `unfiltered`, for the edge case where a purely-negative matcher can't scope `list_loki_label_*` (Loki rejects a standalone negative selector).
- `--instructions-append` — text appended to the server instructions returned to clients on initialize. Enforcement is invisible to a connecting agent otherwise: results are silently narrowed, so an LLM cannot tell a restricted stream from an empty one, and cannot warn the user. This lets the operator say so — along with anything else worth telling every connecting agent.

## Coverage & security completeness

Enforced: `query_loki_logs`, `query_loki_stats`, `query_loki_patterns`, `list_loki_label_names`, `list_loki_label_values`, and `run_panel_query` (reuses the enforced path). VictoriaLogs datasources are refused while enforcement is on (LogsQL can't be safely rewritten).

Enforcement only covers the Loki tools, so other tools can reach Loki data through server-side paths. The server logs a per-tool startup warning (honoring both `--enabled-tools` and `--disable-*`) for the live bypasses:

- `--disable-api` — `grafana_api_request` hits the datasource proxy directly (full bypass)
- `--disable-rendering` — `get_panel_image` renders Loki panels server-side
- `--disable-sift` — Sift analyzes Loki logs server-side across all streams
- `--disable-assistant` — `ask_assistant` delegates to Grafana Assistant, which reads Loki server-side across all streams. Only registered when write tools are enabled, so `--disable-write` closes it too

Plus INFO notes for proxied tools (Tempo-only today, not a Loki path) and snapshots.

## Testing

- Unit tests for the injector: bare/pipeline/metric/binary-op selectors, braces inside double-quoted, backtick, `#`, `//` and `/* */` contexts, `line_format` templates, a brace inside a selector's own label value, trailing commas, fail-closed cases, positive & negative matchers, and the enumeration fallback.
- Integration tests (against docker-compose Loki) proving a positive matcher restricts results, a conflicting user query returns empty, label enumeration is scoped, and negative matchers fail-closed on label tools while still enforcing content queries.

Full unit suite, the Loki integration tests, and `make lint` all pass.


sd2k (commented):
This is a great idea and I'd love to see it in some form.

My only real hesitation is the new grafana/loki/v3 dependency. Since Loki is one big module, taking it as a require pins us to its whole version graph. You can already see it pulling prometheus, otel, alertmanager and a bunch of aws-sdk deps along with it. That coupling tends to be a recurring headache to reconcile whenever anything bumps.

I think we can keep everything great about this approach without the Loki module, reusing the Prometheus parser we already depend on:

  - In LogQL a { only ever opens a stream selector, so a small scanner that tracks strings/backticks/comments can find every {...} span
  reliably.
  - Validate each span with parser.ParseMetricSelector (LogQL selector syntax is Prometheus label-matcher syntax), failing closed on error.
  - Inject by splicing the enforced matchers in before the closing }, preserving the user's formatting.

  That keeps injection into every selector, the fail-closed behavior, and immunity to the line-filter bypass ({app="x"} |=
  "namespace=\"foo\"" - the literal lives inside string state, so it's never mistaken for a selector) with zero new deps.

  Would you be up for reworking it this way? Happy to help sketch the scanner, and your existing tests should carry over almost as-is. 🙂

madaraszg-tulip:
Done — rebased and reworked as suggested, in 2303917. `go.mod`/`go.sum` are now identical to `main`; the semconv bump went with the dep, since it only existed because Loki forced otel v1.44.

Scan for `{...}` spans, validate each with `ParseMetricSelector`, splice the matchers in before the closing `}`. `loki_guardrail.go` already hand-rolls a LogQL scanner for the same reason you gave, so I reused its `indexOutsideStrings`/`closingBrace` instead of adding a second one — it fails open by design, so the fail-closed policy stays in the enforcement layer.

Two notes: a query with no locatable selector is now an explicit error, since `ParseExpr` used to give that for free. And `ParseMetricSelector` skips the "at least one non-empty matcher" check, so purely-negative policies still parse without Loki's `validate=false`.

madaraszg-tulip:
Also picked `--instructions-append` onto this branch (f9fa9e1). Enforcement is invisible to a connecting agent otherwise — a narrowed result looks the same as an empty one, so the LLM can't tell the user some streams are off limits. This gives the operator a way to say so. Kept generic rather than Loki-specific, since there are other things a server owner may want every client to know.

sd2k (approved):
Thanks for handling all the conflicts!

PR #982, opened 2026-07-06
Fix double-encoded timestamp in query_loki_logs streams results

### LINKED ISSUE #803: query_loki_logs returns malformed JSON in content[0].text when log lines contain quotes

MohamedAklamaash:
Fixes #803.

Loki's nanosecond timestamp comes back as a JSON string, so `value[0]` in the streams result is raw bytes like `"1234567890123456789"` including the quote characters. Casting that straight to a Go string kept the quotes in the value, so `content[0].text` ends up with the timestamp double-quoted once the whole entry gets marshaled again.

Unmarshaling `value[0]` into a string instead fixes it, and matches how the vector/matrix branch a bit further down already handles its own timestamp field.

Added a test that builds a streams response with a quoted line value and asserts the parsed timestamp round-trips as a plain string with no leftover escaping.


MohamedAklamaash:
Didn't notice until after opening this that #804 already covers the same bug, and it's already got an approval from @sd2k back in April. Sorry about the overlap, should've checked more carefully first.

Looks like #804 has drifted out of sync with main since then, `queryLokiLogs` got split so the timestamp parsing now lives in `parseLokiQueryResponse`, which is probably why it's showing as conflicting. This one's written against current main so it applies cleanly, but #804 already has the approval and got there first, so whichever is easier to land is fine by me. Happy to close this if you'd rather just rebase #804.

gerboland (approved):
Thank you for this! I will close https://github.com/grafana/mcp-grafana/pull/804 once this lands as the author has not responded in some time

gerboland:
Commit [c7b2f69](https://github.com/grafana/mcp-grafana/pull/982/commits/c7b2f69124abdb6c3820b958afb4e76231093758) is not signed, our tooling won't let us land this unless you sign it & push.

sd2k:
@MohamedAklamaash gentle nudge here, could you please rebase your branch on main and sign your commits so we can merge this? thanks!

MohamedAklamaash:
@sd2k Apologies for the delay. I've just rebased the PR, so it should be good to go. Happy to address any feedback or feel free to merge if it looks good to you.

PR #990, opened 2026-07-10
feat(tools): add compact output format to query_loki_logs

### LINKED ISSUE #203: Make logs output consume less tokens
The current logs output when querying Loki is very verbose. This ends up consuming a lot of context and often hits the size limit.

Image

<details><summary>Details</summary>
<p>

```
      {
        "timestamp": "\"1752066620205756290\"",
        "line": "ts=2025-07-09T13:10:20.183243557Z caller=otel.go:76 level=error msg=OpenTelemetry.ErrorHandler err=\"traces export: failed to send to http://alloy-otlp.alloy-otlp.svc.cluster.local.:4318/v1/traces:
    400 Bad Request (body: \\u0008\\u0003\\u0012\\u001chttp: request body too large)\"",
        "labels": {
          "apps_kubernetes_io_pod_index": "78",
          "cluster": "prod-eu-west-2",
          "container": "ingester",
          "controller_revision_hash": "ingester-zone-c-589567f59f",
          "detected_level": "error",
          "gossip_ring_member": "true",
          "job": "mimir-dedicated-48/ingester-zone-c",
          "name": "ingester-zone-c",
          "namespace": "mimir-dedicated-48",
          "pod": "ingester-zone-c-78",
          "rollout_group": "ingester",
          "service_name": "mimir/ingester",
          "statefulset_kubernetes_io_pod_name": "ingester-zone-c-78",
          "stream": "stderr"
        }
```

</p>
</details> 

An easy way to solve this is to return simple line output without the labels. That should save a lot of repetitive an unhelpful json

### LINKED ISSUE #761: Feature request: pagination / chunking for query_loki_logs results
## Problem

When using `query_loki_logs` in an LLM-powered investigation workflow (Claude Code), broad queries return a single JSON response that can exceed the LLM's output processing budget. For example:

```logql
{namespace="prod"} |= "<encounterId>"
```

This might match hundreds of lines, but the tool returns them all in one response (up to the `limit` cap). Each log line includes full label metadata as JSON, so even 50-100 lines can produce 100KB+ of output.

A human querying Grafana Explore doesn't hit this — they have infinite scroll and visual filtering. But any programmatic consumer (LLM agents, scripts, automation) gets the full payload at once.

## Current workarounds

- Use narrow time ranges
- Apply label filters before line filters (`{app="backend"}` not `{namespace="prod"}`)
- Use `query_loki_stats` pre-flight to estimate result size
- Keep `limit` low (10-20 lines per query)

These work but make exploratory investigation tedious — you end up doing many small queries manually shifting time windows.

## Proposed solutions (any of these would help)

1. **Cursor-based pagination** — return a cursor/token with truncated results that can be passed to get the next page
2. **Time-window chunking** — accept a `chunk_duration` parameter that automatically splits a range query into smaller windows
3. **Line-count pre-check** — before returning results, report the estimated line count so the caller can decide whether to narrow the query
4. **Compact output mode** — option to return only the log line text without full label metadata (labels are already in the stream selector)

## Context

We're building incident investigation workflows where an LLM agent queries Loki to trace issues across services. The current all-or-nothing response model means the agent either gets too little data (low limit) or blows its context budget (high limit). Pagination would make iterative investigation practical.

## Environment

- `mcp-grafana` via `uvx`
- Loki datasource behind Grafana Cloud
- Consumer: Claude Code (Anthropic CLI)

DrFaust92:
## Summary

Fixes #761 (compact output mode).

Broad `query_loki_logs` queries (e.g. `{namespace="prod"} |= "..."`) return every matching line with its full label metadata repeated on each entry, so even 50–100 lines can produce 100KB+ of output and blow an LLM's context budget.

This adds an opt-in `format` parameter:

- `full` (default) — unchanged behaviour and output shape.
- `compact` — log (streams) results are grouped into one `streams` entry per distinct label set, so labels are emitted **once** instead of on every line, and per-line `structuredMetadata`/`parsed` are dropped.

Metric queries are unaffected. `data` remains a JSON array in both modes (empty in compact).

### Compact output shape
```jsonc
{
  "data": [],
  "streams": [
    { "labels": {"app": "backend", "pod": "backend-1"},
      "lines": [ ["<ts>", "line one"], ["<ts>", "line two"] ] }
  ],
  "metadata": { "linesReturned": 2, ... }
}
```

## Why this shape

I went with **grouped-by-stream** rather than a flat `[]string` of `"<ts> <line>"`: on broad multi-stream queries it preserves which stream each line came from (useful for tracing across services) while still emitting each label set only once. Easy to switch to the flat form if preferred — flagged this in #761 for @sd2k.

## Scope

Compact output only. Cursor pagination (no Loki API support) and time-window chunking (silent-drop risk) are intentionally out; the `query_loki_stats` line-count idea can be a separate PR.

## Testing

- `make test-unit` green, incl. new `TestCompactLogEntries` (grouping/order, content-based dedup, empty input).
- `gofmt`, `go vet`, `go build` clean.

Draft pending a quick shape check on the issue.


tools/loki.go
```diff
@@ -736,13 +789,21 @@ func queryLokiLogs(ctx context.Context, args QueryLokiLogsParams) (*QueryLokiLog
 		}
 	}
 
+	// Compact format only applies to log (streams) responses; metric results
+	// carry values rather than log lines and are left untouched. Data is kept
+	// as an empty array so the field's "always a JSON array" contract holds.
+	if args.Format == "compact" && result.ResultType == "streams" {
```
ioanarm: format isn't validated at runtime and any value other than exactly "compact" (e.g. "COMPACT", a typo, trailing space) silently returns full output. Can we reject unknown values (and case-fold/trim)?

tools/loki.go
```diff
@@ -467,10 +470,20 @@ type QueryMetadata struct {
...
 	Metadata *QueryMetadata    `json:"metadata,omitempty"`
 }
 
+// CompactStream groups log lines that share the same label set. The compact
+// output format emits one CompactStream per distinct stream so labels aren't
+// repeated on every line; per-line structured metadata and parser-extracted
+// labels are intentionally omitted to minimise response size.
+type CompactStream struct {
+	Labels map[string]string `json:"labels"`
+	Lines  [][2]string       `json:"lines"` // each element is [timestamp, line]
```
ioanarm: `[][2]string` mirrors Loki's wire format but is less self-describing to an LLM than named timestamp/line fields. Considered a small named struct? Not blocking.

ioanarm (commented):
Thanks for adding this @DrFaust92! 
Some small comments from me and can we also add an integration test as well for this? 

Last one, before this gets merged the commits need to be signed. Sorry for this but it's new policy so we need to comply!

ioanarm:
Thanks for addressing all the feedback! 🙌
The only thing left before we can merge is getting the commits signed, it's a repo policy now! Sorry for this!

sd2k (approved):
LGTM, thank you!

kevelopment (approved):

VtG242:
Hi, thanks for this — the savings look useful.

One concern as a user. The tool description recommends compact for broad queries, so our agent can pick it on its own. We never opted in.

In our Loki, `pod`, `node` and `detected_level` are structured metadata, not stream labels — so compact drops them. The agent then cannot answer "which pod is failing?". Nothing fails and no warning is returned; the answer just quietly loses the thing the question was about. In one test, four different distributor pods collapsed into a single stream with no pod anywhere in the response. And broad queries — the case compact is meant for — cover more pods per label set, so they lose more, not less.

Emitting the metadata once per stream would not work either: some keys are constant within a group (`deployment`, `detected_level`), but `pod` and `node` change line by line.

Could compact split it instead? Constant keys once in the stream header, changing keys on the line. Still much smaller, but nothing lost. I have measurements from our data if that helps.

If that is too much, could compact be disabled server-side, so operators decide instead of the model?

Thanks!

PR #980, opened 2026-07-05
feat(alerting): add MCP tools for built-in Grafana Alertmanager

ayushi-work:
## Summary

Adds three new MCP tools to expose the built-in Grafana Alertmanager, which was previously uncovered by any MCP tool.

### New tools

- **`alertmanager_list_notification_groups`** — wraps `GET /api/alertmanager/grafana/api/v2/alerts/groups` with filtering by `active`/`silenced`/`inhibited` state, label matchers, and receiver name. Returns the same data shown on the `/alerting/groups` page.

- **`alertmanager_list_silences`** — wraps `GET /api/alertmanager/grafana/api/v2/silences` with filtering by label matchers.

- **`alertmanager_create_silence`** — wraps `POST /api/alertmanager/grafana/api/v2/silences` (gated behind `--disable-write`)

### Implementation

- Extends the existing `alertingClient` with `GetAlertGroups`, `GetSilences`, and `CreateSilence` methods using the same auth/transport path as Prometheus rules endpoints
- Uses `github.com/prometheus/alertmanager/api/v2/models` types (`AlertGroup`, `GettableSilence`, `PostableSilence`) — already in the dependency tree
- Registered under the existing "alerting" category — no `main.go` changes needed

Closes #958

Issue #958, opened 2026-06-20
feat(alerting): add tools for built-in Grafana Alertmanager — list active notifications and silences


antigenius0910:
Problem

The /alerting/groups page in Grafana shows active notification groups from the built-in Grafana Alertmanager (/api/alertmanager/grafana/api/v2/alerts/groups). There is currently no MCP tool that exposes this data. (edited) 
[3:19 AM]The existing list_alert_groups tool routes to Grafana OnCall, which is a separate product. Users who don't have OnCall configured get a 404 error, and even when OnCall is present, it doesn't reflect the Alertmanager notification state.

Issue #587 touches adjacent ground (external Alertmanager support) but does not address the built-in Grafana Alertmanager.

Proposed Tools

1. alertmanager_list_notification_groups
Wraps GET /api/alertmanager/grafana/api/v2/alerts/groups

Returns alert groups as seen on the /alerting/groups UI page, grouped by alertname + grafana_folder, with per-alert state (active, suppressed, unprocessed), receiver routing, and label set.

Filters (optional):
- active / suppressed / unprocessed (notification state)
- label matchers (e.g. grafana_folder=Platform Engineering)
- receiver name

2. alertmanager_list_silences
Wraps GET /api/alertmanager/grafana/api/v2/silences

Returns current silences with matchers, creator, comment, and expiry.

3. (optional) alertmanager_create_silence
Wraps POST /api/alertmanager/grafana/api/v2/silences

Guarded by the existing --disable-write flag.

Why This Matters

The /alerting/groups view is the primary operational surface for on-call engineers checking what's firing right now. Without these tools, AI agents have no native way to answer "what alerts are active?" — they must fall back to raw API calls and manage their own auth, which defeats the purpose of the MCP.

Required Permission

alert.instances:read — already granted to standard viewer service accounts.

Related

- #587 — external Alertmanager support (separate scope, but complementary)
- list_alert_groups — OnCall-only, not a substitute for this

ayushi-work:
Hi @antigenius0910 I've taken a look through the codebase, and this seems pretty straightforward to implement. My proposed approach is:
* Add a new `tools/alerting_alertmanager.go` with three MCP tools:
  * `alertmanager_list_notification_groups` → `GET /api/alertmanager/grafana/api/v2/alerts/groups`
  * `alertmanager_list_silences` → `GET /api/alertmanager/grafana/api/v2/silences`
[...]

PR #991, opened 2026-07-11
feat(alerting): add alerting_manage_silences tool (read/write-gated)

MikeHsu0618:
## Summary

Adds an `alerting_manage_silences` tool covering the built-in Grafana Alertmanager silences API (`/api/alertmanager/grafana/api/v2/silences`), completing the alerting family alongside `alerting_manage_rules` and `alerting_manage_routing`.

### Design

Follows the existing unified manage-tool convention (`alerting_manage_rules`):

- **One tool, operation-dispatched**: `list` / `get` / `create` / `update` / `delete` in a single `alerting_manage_silences` tool, so the LLM sees one schema for the whole silence lifecycle.
- **Read/write gated like rules**: when write tools are disabled, a read-only variant (`list`/`get` only) is registered under the same tool name; the full variant is registered otherwise. No `main.go` changes needed.
- `list` supports filtering by label matchers or by `rule_uid` (matches the reserved `__alert_rule_uid__` label).
- `delete` expires the silence (Alertmanager semantics: it stays retrievable with an `expired` state rather than being removed).
- Wire types are reused from `github.com/prometheus/alertmanager/api/v2/models` (`PostableSilence` / `GettableSilence` / `Matcher`). `SilenceMatcherParam` stays as the tool-facing input shape — plain non-pointer fields keep the generated JSON schema readable — and is converted to `models.Matcher` on the wire.
- Reuses the existing `alertingClient` transport path (auth handled by `mcpgrafana.BuildTransport`); adds a small `silenceRequest` helper for non-GET methods since `makeRequest` is GET-only. Response bodies are bounded with the repo's `readResponseBody` / `defaultResponseLimitBytes` helper.

### A note on `update` semantics

Worth knowing when reviewing, because it is counter-intuitive and an earlier revision of this PR described it incorrectly.

Alertmanager only updates a silence **in place** when the posted matchers and `startsAt` match the stored ones. Grafana clamps a `startsAt` in the past to the moment of creation, so the value the caller originally submitted is usually **not** what got stored. Posting it back therefore fails the in-place check: the existing silence is expired and a **new** silence id is returned.

Verified against Grafana 13.0.2:

```
POST /silences  startsAt=03:48:10Z          -> id=9534bf39…, stored startsAt=03:49:10.711Z
POST /silences  id=9534bf39… startsAt=03:48:10Z      -> id=989f0ff5…   (new id, original expired)
POST /silences  id=803c9e69… startsAt=<stored value> -> id=803c9e69…   (id preserved)
```

So: to keep the id stable on `update`, pass back the `starts_at` returned by `get`. This is documented in the tool and the parameter descriptions, and both branches are pinned by integration tests.

### Validation

`create` and `update` reject a window where `ends_at` is not after `starts_at` (Grafana rejects this server-side with a generic 400, so catching it locally saves a round-trip and names the offending fields), and reject a window that has already closed. Expiring a silence early is what `delete` is for, and the `update` error message says so. `create` additionally rejects a non-empty `silence_id`, since a POST carrying an id is treated as an update by the API.

RBAC: silences are governed by `alert.instances:read`, plus `alert.instances:write` for mutations. Rule-scoped silences additionally use the folder-scoped `alert.silences:*` actions.

### Relation to #958 / #980

Resolved: per [#980 (comment)](https://github.com/grafana/mcp-grafana/pull/980#issuecomment-5106888664), this PR owns the full silence lifecycle and #980 has dropped its silence tools, keeping only `alerting_list_notification_groups`. No overlap remains.

### Test plan

- 50 unit subtests across 5 test functions in `tools/alerting_silences_unit_test.go` (operation dispatch, matcher building, rule-UID filtering, read-only gating, window validation, request marshalling, error paths).
- 5 integration test functions (9 subtests) in `tools/alerting_silences_test.go` covering the full lifecycle against a live Grafana: create, get, list (unfiltered / matcher-filtered / `rule_uid`-filtered), update in place, update-that-replaces, delete-expires, the read-only variant, and error paths.
- `go test -tags unit ./...` and `go test -tags integration -run "TestManageRouting|TestManageRules|TestManageSilences" ./tools/` both green against the docker-compose stack (Grafana 13.0.2 + Alertmanager).
- `golangci-lint` v2.11.4 (the CI version) reports 0 issues; `gofmt` and `make lint-jsonschema` clean.

### Docs

`alerting_manage_silences` is added to the tool tables in `README.md` and `docs/sources/reference/mcp-tools-table.md`, and to the write-gated "Alerting Tools" list in `README.md`.


tools/alerting_silences.go
```diff
@@ -0,0 +1,426 @@
...
+
+	resp, err := c.httpClient.Do(req)
+	if err != nil {
+		return fmt.Errorf("failed to execute request to %s: %w", p, err)
+	}
+	defer func() {
+		_ = resp.Body.Close() //nolint:errcheck
+	}()
+
+	bodyBytes, err := io.ReadAll(resp.Body)
```
ioanarm: io.ReadAll(resp.Body) reads the whole body with no size limit, on both the success and error paths. The repo has a helper for exactly this (readResponseBody, 10 MB, errors instead of truncating), and the sibling makeRequest already limits error bodies. Could you use it here?



```suggestion
	bodyBytes, err := readResponseBody(resp.Body, defaultResponseLimitBytes)
if err != nil {
	return fmt.Errorf("failed to read response body from %s: %w", p, err)
}
```

tools/alerting_silences_unit_test.go
```diff
@@ -0,0 +1,433 @@
+package tools
```
ioanarm, line 1: as we are adding a new tool I would like to see an integration test for this as well. Can you add one?

tools/alerting_silences.go
```diff
@@ -0,0 +1,426 @@
...
+type silenceMatcher struct {
+	Name    string `json:"name"`
+	Value   string `json:"value"`
+	IsRegex bool   `json:"isRegex"`
+	IsEqual *bool  `json:"isEqual,omitempty"`
+}
+
+// postableSilence is the request body for creating or updating a silence.
+// A non-empty ID turns a create into an update.
+type postableSilence struct {
```
ioanarm: `alerting_client.go` already imports `github.com/prometheus/alertmanager/api/v2/models`, which has `PostableSilence`/`GettableSilence`/`Matcher`. Any reason not to reuse those here instead of hand-rolling? Keeps us in sync with the API.

tools/alerting_silences.go
```diff
@@ -0,0 +1,426 @@
...
+	for i, m := range p.Matchers {
+		if m.Name == "" {
+			return fmt.Errorf("matcher at index %d: name is required", i)
+		}
+	}
+	if err := requireRFC3339("starts_at", p.StartsAt); err != nil {
+		return err
+	}
+	if err := requireRFC3339("ends_at", p.EndsAt); err != nil {
+		return err
```
ioanarm: We are validating start and end time here but not their relationship. That lets an agent create:
* ends_at before starts_at → a negative/invalid window
* ends_at equal to starts_at → a zero-length silence that does nothing
* both in the past → a silence that's already expired the moment it's created

Maybe we can parse both here and validate ordering (and that ends_at is in the future)? Would give a clearer, cheaper error than a round-trip to Grafana.

tools/alerting_silences.go
```diff
@@ -0,0 +1,426 @@
...
+	manageSilencesReadDescription,
+	manageSilencesRead,
+	mcp.WithTitleAnnotation("Manage alerting silences"),
+	mcp.WithIdempotentHintAnnotation(true),
+	mcp.WithReadOnlyHintAnnotation(true),
+)
+
+// ManageSilences is the write-capable variant (create/update/delete silences),
+// so it is not marked read-only.
+var ManageSilences = mcpgrafana.MustTool(
```
ioanarm: The naming here diverges slightly from the rest. Rules use `ManageRulesRead`/`ManageRulesReadWrite`, whereas this is `ManageSilencesRead`/`ManageSilences`. Consider `ManageSilencesReadWrite` here as well. Purely cosmetic.

ioanarm (commented):
Thanks for adding this! 
A couple of comments from me. Also I left a comment in #980 so we can sync the work across the two PRs. 

Could you also add an `alerting_manage_silences` row to the tools tables `docs/sources/reference/mcp-tools-table.md`? Since it's write-gated, it should also go in the "Alerting Tools" write-gated list in README.md. 

Also double-check the RBAC permission and update this as well. I believe silences use `alert.instances:read` + write scope.

MikeHsu0618:
Thanks for the review @ioanarm! All points addressed in b8a41e3:

- **Docs** — added to both tool tables (`README.md`, `docs/sources/reference/mcp-tools-table.md`) and to the write-gated Alerting list.
- **RBAC** — you were right: `alert.instances:read` + `alert.instances:write` for mutations. (Rule-scoped silences also use the folder-scoped `alert.silences:*` actions; left out of the table for width.)
- **`readResponseBody`** — done, with one deliberate deviation from the suggestion: the helper errors rather than truncates, so I kept the status code in the error when a non-2xx body can't be read. Otherwise an oversized error page loses the one useful fact.
- **Reusing `models`** — wire types are now `models.PostableSilence` / `GettableSilence` / `Matcher`. I kept `SilenceMatcherParam` as the *input* type, since `models.Matcher`'s pointer fields generate a JSON schema that reads poorly for an LLM — so it's plain struct in, `models.Matcher` on the wire. Happy to go all-in if you'd prefer. This also cleared the `S1016` staticcheck failure that was red on CI.
- **Window validation** — `ends_at` must be after `starts_at`, and must not already be in the past; the `update` error points at `delete` for expiring a silence early.
- **Naming** — `ManageSilences` → `ManageSilencesReadWrite`.
- **Integration test** — added `tools/alerting_silences_test.go`: full lifecycle, the three list filters, read-only variant, error paths.

Writing that integration test caught a bug in my own PR description: `update` does **not** keep the same silence id. Grafana clamps a past `starts_at` to the moment of creation, so posting back the value you originally sent fails Alertmanager's in-place-update check and returns a new id; passing back the `starts_at` from `get` preserves it. Both branches are now pinned by tests and the description is corrected.

Also, #980 has dropped its silence tools per your proposal, so no overlap remains.

Would you mind taking another look? The workflows are sitting in `action_required` — could you approve a run? Locally `golangci-lint` v2.11.4 reports 0 issues, and unit + alerting integration tests are green against Grafana 13.0.2.

ioanarm:
For the record: 

## MCP Token Analysis

✅ Passed

Metric | Value
-- | --
Baseline | 16127 tokens
Current | 16874 tokens
Change | +747 (+4.63%)

ioanarm:
Thanks for iterating on the comments @MikeHsu0618! 
I think it's good to go though there is one housekeeping item before we merge. 
Can you please sign your commits?

sd2k (approved):
Looks great, thanks!

PR #993, opened 2026-07-13
fix(prometheus): always convert POST to GET to avoid 422 from Grafana proxy

### LINKED ISSUE #992: Bug Report: run_panel_query fails with 422 due to expr/query parameter mismatch
Here's the issue-ready summary:

---

### Environment
- **mcp-grafana version**: 0.17.1
- **Datasource type**: Prometheus

### Steps to Reproduce

1. Get a dashboard panel with a Prometheus query (e.g., `get_dashboard_summary` + `get_dashboard_panel_queries` for any dashboard with a Prometheus datasource)
2. Call `run_panel_query` with valid `dashboardUid`, `panelIds`, `variables`, `start`, `end`

Example call:
```
run_panel_query(
  dashboardUid: "sSWLKZKVz",
  panelIds: [11],
  variables: {"region": "br", "service": "credit_core_product", ...},
  start: "1783819466606",
  end: "1783835899295"
)
```

### Expected Behavior

The tool resolves the panel's PromQL expression, substitutes dashboard variables, and executes the query via the Grafana DS proxy — returning the time series data for the panel.

### Actual Behavior

The tool correctly resolves the panel ID, variables, and datasource, but fails when executing the actual Prometheus query:

```
executing query: querying Prometheus range: 422: missing `query` arg
```

### Root Cause

`query_prometheus` (called internally by `run_panel_query`) sends the PromQL expression under the parameter name `expr`, but Grafana's datasource proxy API and the Prometheus HTTP API both expect the parameter name `query`.

The fix should be in the `query_prometheus` tool implementation — when calling the Grafana datasource proxy `/api/datasources/proxy/uid/{uid}/api/v1/query_range`, the parameter must be `query`, not `expr`.

### Workaround

Use the `grafana_api_request` tool to manually construct the Prometheus query_range API call:

```
grafana_api_request(
  endpoint: "/api/datasources/proxy/uid/{datasource_uid}/api/v1/query_range?query={url_encoded_promql}&start={unix_timestamp}&end={unix_timestamp}&step=60"
)
```

This works because it bypasses `query_prometheus` and sends parameters directly to the Grafana API.

### Additional observation

The same bug exists when calling `query_prometheus` directly:
```
query_prometheus(
  expr: "sum(rate(metric[1m]))",    // 422: missing `query` arg
  query: "sum(rate(metric[1m]))",   // schema validation rejects this — tool expects `expr`
  ...
)
```

Neither `expr` nor `query` works — the tool's schema demands `expr`, but the API demands `query`. The parameter name in both the tool schema and the underlying HTTP request needs to be `query`.

Levine02:
## Problem

The Prometheus Go client (`org.prometheus.client_golang`) sends query requests as **POST** by default, placing the `query` parameter in the request body. However, **Grafana's datasource proxy** drops the POST body when forwarding to Prometheus, causing:

```
422: missing query arg
```

## Root Cause

In `newPrometheusBackend()`, the `postToGetRoundTripper` was conditionally installed — only when `httpMethod: GET` was explicitly configured in the datasource JSON:

```go
if httpMethod, ok := jsonData["httpMethod"].(string); ok && strings.EqualFold(httpMethod, "GET") {
    rt = &postToGetRoundTripper{underlying: rt}
}
```

Most datasources do **not** have this config set, so the POST→GET conversion was skipped and queries consistently failed with 422.

## Fix

Always install the `postToGetRoundTripper` for all Prometheus-compatible backends. The Prometheus HTTP API accepts both GET and POST identically, so this changes nothing semantically — it only avoids body-handling issues in intermediate proxies.

## Verification

Tested end-to-end against a real Grafana instance:
- `list_datasources` → found the Prometheus datasource UID
- `query_prometheus` with `run_panel_query` → returned 60 time series data points as expected

Fixes #992


gerboland:
Hi, is this the "Grafana datasource proxy" you're referring to? https://grafana.com/docs/grafana/latest/setup-grafana/configure-grafana/proxy/

Prometheus is indeed ok with both GET & POST requests so this makes sense, but I want to verify the complaint before proceeding.

sd2k:
Hey @Levine02, I think this is the right call, if you can sign the CLA and rebase on main with signed commits then we can get this in. Thanks

Levine02:
OKAY&nbsp;



发自我的iPhone


------------------ Original ------------------
From: Ben Sully ***@***.***&gt;
Date: Tue,Aug 25,2026 8:37 PM
To: grafana/mcp-grafana ***@***.***&gt;
Cc: Levine02 ***@***.***&gt;, Mention ***@***.***&gt;
Subject: Re: [grafana/mcp-grafana] fix(prometheus): always convert POST to GET to avoid 422 from Grafana proxy (PR #993)



sd2k left a comment (grafana/mcp-grafana#993)
 
Hey @Levine02, I think this is the right call, if you can sign the CLA and rebase on main with signed commits then we can get this in. Thanks

Levine02:
Hi@sd2k, CLA is signed.

sd2k:
@Levine02 on second thought I think that [bugbot's comment](https://github.com/grafana/mcp-grafana/pull/993#discussion_r3861771862) is valid and worth resolving. Rather than always converting to GET, can we instead keep the conditional check, but also have a round tripper that retries with a GET on 422/500 responses? That seems safer in the face of long query params that may otherwise trigger URL length limits.

Levine02:
Hi@sd2k ,the potential issue has been fixed.

tools/prom_backend.go
```diff
@@ -190,16 +210,94 @@ func (rt *postToGetRoundTripper) RoundTrip(req *http.Request) (*http.Response, e
...
+
+	return rt.underlying.RoundTrip(retryReq)
+}
+
+func shouldRetryPostAsGet(resp *http.Response) bool {
+	if resp == nil {
+		return false
+	}
+
+	return resp.StatusCode == http.StatusUnprocessableEntity || resp.StatusCode == http.StatusInternalServerError
```
sd2k, line 266: The 500 in `shouldRetryPostAsGet` overlaps with `datasourceFallbackTransport.shouldRetry` (in fallback_transport.go), which also retries on 500. Since the fallback transport sits *inside* this wrapper, a genuine Prometheus 500 (e.g. query timeout or OOM) causes up to 4 round-trips:

1. POST → primary (500)
2. POST → fallback (500) — inner fallback transport retry
3. GET → primary (500) — outer POST→GET retry
4. GET → fallback (500) — inner fallback transport retry again

That's a lot of traffic against an already-stressed server, and the GET retry also moves the full PromQL expression into the URL with no length guard — which could hit URI-too-long errors on intermediate proxies for large queries.

Would it make sense to only retry POST→GET on 422 here? That's the specific "missing query arg" signal from the Prometheus API when the proxy drops the body. A 500 from the inner transport has already been through the datasource path fallback, so converting to GET on top of that seems unlikely to help and risks the URL length issue.

tools/prom_backend.go
```diff
@@ -190,16 +210,94 @@ func (rt *postToGetRoundTripper) RoundTrip(req *http.Request) (*http.Response, e
...
+func postToGetRequest(req *http.Request, body []byte) (*http.Request, error) {
 	cloned := req.Clone(req.Context())
 	cloned.Method = http.MethodGet
+	cloned.Body = nil
+	cloned.ContentLength = 0
+	cloned.GetBody = nil
+	cloned.Header.Del("Content-Type")
 
 	// Move URL-encoded form body to query string
 	if req.Body != nil && strings.HasPrefix(req.Header.Get("Content-Type"), "application/x-www-form-urlencoded") {
```
sd2k, line 300: Nit: this checks `req.Body != nil`, but the body bytes have already been buffered and passed in as the `body` parameter. After the POST round-trip through the transport chain, `req.Body`'s state is incidental — the `net/http` RoundTripper contract doesn't guarantee transports preserve it. Using `len(body) > 0` instead would be more robust and semantically clearer, since `body` is the authoritative copy.

tools/prom_backend.go
```diff
@@ -190,16 +210,94 @@ func (rt *postToGetRoundTripper) RoundTrip(req *http.Request) (*http.Response, e
...
+	if resp == nil {
+		return false
+	}
+
+	return resp.StatusCode == http.StatusUnprocessableEntity || resp.StatusCode == http.StatusInternalServerError
+}
+
+// bufferRequestBody reads and restores a request body so the request can be
+// sent once as POST and replayed as GET if needed.
+func bufferRequestBody(req *http.Request) ([]byte, error) {
```
sd2k, line 271: Minor, not a blocker: `fallback_transport.go` (lines 102–112) has a very similar body-buffering pattern, but the two copies diverge subtly — this one sets `GetBody` while the other doesn't. Neither is wrong today (both go through `RoundTrip` directly, so `GetBody` is never called), but two copies that can drift independently is a maintenance risk. Worth considering consolidating into a shared helper in a follow-up.

sd2k (commented):
Thanks for iterating on this, @Levine02 — the layering of datasource-path fallback inside the POST→GET fallback is a nice design and the test coverage is solid. I have a few suggestions below, mostly around tightening the retry behavior.

Issue #594, opened 2026-02-23
Alert rule tools emit bare boolean JSON Schema values, breaking Fireworks AI and other providers


shawnpetersen:
## Description

The `create_alert_rule` and `update_alert_rule` tools emit bare boolean JSON Schema values (e.g., `"model": true`) in their tool definitions. While valid per the JSON Schema specification (`true` means "accept any value"), LLM providers like **Fireworks AI** reject these with a 500 `internal_server_error`.

## Reproduction

1. Connect mcp-grafana to any LLM gateway that routes to Fireworks AI
2. Include `create_alert_rule` or `update_alert_rule` in the tool list
3. Send a chat completion request with 26+ tools (including the alert rule tools)
4. Fireworks returns: `{"error":{"object":"error","type":"internal_server_error","code":"invalid_request_error","message":"server had an error while processing your request, please retry again after a brief wait"}}`

With 25 tools (excluding alert rule tools): succeeds.
With just `create_alert_rule` alone: fails.

## Root Cause

The alert rule params include Go `interface{}` fields (likely `QueryData` or the alert rule model). When reflected to JSON Schema via the `jsonschema` library, `interface{}` produces a bare `true`. Fireworks AI's schema parser does not handle bare boolean schemas.

## Affected Tools

- `create_alert_rule` — has `"model": true` in schema
- `update_alert_rule` — same

## Expected Behavior

Bare boolean schemas should be replaced with equivalent object schemas:
- `true` → `{"type": "object"}` (or more specific if the type is known)
- `false` → `{"not": {}}`

## Workaround

We applied a `sanitizeSchema()` function in our MCP adapter layer that recursively converts bare booleans before passing schemas to the LLM provider. See: androidStern-personal/openclaw-mcp-adapter#9

## Environment

- mcp-grafana: Docker image `grafana/mcp-grafana` (latest as of 2026-02-23)
- Grafana: 11.x with Editor-role service account
- LLM Provider: Fireworks AI (Kimi K2.5 model)
- Gateway: LiteLLM v1.81.13 + OpenClaw v2026.2.22

materemias:
## Confirmed: Also affects GitHub Copilot + Gemini 3 Flash (Preview)

[...]

PR #997, opened 2026-07-16
tools: reject unknown tool arguments instead of silently dropping them

aleks-p:
An MCP client that sends a misspelled or unknown argument name currently gets a successful response computed from defaults: the argument decode path (`json.Unmarshal`) silently drops unknown keys, and generated schemas don't declare `additionalProperties: false`, so nothing between the caller and the handler catches the mistake. For example, `query_pyroscope` with `start_rfc3339` (instead of `start_rfc_3339`) silently queries the default now-1h window instead of the requested one — the response looks healthy but answers a different question. LLM agents hit this frequently, plausibly because time-param names are inconsistent across the tool surface (`startRfc3339` in the Loki tools, `startTime` in the Prometheus tools, `start_rfc_3339` in the Pyroscope tools), so a blended spelling like `start_rfc3339` is a natural failure mode.

This change makes `ConvertTool` reject unknown top-level argument keys. The error is returned as a tool result (`IsError`) rather than a protocol error, so LLM callers see it and can self-correct:

```
unknown argument "start_rfc3339"; valid arguments: data_source_uid, end_rfc_3339, group_by, matchers, max_node_depth, profile_type, query_type, start_rfc_3339, step
```

Generated schemas now also declare `additionalProperties: false`, so schema-validating clients and providers with strict function calling catch the mistake before it reaches the server.

Notes:
- Key matching is case-insensitive, mirroring `encoding/json`, so calls that previously populated fields keep working; only keys that were being silently dropped now fail.
- Only top-level keys are checked — nested object schemas are unchanged.
- `validateNoBooleanSchemas` gets a carve-out for `additionalProperties: false`: it's valid JSON Schema and universally supported (OpenAI structured outputs requires it). The bare-boolean concern from #594 was `true` schemas coming from `interface{}` fields, and those are still rejected.
- Verified against a live Grafana with a Pyroscope datasource: correct calls are unchanged; typo'd/unknown args on `query_pyroscope`, `list_pyroscope_label_names`, `list_datasources`, and `query_prometheus` are all rejected with the message above.

### DIFF
--- a/tools.go
+++ b/tools.go
@@ -7,7 +7,6 @@ import (
 	"fmt"
 	"net/http"
 	"reflect"
-	"sort"
 	"strings"
 
 	"github.com/invopop/jsonschema"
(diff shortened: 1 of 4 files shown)


tools.go
```diff
@@ -254,6 +338,17 @@ func ConvertTool[T any, R any](name, description string, toolHandler ToolHandler
...
+		// Reject unknown argument keys instead of silently dropping them: a typo'd
+		// optional argument (e.g. "start_rfc3339" for "start_rfc_3339") would
+		// otherwise leave the field zero-valued and the tool would answer a
+		// different question than the caller asked. The error is returned as a
+		// tool result (not a protocol error) so LLM callers can see it and retry.
+		if unknown := unknownArguments(argBytes, argType); len(unknown) > 0 {
+			msg := unknownArgumentsError(unknown, knownJSONFieldNames(argType))
+			span.SetStatus(codes.Error, "unknown arguments")
+			return mcp.NewToolResultError(msg), nil
+		}
```
kolesnikovae, line 273: I think we should use the generated json schema, which contains the list of declared properties.

`unknownArguments` is called from the handler closure in `ConvertTool`. Immediately after that closure is declared, we build the schema needed for the returned `mcp.Tool`:

```go
	jsonSchema := createJSONSchemaFromHandler(toolHandler)
	properties := make(map[string]any, jsonSchema.Properties.Len())
	for pair := jsonSchema.Properties.Oldest(); pair != nil; pair = pair.Next() {
		properties[pair.Key] = pair.Value
	}
```

Could we move this block before the handler closure and capture `properties` there as the set of known arguments? Then could then validate `request.Params.Arguments` directly:

```go
unknownArguments(request.Params.Arguments, properties)
```

This would avoid independently deriving the property names. The reflection-based helpers in this PR don't cover all of json field-selection rules in Go.

kolesnikovae (commented):
Thanks for the PR, Aleks! LGTM but I suggest that we try using the generated json schema to infer known args – please check my comment

tools.go
aleks-p: Thank you Anton, can you take a look at 7d612c9?

gerboland (approved):
Tested and this works as described, thank you

PR #1002, opened 2026-07-21
chore: drop dead toolToDatasources map from proxied tool set

Rbillon59:
Follow-up to #1001 (stacks on it; please merge after #1001).

`proxiedToolSet.toolToDatasources` (and its `builtProxiedTools` counterpart) is populated on every build but never read anywhere in production: proxied tool calls route by the `datasourceUid` parameter through `GetProxiedClient`, not through this map. It was already write-only before #1001; this removes the field from both structs and the per-tool bookkeeping that filled it.

No behavior change. Unit tests pass with `-race`; lint clean.

> Note: opened against `main` for review convenience, but the branch is stacked on #1001 — its diff will simplify to just the dead-field removal once #1001 merges. If you prefer, I can rebase this onto `main` after #1001 lands.

### DIFF
--- a/cmd/mcp-grafana/main.go
+++ b/cmd/mcp-grafana/main.go
@@ -330,6 +330,11 @@ func newServer(transport string, dt disabledTools, obs *observability.Observabil
 	// unregister sessions from the SDK's internal session map.
 	sm.SetMCPServer(s)
 
+	// Give the SessionManager a reference to the ToolManager so tearing down a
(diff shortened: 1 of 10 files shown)


Rbillon59:
Closing: as a stacked PR against main this renders the whole of #1001 on top of the 14-line cleanup, which is misleading. 

Those changes will pe proposed in another PR if the 1001 changes are merged

PR #1004, opened 2026-07-21
fix(observability): allow enabling OTLP tracing without log export

madaraszg-tulip:
## Summary

Enabling OpenTelemetry **tracing** currently forces OTLP **log export** on as well. Trace export is gated solely on the generic `OTEL_EXPORTER_OTLP_ENDPOINT`, and when log export is on the server routes *all* of its `slog` output through the otelslog bridge to the collector. An operator who just wants traces in their collector unavoidably ships every application log line there too — with no way to get one signal without the other.

## What changes

Trace export now honors the signal-specific `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` (falling back to the generic endpoint), via a new `OTLPTracesEndpoint()` helper that mirrors the existing `OTLPLogsEndpoint()`. Setting only the traces-specific variable enables tracing **without** log export.

**Log export behaviour is unchanged** — it was already independently controllable via `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT`. The only new capability here is on the tracing side.

Resulting behaviour by environment:

- `OTEL_EXPORTER_OTLP_ENDPOINT` (generic) — traces **on**, logs **on** (unchanged)
- `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` only — traces **on**, logs **off** (new; previously enabled nothing)
- `OTEL_EXPORTER_OTLP_LOGS_ENDPOINT` only — traces **off**, logs **on** (unchanged; already worked)

This also fixes a latent gap: before, setting only `OTEL_EXPORTER_OTLP_TRACES_ENDPOINT` enabled nothing, because the gate checked only the generic variable even though the exporter itself honors the traces-specific one.

Also adds a `TracerProvider()` accessor and a symmetric `"OTLP trace export configured"` startup log line, matching the existing log-export announcement.

## Testing

- Unit tests for `OTLPTracesEndpoint()` (precedence, fallback, empty) plus a regression guard asserting a traces-only endpoint brings up the tracer provider while `LoggerProvider()` stays nil.
- Verified end-to-end against all four env configurations (traces-only, generic, logs-only, none) — each announces the correct signal(s) on startup.
- `go build`, `go vet`, and the observability unit tests pass.

README and the observability docs page are updated to describe the per-signal endpoints.

### DIFF
--- a/observability/logs.go
+++ b/observability/logs.go
@@ -24,6 +24,17 @@ func OTLPLogsEndpoint() string {
 	return os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")
 }
 
+// OTLPTracesEndpoint returns the resolved OTLP traces endpoint, preferring the
+// signal-specific OTEL_EXPORTER_OTLP_TRACES_ENDPOINT over the generic
+// OTEL_EXPORTER_OTLP_ENDPOINT. Returns "" when neither is set, which is the
+// signal that OTLP trace export is disabled.
+func OTLPTracesEndpoint() string {
+	if v := os.Getenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT"); v != "" {
+		return v
+	}
+	return os.Getenv("OTEL_EXPORTER_OTLP_ENDPOINT")
+}
+
 type fanoutHandler struct {
 	children []slog.Handler
 }
--- a/observability/observability.go
+++ b/observability/observability.go
@@ -134,12 +134,11 @@ func newSlowRequestLogger(level slog.Level) *slog.Logger {
 // a global MeterProvider. The otelhttp instrumentation will automatically
 // use this provider for HTTP metrics.
 //
-// Tracing configuration is handled via standard OTEL_* environment variables
-// (e.g., OTEL_EXPORTER_OTLP_ENDPOINT, OTEL_TRACES_SAMPLER).
-//
-// Log export is enabled when OTEL_EXPORTER_OTLP_ENDPOINT or
-// OTEL_EXPORTER_OTLP_LOGS_ENDPOINT is set; use LoggerProvider() to retrieve
-// the provider for wiring into an slog.Handler (e.g., via the otelslog bridge).
+// Trace export is gated on OTLPTracesEndpoint and log export on
+// OTLPLogsEndpoint; use LoggerProvider() to retrieve the log provider for
+// wiring into an slog.Handler (e.g., via the otelslog bridge). Other tracing
+// behaviour is configured via standard OTEL_* environment variables
+// (e.g., OTEL_TRACES_SAMPLER).
 func Setup(cfg Config) (_ *Observability, err error) {
 	// Ensure OTel SDK internal errors (async export failures, queue drops, etc.)
 	// surface through slog instead of the stdlib log package where operators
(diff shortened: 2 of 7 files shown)


observability/logs.go
```diff
@@ -24,6 +24,23 @@ func OTLPLogsEndpoint() string {
...
+// signal-specific OTEL_EXPORTER_OTLP_TRACES_ENDPOINT over the generic
+// OTEL_EXPORTER_OTLP_ENDPOINT. Returns "" when neither is set, which is the
+// signal that OTLP trace export is disabled.
+//
+// Resolving the traces-specific variable independently lets operators enable
+// tracing WITHOUT also enabling log export: setting only
+// OTEL_EXPORTER_OTLP_TRACES_ENDPOINT turns on traces and leaves
+// OTLPLogsEndpoint empty (logs off), whereas the generic
+// OTEL_EXPORTER_OTLP_ENDPOINT enables both signals.
+func OTLPTracesEndpoint() string {
```
gerboland, line 31: These last 5 lines say the same as the 4 lines above that, please remove.

observability/observability.go
```diff
@@ -134,12 +134,19 @@ func newSlowRequestLogger(level slog.Level) *slog.Logger {
...
 // Log export is enabled when OTEL_EXPORTER_OTLP_ENDPOINT or
-// OTEL_EXPORTER_OTLP_LOGS_ENDPOINT is set; use LoggerProvider() to retrieve
-// the provider for wiring into an slog.Handler (e.g., via the otelslog bridge).
+// OTEL_EXPORTER_OTLP_LOGS_ENDPOINT is set (see OTLPLogsEndpoint); use
+// LoggerProvider() to retrieve the provider for wiring into an slog.Handler
+// (e.g., via the otelslog bridge).
+//
+// The two signals resolve their endpoints independently, so setting only the
+// signal-specific OTEL_EXPORTER_OTLP_TRACES_ENDPOINT enables tracing without
+// turning on log export; the generic OTEL_EXPORTER_OTLP_ENDPOINT enables both.
```
gerboland: This is mentioned too many times in this diff now. Please shorten these comments, they actualyl confuse me with their verbosity.

observability/observability_test.go
gerboland, line 106: Logical to add
```
t.Setenv("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT", "")
```
in this test and verify tracer provider is nil.

observability/observability_test.go
```diff
@@ -51,6 +51,26 @@ func TestSetup(t *testing.T) {
...
+		require.NoError(t, err)
+		require.NotNil(t, obs)
+		assert.NotNil(t, obs.tracerProvider, "tracer provider should be set when traces endpoint is configured")
+		assert.Nil(t, obs.LoggerProvider(), "log export must stay off when only the traces endpoint is set")
+
+		shutdownCtx, cancel := context.WithTimeout(context.Background(), 2*time.Second)
+		defer cancel()
+		assert.NoError(t, obs.Shutdown(shutdownCtx))
+	})
+
```
gerboland, line 71: I'd appreciate a test that ensures both logs & traces are exported if just `OTEL_EXPORTER_OTLP_ENDPOINT` is set.

gerboland (changes requested):
Thank you for the contribution! I've some small code niggles inline, but I'm happy with what you want to do here.

madaraszg-tulip:
Thanks for the review! Addressed in the latest push: trimmed the verbose/duplicated comments across the diff, and updated the tests — the logs-only case now asserts the tracer provider stays nil, and added a case verifying the generic `OTEL_EXPORTER_OTLP_ENDPOINT` enables both signals.

gerboland (approved):
Thank you!

PR #1006, opened 2026-07-22
fix: restrict prom backend to known Prometheus-compatible datasource types

millescasGoto:
## Description

Add `tempo` as an explicitly rejected datasource type in `backendForDatasource`.

The existing `default` branch already routed any unrecognised datasource type to the native Prometheus backend, preserving forward compatibility with Prometheus-compatible community plugins. This change adds a single explicit `case "tempo"` before that default to short-circuit with a clear error.

Tempo is a tracing datasource that uses TraceQL and does not implement the Prometheus querying API. Without this guard, an LLM client passing a Tempo datasource UID to `list_prometheus_metric_names` would make a real HTTP request to the Tempo proxy, receive a 404, and record a `STATUS_CODE_ERROR` span 

## Changes

- `tools/prom_backend.go`: Add `case "tempo"` returning a clear unsupported-type error before the existing `default` Prometheus fallback
- `tools/prom_backend_test.go`: Add `TestBackendForDatasource_RejectsTempoDatasource` and `TestBackendForDatasource_AcceptsPrometheusDatasource` to lock in routing behaviour and prevent regression

## Testing

```bash
go test -tags=unit ./tools -run 'TestBackendForDatasource'


StephaneBrillant (commented):
This is a breaking change. I think you should list the datasources type that you want to generate an error instead of changing the default behavior.

ioanarm (approved):

PR #882, opened 2026-05-14
Add MCP App panel viewer for get_panel_image

adela-almasan:
Adds MCP Apps support so `get_panel_image` renders the panel image inline via an HTML viewer in hosts that support it (Claude, Cursor, etc.). Hosts without MCP Apps support continue to receive the base64 PNG as before.

### Changes
- `WithUIResource()` tool option to attach `_meta.ui.resourceUri` to tool definitions
- Panel viewer MCP App using `@modelcontextprotocol/ext-apps`
- `RegisterAppResources()` to serve HTML via the MCP Resources API
- Built HTML checked in for `//go:embed` (no Node.js needed for `go build`)

Screenshot 2026-05-14 at 10 19 14 AM

---

> [!NOTE]
> **Low Risk**
> Read-only rendering tool change with additive response content and optional MCP App metadata; behavior is covered by updated unit and integration tests.
> 
> **Overview**
> Adds **MCP Apps** support so `get_panel_image` can render panel snapshots inline in capable hosts, while still returning the PNG for everyone else.
> 
> The tool now returns **two** content parts: the existing base64 PNG plus a **Grafana deeplink** (stored dashboards via `/d/<uid>`, provisioning previews via `/dashboard/provisioning/...`, with optional `viewPanel`). `WithUIResource` sets `_meta.ui.resourceUri` on the tool; `RegisterAppResources` serves the embedded panel-viewer HTML over the MCP Resources API. Startup registers those resources after tools are loaded.
> 
> Build wiring adds a **`build-ui`** Makefile target (npm build for `ui/*`) as a dependency of **`make build`**, with `.gitignore` updates for UI artifacts while keeping checked-in `ui/*/dist/` for `go:embed`. Unit and integration tests assert the second text payload and tool `_meta.ui`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 046fcd3b857d1f385552103d73efd6f7ab7f2fdf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #825, opened 2026-04-30
feat: add MCP Apps support with interactive time series viewer

sd2k:
## Summary

- Adds support for the [MCP Apps extension](https://modelcontextprotocol.io/extensions/apps/overview), which lets tools return interactive HTML UIs that render inline in MCP hosts (Claude, VS Code Copilot, Goose, etc.)
- Implements an interactive time series chart using [uPlot](https://github.com/leeoniya/uPlot) (same lib Grafana uses) for `query_prometheus` and `query_prometheus_histogram` results
- Hosts that don't support MCP Apps ignore the metadata entirely — the text result still works as before

Example of how mcpjam looks:

image

### What's included

**Go server infrastructure:**
- `WithUIResource()` tool option to attach `_meta.ui.resourceUri` to any tool definition
- `RegisterAppResources()` to serve HTML resources via the MCP Resources API
- `//go:embed` to bundle pre-built HTML into the Go binary
- `--dev-cors` flag for local development with the [MCP Apps basic-host](https://github.com/modelcontextprotocol/ext-apps/tree/main/examples/basic-host)

**Time series UI app (`ui/timeseries/`):**
- Parses Prometheus matrix (range) and vector (instant) query results
- Interactive uPlot chart with zoom/pan, series toggle via legend click, refresh
- Dark/light theme support via `prefers-color-scheme`
- Dev mode with mock data when opened directly in a browser (`npm run dev`)
- Grafana-inspired color palette

**Build integration:**
- `make build-ui` target builds all UI apps under `ui/*/`
- `make build` now depends on `build-ui`
- Built HTML is checked in (for `//go:embed` — no Node.js needed to `go build`)

### How it works

1. When a host lists tools, `query_prometheus` includes `_meta.ui.resourceUri: "ui://mcp-grafana/timeseries.html"`
2. Host fetches the resource → gets the single-file HTML bundle with uPlot inlined
3. HTML renders in a sandboxed iframe, receives the tool result via `ontoolresult`, parses the data, draws the chart
4. "Refresh" button calls `app.callServerTool()` to re-execute the query without a new prompt

### Future work

- Log viewer app for `query_loki_logs`
- Generic panel viewer for `run_panel_query`
- App-only tools (e.g. `poll-metrics` with `visibility: ["app"]`) for live updates

## Test plan

- [x] `go vet ./...` passes
- [x] `go test ./...` passes
- [x] `make build` succeeds (builds UI then Go binary)
[...]

PR #1008, opened 2026-07-23
Add insight-cell: a generic MCP App render surface + multi-app framework

thanos-karachalios:
## What

Builds on the MCP Apps foundation (#825 / #882) in two ways:

1. **Generalizes the single-app plumbing into a multi-app framework.** `RegisterAppResources` currently serves one hardcoded app (the panel viewer from #882). This replaces the hardcode with a small `UIApp` registry that it loops over, so adding an MCP App is one registry entry + one `//go:embed` line rather than new bespoke code.
2. **Adds the "insight cell"** — a generic Grafana render surface — as the second app, wired to a new `render_insight_cell` tool.

Fully wired into the Go server, embedded via `//go:embed`, and covered by tests. (A standalone Node/TS prototype of the same render contract is in #1003, for iterating on the renderer; this PR is the self-contained Go integration and does not depend on it.)

## The insight cell

One render contract + one renderer that dispatches on `renderHint.type`, so adding a visualization is a new branch, not a new app:

- **Core panels:** `timeseries`, `stat`, `bar`, `table`, `logs`, `trace`
- **Synthesis views:** `worklist` (ranked triage), `rca` (root cause → evidence), `timeline` (change correlation), `cost` (cardinality/spend drivers)
- **Guardrailed write:** `rulediff` renders a proposed alert-rule change as a before/after diff; applying it routes to the existing write-gated `update_alert_rule` — no new write path.

`render_insight_cell` is a **render substrate**: the agent gathers data with the existing query tools (`query_prometheus`, `query_loki_logs`, `list_alert_rules`, `get_annotations`, Sift, …), does the analysis, and passes the results here. The tool does **not** query datasources or fabricate data. Results ride three channels so they degrade across hosts — a text verdict (fallback), an embedded `application/json` resource block, and `structuredContent` — plus `_meta` with `ui.resourceUri` and a `grafana.insightCell/v0` trust profile (attestation, provenance, reasoning). Hosts without MCP Apps support still get the text + JSON.

## Changes

- `ui_apps.go` — `UIApp` registry + `InsightCellResourceURI`; `RegisterAppResources` loops the registry
- `ui_embed.go` — embed the insight-cell bundle
- `ui/insight-cell/` — the single-file MCP App (Vite + `vite-plugin-singlefile`), built `dist/mcp-app.html` committed for `//go:embed` (`make build-ui` already globs `ui/*/`)
- `tools/insight_cell.go` — the `render_insight_cell` tool + render-contract types
- `cmd/mcp-grafana/main.go` — registers the `insight-cell` category (`--disable-insight-cell`)
- Tests: `tools/insight_cell_test.go`, `ui_apps_test.go`
- Docs: `docs/mcp-apps.md` (framework + render contract + how to add an app) and a README MCP Apps section

## Test plan

- [x] `make build-ui && make build` — compiles with the embedded bundle
- [x] `go test . ./tools/ ./cmd/...` — pass
- [x] `make lint-jsonschema`, `gofmt` — clean
- [x] stdio smoke: `tools/call render_insight_cell` returns all three channels + `_meta.ui.resourceUri` + `grafana.insightCell/v0`; `resources/read ui://mcp-grafana/insight-cell.html` serves the embedded bundle
- [x] Verify in a host with MCP Apps support — Claude Desktop against a live Grafana Cloud stack + the MCP Apps basic-host ([screenshots](https://github.com/grafana/mcp-grafana/pull/1008#issuecomment-5094058590))


thanos-karachalios:
b4fb2a3 addresses the two latest Bugbot findings:

- **Share menu prototype stubs**: the dropdown now only offers actions that work in this surface — Copy link (only when the cell carries a real link action; the fake placeholder URL fallback is gone) and Download. Cross-user share needs a server-side store + retrieval tool, so it's deliberately out of scope for this PR; the full persist → link → reopen-in-another-session loop is demonstrated in the standalone prototype (#1003, `share_cell` / `open_shared_cell` + a read-only web fallback). Bringing that to the Go server is a follow-up, since the interesting question there is where the cell store lives in a stateless mcp-grafana deployment.
- **Rule-diff applied state**: `render_insight_cell` gains an `applied` input so the diff can be re-rendered in the APPLIED state after a successful `update_alert_rule`; covered by a test asserting the marshaled payload.

cmd/mcp-grafana/main.go
```diff
@@ -106,7 +107,7 @@ type grafanaConfig struct {
 }
 
 func (dt *disabledTools) addFlags() {
-	flag.StringVar(&dt.enabledTools, "enabled-tools", "search,datasource,incident,prometheus,loki,alerting,dashboard,folder,oncall,asserts,sift,pyroscope,navigation,proxied,annotations,rendering,snapshot,plugin,api,config,provisioning", "A comma separated list of tools enabled for this server. Can be overwritten entirely or by disabling specific components, e.g. --disable-search.")
+	flag.StringVar(&dt.enabledTools, "enabled-tools", "search,datasource,incident,prometheus,loki,alerting,dashboard,folder,oncall,asserts,sift,pyroscope,navigation,proxied,annotations,rendering,snapshot,plugin,api,config,provisioning,insight-cell", "A comma separated list of tools enabled for this server. Can be overwritten entirely or by disabling specific components, e.g. --disable-search.")
```
ioanarm: `insight-cell` is added in the default enabled-tools list here, so this tool loads by default. There is a token-budget concern the repo optimizes hard against. Can we make this opt-in? Drop it from here and keep the --disable-insight-cell / registration plumbing?

.claude/skills/author-mcp-app/SKILL.md
```diff
@@ -0,0 +1,94 @@
+---
```
ioanarm: I don't think this should be committed. Maybe if we want it as a skill we can ship it separately.

.claude/skills/test-insight-cell/SKILL.md
```diff
@@ -0,0 +1,82 @@
+---
```
ioanarm: Can we make this an integration test instead?

docs/mcp-apps.md
```diff
@@ -0,0 +1,101 @@
...
+2. **an embedded `application/json` resource block** in `content[]` — kept by hosts that drop
+   `structuredContent`. Claude Desktop converts it to a text block, so the app also scans text
+   blocks for JSON starting with `{`.
+3. **`content[0]` text** — a human-readable verdict; the fallback when there is no app at all.
+
+## The insight-cell render contract
+
+The insight cell is a generic surface: **one contract + one renderer that dispatches on
+`renderHint.type`**. Adding a visualization is a new branch in the renderer, not a new app. The Go
+types in `tools/insight_cell.go` mirror `contrib/insight-cell/src/schema.ts` (the source of truth
```
ioanarm: I think this reference here belongs to another PR

docs/mcp-apps.md
```diff
@@ -0,0 +1,101 @@
...
+
+Render types:
+
+- **Core panels:** `timeseries`, `stat`, `bullet` (a value vs a target/SLO with qualitative
+  bands), `bar`, `table` (read `frames`), `logs` (read `logs`), `trace` (read `trace`).
+- **Synthesis views:** `worklist` (ranked triage), `rca` (root cause → evidence), `timeline`
+  (change correlation), `cost` (cardinality/spend drivers).
+- **Guardrailed write:** `rulediff` renders a proposed alert-rule change as a before/after diff.
+  Applying it routes to the existing **write-gated** `update_alert_rule` tool — the cell proposes,
+  it does not write.
```
ioanarm: I think the cell can call write tools as well. There is a `ReadOnlyHint: true` annotation for the tool but this is not true. Comment below.

tools/insight_cell.go
```diff
@@ -0,0 +1,591 @@
...
+		"actionable findings for alert triage / deprecations), 'rca' (root-cause investigation), 'rulediff' (a proposed "+
+		"alert-rule fix shown as a before/after diff), 'timeline' (change correlation) or 'cost' (cardinality/cost drivers) "+
+		"— with a verdict, attestation, provenance and drill actions. This renders data you already have: first use the "+
+		"query tools (query_prometheus, query_loki_logs, list_alert_rules, get_annotations, Sift, ...), do the analysis, "+
+		"then pass the results here as 'frames' (chart types) or the matching payload (items/findings/changes/events/drivers), "+
+		"along with a one-line 'verdict' and a 2-4 sentence 'insight'. It does not query datasources itself. Hosts without "+
+		"MCP Apps support still get the text verdict and the JSON payload.",
+	renderInsightCell,
+	mcp.WithTitleAnnotation("Render a Grafana insight cell"),
+	mcp.WithReadOnlyHintAnnotation(true),
```
ioanarm, line 661: This says `WithReadOnlyHintAnnotation(true)` but the actions[] channel let the rendered UI invoke write tools (update_alert_rule) via callServerTool in `ui/insight-cell/src/mcp-app.ts:83-88`. A read-only-hinted tool that can initiate writes is a wrong signal to hosts that gate on this hint. 

I would keep it and make the cell structurally read-only. Remove the kind:"tool" action branch in the UI, strip tool-kind actions server-side in buildInsightCell, and turn rulediff. Every write should go through the agent calling the MCP tools, never from inside the cell.

ui/insight-cell/src/mcp-app.ts
```diff
@@ -0,0 +1,185 @@
...
+    btn.disabled = true;
+    btn.textContent = "Working…";
+    if (a.kind === "refresh") {
+      const res = await app.callServerTool({ name: "render_insight_cell", arguments: specFrom(cell) });
+      if (!applyResult(res)) restoreButton(btn, a);
+    } else if (a.kind === "tool" && a.tool) {
+      const res = await app.callServerTool({ name: a.tool, arguments: a.args ?? {} });
+      // Not every tool returns a cell (e.g. a write like update_alert_rule), so
+      // if there's nothing to re-render, re-enable the control instead of
+      // leaving it stuck on "Working…".
```
ioanarm: This can call any tool by name `(callServerTool({ name: a.tool }))` including write tools like `update_alert_rule`. This is the single line that makes the ReadOnlyHint false.

ui/insight-cell/src/format.ts
```diff
@@ -0,0 +1,103 @@
+// Backend B (real): format values through Grafana's own field-config pipeline.
+//
+// Uses @grafana/data's `getDisplayProcessor` — the per-field core of
+// `applyFieldOverrides` — so units, decimals, thresholds, and value mappings
+// come out exactly as a Grafana panel renders them (e.g. "1.46 MiB", "23.4 ms",
+// threshold colors like #F2495C). Runs in the browser/iframe bundle only
+// (@grafana/data doesn't import cleanly in Node). This is IC-9: in real
+// mcp-grafana the same code lives in the ui/ app.
```
ioanarm: I think the IC-9 is stale?

tools/insight_cell.go
```diff
@@ -0,0 +1,591 @@
...
+	DataMode    string        `json:"dataMode"` // mock | live
+	ResultInfo  string        `json:"resultInfo,omitempty"`
+}
+
+type icTimeRange struct {
+	From string `json:"from"`
+	To   string `json:"to"`
+}
+
+// InsightCell is the full render contract emitted as structuredContent.
```
ioanarm: nit: `insightCell` not `InsightCell`.

tools/insight_cell.go
```diff
@@ -0,0 +1,591 @@
...
+type icMeta struct {
+	Question    string        `json:"question"`
+	Verdict     string        `json:"verdict"`
+	Confidence  string        `json:"confidence"`
+	TimeRange   icTimeRange   `json:"timeRange"`
+	Attestation icAttestation `json:"attestation"`
+	Provenance  icProvenance  `json:"provenance"`
+	Query       []icQueryRef  `json:"query"`
+	DataMode    string        `json:"dataMode"` // mock | live
+	ResultInfo  string        `json:"resultInfo,omitempty"`
```
ioanarm: is this used anywhere?

ioanarm (changes requested):
that's really nice @thanos-karachalios, I like the direction this is taking. 
I left some comments if you can have a look at some point. This is a more complicated approach than the "just render it" from #825 but the grafana formatting introduced here is quite valuable imo.

Also there are some lint issues I think and open bugbot comments maybe worth to look at!

cmd/mcp-grafana/main.go
thanos-karachalios: Done in 55597cf — `insight-cell` is now opt-in: removed from the default `--enabled-tools` list, registration + `--disable-insight-cell` plumbing kept (flag help now says how to enable). This should also fix the failing `analyze` job, which is the token-threshold check measuring the default toolset.

.claude/skills/author-mcp-app/SKILL.md
thanos-karachalios: Removed in 55597cf. docs/mcp-apps.md now carries the short how-to-add-an-app steps inline; the skill can ship separately (e.g. in our ai-kit) if we want it.

.claude/skills/test-insight-cell/SKILL.md
thanos-karachalios: Done in 55597cf — replaced with `tools/insight_cell_integration_test.go` (`//go:build integration`). It drives the tool through an in-process MCP client (no external services needed): every panel type × the three output channels, the trust `_meta`, live/mock dataMode, the read-only guarantees (ReadOnlyHint + action stripping), and `resources/read` of the embedded bundle. The skill is deleted.

docs/mcp-apps.md
thanos-karachalios: Right — that referenced the contrib prototype from #1003. Fixed in 55597cf: the source of truth for field names is now `ui/insight-cell/src/schema.ts` (in this PR), and the remaining contrib references in code comments are gone too.

docs/mcp-apps.md
thanos-karachalios: You were right — the actions channel made the hint false. Fixed in 55597cf as you suggested (details on the code threads below); this paragraph is rewritten to describe the actual guarantee: action kinds limited to link/refresh/ask, tool-kind actions stripped server-side, and applying a rulediff goes agent-side via the write-gated `alerting_manage_rules` (which also fixes the stale `update_alert_rule` references — that tool no longer exists).

tools/insight_cell.go
thanos-karachalios: Implemented exactly this in 55597cf: the `kind:"tool"` branch is removed from the UI (the only `callServerTool` left is `render_insight_cell` for refresh), `buildInsightCell` strips any action that isn't link/refresh/ask via `sanitizeActions` (cell-level and worklist items), and the `tool`/`args` fields are deleted from the action contract on both sides. rulediff Apply is now an `ask` action: the cell hands text back to the agent, the agent applies via `alerting_manage_rules` and re-renders with `applied=true`. The ReadOnlyHint stays and is now structurally true — the integration test asserts both the hint and the stripping.

ui/insight-cell/src/mcp-app.ts
thanos-karachalios: Gone in 55597cf — the generic tool-call branch is deleted. The UI can now only call `render_insight_cell` (hardcoded, for refresh), which is this same read-only tool, so the hint holds.

ui/insight-cell/src/format.ts
thanos-karachalios: Yes, stale prototype-era numbering — removed in 55597cf.

tools/insight_cell.go
thanos-karachalios: Fixed in 55597cf (the type was already unexported; the doc comment said `InsightCell`).

tools/insight_cell.go
thanos-karachalios: Good catch — the Go side never populated it (only the UI would render it if present), so it was dead on the wire. Removed from the contract in 55597cf, both `tools/insight_cell.go` and `ui/insight-cell/src/schema.ts`.

tools/insight_cell.go
```diff
@@ -0,0 +1,578 @@
...
+			from = fromISO
+		}
+		if to == "" {
+			to = toISO
+		}
+		cell.Timeline = &icTimelinePayload{From: from, To: to, Events: args.Events}
+	}
+	if len(args.Drivers) > 0 || args.CostTotal != nil {
+		cell.Cost = &icCostPayload{Total: args.CostTotal, Drivers: orEmpty(args.Drivers), Headroom: args.Headroom}
+	}
```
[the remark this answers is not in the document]
thanos-karachalios: This was fixed in b4fb2a3 (the renderer shows the headline total even with no drivers — `costPanel` bails only when there's neither a total nor drivers), and the unit test covers a total-only cell.

tools/insight_cell.go
```diff
@@ -0,0 +1,591 @@
...
+	To     string          `json:"to,omitempty" jsonschema:"description=For panel='timeline': ISO window end."`
+
+	// cost
+	Drivers   []icCostDriver  `json:"drivers,omitempty" jsonschema:"description=For panel='cost': ranked cost/cardinality drivers."`
+	CostTotal *icCostTotal    `json:"costTotal,omitempty" jsonschema:"description=For panel='cost': the headline total\\, e.g. {label:'1.2M active series'\\, value:1200000}."`
+	Headroom  *icCostHeadroom `json:"headroom,omitempty" jsonschema:"description=For panel='cost': the headroom trade-off making 'adding headroom costs money' explicit."`
+
+	// chrome
+	Callout *icCallout `json:"callout,omitempty" jsonschema:"description=Optional callout banner shown above the panel (tone: warn|crit|info)."`
+	Actions []icAction `json:"actions,omitempty" jsonschema:"description=Drill/link/refresh/ask actions shown under the panel. kind 'tool' calls another MCP tool (e.g. update_alert_rule to apply a rulediff); 'link' opens a URL; 'refresh' re-runs render_insight_cell; 'ask' sends text back to the agent."`
```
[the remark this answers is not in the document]
thanos-karachalios: Fixed in 55597cf — all references now point at `alerting_manage_rules` (operations `list`/`update`): the rulediff schema descriptions, the tool description, and docs/mcp-apps.md. More fundamentally, actions can no longer name tools at all: the cell is structurally read-only and applying a rulediff goes through the agent calling `alerting_manage_rules` itself.

ui/insight-cell/src/mcp-app.ts
```diff
@@ -0,0 +1,185 @@
...
+      if (!applyResult(res)) restoreButton(btn, a);
```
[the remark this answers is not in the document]
thanos-karachalios: Fixed in 55597cf — Apply is now an `ask` action: the agent performs the write and re-renders the cell with `applied=true` (the input added in b4fb2a3), so the badge flips to Applied on the re-render. `specFrom` also carries `applied` through refresh so the state survives.

ui/insight-cell/src/mcp-app.ts
```diff
@@ -0,0 +1,185 @@
...
+    return true;
+  }
+  return false;
+}
+
+/** Re-enable an action button after a call that didn't re-render the cell. */
+function restoreButton(btn: HTMLButtonElement, a: InsightCellAction) {
+  btn.disabled = false;
+  btn.textContent = a.label;
+}
```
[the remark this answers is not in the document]
thanos-karachalios: Fixed in 55597cf — `restoreButton` now captures and restores the button's original `innerHTML` instead of writing a text label over the SVG glyph.

thanos-karachalios:
Thanks for the review @ioanarm! All comments addressed in 55597cf:

- **Structurally read-only cell** (your prescription, implemented as-is): `kind:"tool"` removed from the UI and the contract, non-link/refresh/ask actions stripped server-side in `buildInsightCell`, and rulediff Apply reworked to go through the agent (`ask` action → agent calls the write-gated `alerting_manage_rules` → re-renders with `applied=true`). `ReadOnlyHint` stays and is now true; the only server tool the UI can call is `render_insight_cell` itself.
- **Opt-in**: `insight-cell` dropped from the default `--enabled-tools` (plumbing kept) — this also brings the `analyze` token check back under threshold.
- **Skills removed**; the test skill is now a real integration test (`tools/insight_cell_integration_test.go`, in-process MCP client, no external services: all panel types × three channels, trust `_meta`, read-only assertions, `resources/read` of the bundle).
- **Lint + Bugbot**: the staticcheck failures are fixed, and the open Bugbot findings are addressed (stale `update_alert_rule`/`list_alert_rules` → `alerting_manage_rules`, applied-state now reachable, icon buttons keep their glyphs; the cost-total one was already fixed in b4fb2a3).
- Plus the small ones: contrib/IC-9 stale refs, the `insightCell` doc comment, and the never-populated `resultInfo` field removed from the contract.

Replied on each thread with specifics.

thanos-karachalios:
## Host-render verification ✅ (last test-plan item)

Verified 55597cf in two MCP Apps hosts:

**Claude Desktop, live Grafana Cloud stack** (stdio, `--enabled-tools "…,insight-cell"`) — the agent gathered data with the query tools and rendered via `render_insight_cell`:

*timeseries — node_load1, provenance panel open (`live: true`, real query + datasource):*


*rca — root cause + checks + evidence for a firing alert:*


*timeline — alert-state cluster correlation over 24h:*


**MCP Apps basic-host** (the #825 reference host, streamable-http) — including the reworked read-only rulediff where Apply is an `ask` pill routed through the agent:



One note from the basic-host run: browser hosts can't complete CORS preflight against `--transport streamable-http` because mcp-go's default `Access-Control-Allow-Headers` omits `MCP-Protocol-Version` (which the official TS SDK sends). Pre-existing on main (affects the #882 panel viewer too) — I used a local header shim for the test; will file separately rather than widen this PR.

---

**More panel types, same Claude Desktop + live stack setup** — logs (Loki), bullet, and bar:

*logs — Alloy logs with level filter chips and a data-quality callout:*


*bullet — value vs target with threshold bands:*


*bar — ranked CPU modes, provenance open (`live: true`):*

thanos-karachalios:
Rebased onto current main (0049a0e) — clean replay, no functional changes. Conflicts were limited to `cmd/mcp-grafana/main.go` where the new Agent Observability category (#944) landed on the same lines as the insight-cell wiring; resolved as a union (both categories registered, insight-cell still opt-in). Re-verified locally after picking up #997's strict argument validation: build, vet, lint, full unit suite, and the insight-cell integration tests all pass. Head is now 0e57665 (was 55597cf).

tools/insight_cell.go
```diff
@@ -0,0 +1,616 @@
...
+		title = defaultTitleForPanel(args.Panel)
+	}
+
+	now := time.Now().UTC()
+	rangeHours := 1
+	if args.RangeHours != nil && *args.RangeHours > 0 {
+		rangeHours = *args.RangeHours
+	}
+	fromISO := now.Add(-time.Duration(rangeHours) * time.Hour).Format(time.RFC3339)
+	toISO := now.Format(time.RFC3339)
```
ioanarm: On a refresh the UI replays the same data and this handler stamps a brand new now/toISO. So old data end up claiming being fresh. This expands to `ui/insight-cell/src/mcp-app.ts`

tools/insight_cell.go
```diff
@@ -0,0 +1,616 @@
...
+
+	// Provenance / attestation. "live" means the agent recorded a query + datasource
+	// it read from a real datasource; otherwise this is representative/sample content.
+	live := args.Query != "" && args.DatasourceUID != ""
```
ioanarm, line 411: `live` and `provenance.author: "Grafana MCP"` are self-asserted. There is no verification that the frames came from a specific query/datasource. The model gathered and shaped the data but we are implying that users should trust it as the Author is Grafana MCP. So we should either rename it (maybe `author` → `renderedBy` ) so we stop implying that the data is verified or maybe split provenance to what is actually verifiable (a reference to the query-tool result vs what the agent did)

tools/insight_cell.go
```diff
@@ -0,0 +1,616 @@
...
+			mcp.EmbeddedResource{
+				Type: "resource",
+				Resource: mcp.TextResourceContents{
+					URI:      "insightcell://payload.json",
+					MIMEType: "application/json",
+					Text:     string(payload),
+				},
+			},
+		},
+		StructuredContent: cell,
```
ioanarm, line 643: The full cell is emitted in both structuredContent and the embedded application/json block, and the same data is already in the tool-call args, so a large frames/table payload can travel through context ~2–3×. Maybe it's worth documenting the token cost here if it's a deliberate choice and maybe consider if this is needed.

tools/insight_cell.go
```diff
@@ -0,0 +1,616 @@
...
+	cell, err := buildInsightCell(ctx, args)
+	if err != nil {
+		return nil, err
+	}
+	return insightCellResult(cell)
+}
+
+func buildInsightCell(ctx context.Context, args RenderInsightCellParams) (*insightCell, error) {
+	if args.Panel == "" {
+		return nil, fmt.Errorf("panel is required")
```
ioanarm, line 365: `panel` is an enum in the schema, but it's not enforced at call time, so only the empty string is guarded here.

ioanarm (commented):
Thanks for addressing the comments @thanos-karachalios! 
I have some more! :)

dafydd-t:
Some usability feedback from clicking around on this, locally:

- This visualization dropdown sometimes hides beneath the UI element below 
Screenshot 2026-07-30 at 11 07 04

---

- I click "download", i see this "downloaded cell json" message, but i have nothing in my downloads dir. Tell me where its downloading to? (if it works)
Screenshot 2026-07-30 at 11 07 26

---

- The `copy` button functionality gets blocked?

Screenshot 2026-07-30 at 11 08 48

tools/insight_cell.go
thanos-karachalios: Good catch — fixed in 671a140. `attestation.asOf` and the recorded time range now anchor on a new optional `dataAsOf` arg instead of `time.Now()`, and the cell's refresh (`specFrom` in `mcp-app.ts`) passes the original `attestation.asOf` back — so a replay keeps the timestamp the data was gathered at instead of claiming to be fresh. Malformed `dataAsOf` is rejected. Covered by `TestRenderInsightCellPreservesDataAsOf`.

tools/insight_cell.go
thanos-karachalios: Agreed, went with the rename: `provenance.author` → `provenance.renderedBy` (Go contract, `schema.ts`, `render.ts`, docs) — it now names who packaged/rendered the cell, not an implied data author. The UI provenance rows also label the self-asserted fields as declared ("datasource (declared)", "live (declared)"), and the contract comments state explicitly that the server records the agent's declaration without verifying that the frames came from that query/datasource. The declared query ref stays in `meta.query` as the re-runnable reference.

tools/insight_cell.go
thanos-karachalios: It's deliberate, and you're right it deserved documenting — done at `insightCellResult` in 671a140. Host support is inconsistent: some hosts drop `structuredContent` entirely, and Claude Desktop rewrites the resource block into a text block, so the app needs the fallback chain in `extractCell`. The comment now spells out the ~2–3× token cost as the price of rendering on hosts we don't control, with a note to revisit dropping the embedded block once `structuredContent` forwarding is reliable across major hosts.

tools/insight_cell.go
thanos-karachalios: Fixed — `panel` is now validated at call time against the enum list (the jsonschema enum is advisory to the model only), with an error naming the valid values. Covered by `TestRenderInsightCellRejectsUnknownPanel`.

dafydd-t:
There's a lot of UI in this PR, and i'd want to get someone from dataviz to weigh in to make sure it feels like "grafana" to their expert eyes. 

With that in mind, i think we could split it along these lines, after some back and forth with claude. It splits it into 3 PRs, but ultimately, its PR 1 for the go changes, and then PR 2 and 3 for UI components of the app. WDYT? Left the detail so your agent can reason about it :)

### Suggested split

**1. generalise the app registry + add the render contract and tool. (_Daf: the tool would NOT actually render UI at this point, so we might want to make that clear in the interim tool description)_** — Go only, ~950 lines, mostly tests

- `ui_apps.go` (`UIApp`, the `appResources` registry with just the panel-viewer entry, the loop), `ui_apps_test.go`
- `tools/insight_cell.go` — contract types, params, `buildInsightCell`, `sanitizeActions`, `orEmpty`, `insightCellResult`, and the tool definition **without** `WithUIResource` and without the result-level `_meta.ui`
- both test files, plus the four-line `main.go` wiring
- `panel` enum covers everything except `rulediff`

no behaviour change to existing tools, and no Node needed to review it. the value here is that all the subtle Go lands in one place a Go reviewer can actually see: the trust profile, the `dataMode` derivation, the `*float64`-vs-`omitempty` handling, the nil-slice normalisation, the action whitelist. right now those are spread through 8k lines of mostly frontend.

worth rewording the description in this PR — it currently opens "Render a Grafana result ... as an interactive 'insight cell' in the chat", which isn't true until the app exists, and it's agent-facing.

**2. add the insight-cell MCP App** — ~7k lines

- the whole `ui/insight-cell/` tree, all panels except `rulediff`, plus the committed `dist/mcp-app.html`
- `ui_embed.go`, the `InsightCellResourceURI` const + second registry entry
- `WithUIResource` and the result `_meta.ui` added to the tool, description reworded
- `docs/mcp-apps.md` + README

feature works end to end here. keeping the bundle rebuild as its own final commit makes it easy to skip on review.

**3. add the `rulediff` panel** — ~320 lines

Go types/params, the enum value, `ruleDiffPanel`, the ~7 lines in `specFrom`, the `applied` state, tests, and the read-only argument. separating this is the one I'd push hardest for: it's the only piece where being wrong means an agent mutates production alerting, and the `sanitizeActions` + `ask`-path + `ReadOnlyHint` reasoning deserves its own review — plausibly from alerting owners, who otherwise have to read a 2.3k-line frontend PR to get to it. it also detaches cleanly, nothing else references it.

I did consider splitting the synthesis views out too, but I don't think it earns it — they're only ~250 lines of `render.ts` between them, and splitting means three separate rebuilds of `dist/mcp-app.html`, three edits to the `panel` enum, and a window where the hand-maintained `insight_cell.go` ↔ `schema.ts` mirror can drift. better together.

two constraints that fix the 1/2 boundary, for the descriptions:

- `go:embed` reads at compile time, so registry entry + embed directive + `dist/mcp-app.html` have to land together
- both references to `InsightCellResourceURI` need to wait for PR 2 — the tool option *and* the one in `insightCellResult` — otherwise PR 1 advertises a resource a host can't read
- PR 1 shipping a tool that renders nowhere is only safe because `insight-cell` isn't in the default `--enabled-tools`. that opt-in gate is doing real work for the incremental landing

### on `dataMode`

separate from the split, and the thing I'd most want settled before the UI commits to the badge semantics: `live` is a claim, not a verification — `args.Query != "" && args.DatasourceUID != ""`. nothing cross-checks `frames` against `query`, so an agent can pass fabricated numbers with a real PromQL string and a real datasource UID and the cell renders them under a green "live" badge with a provenance block naming the stack.

I don't think that's fixable inside a render substrate, so it may just need to be explicit — but it's currently spread across the Go, the TS and the docs, and it's the claim a reader is most likely to over-trust.

### two small things from running it

- `docs/mcp-apps.md` and the PR body both describe dispatch on `renderHint.type`. that's right for the output contract, but the tool *input* is a flat `panel` string and `renderHint` is assembled server-side — so following the doc to call the tool gets you `unknown arguments "renderHint"`
- the doc's suggested dev loop is `--dev-cors` on streamable-http, but there's no such flag in the binary. closest equivalent today is `--allowed-origins '*'`

dafydd-t:
I have been thining about what @ioanarm said on the call yesterday that I now understand: There is no guarantee that the dataframes passed to render_insight_cell are "real". the agent could just make them up. I know we categorise data as "live" (which is not guaranteed either) or not, but this breaks my mental model that I can 100% rely on the data I see in these graphs as real data - especially as the working model here is "agent analysis, backed up by graphs" - that sentence isn't guarantee.

Maybe we could label these panels as something that represents "data was supplied by the agent" (i.e. it should be real, but we cannot know for sure), and "synthesized" (for the new synthesis views really)

thinking out loud: In the future, we could change the backend logic to have a guarantee of real data, for more work: If we pass in the query & datasource UID, the backend could make the query itself, and render the data it receives back from grafana. There would be a fair bit of branching etc to support every datasource there, maybe, or pick the right visualization type. The other negative is that we're making the datasource query twice to achieve this guarantee of accuracy, and as well as being an extra load, the queries could show different data depending on how far apart the queries are made (assuming its made after the agent analysis, this could actually be a good few seconds, or even minutes) - this could be solved by requiring from/to params though.

thanos-karachalios:
Thanks @dafydd-t — really useful, both the click-around findings and the split analysis. Plan below.

### On your three usability finds

All real, and all the same family: the host's iframe sandbox.

- **dropdown clipped** — an ancestor `overflow: hidden` clips the absolutely-positioned menu. straight CSS fix.
- **download** — to answer "tell me where it's downloading to": nowhere. it's a blob-anchor click, and a sandbox without `allow-downloads` silently blocks it while the toast says "downloaded" unconditionally. the toast is lying. i'll make it honest and fall back to showing the JSON for manual save when the download can't happen.
- **copy blocked** — `navigator.clipboard` needs a permission the sandbox doesn't grant. adding a select-the-text fallback + honest failure message.

Bugbot also caught two refresh-replay omissions on the latest commit (`confidence` and `rangeHours` both silently reset to defaults on refresh) — fixing those in the same pass.

### On the split: agreed, let's do it

Your boundaries check out on the details I verified — `go:embed` forces registry entry + embed directive + `dist/` to land together, both `InsightCellResourceURI` references have to wait for PR 2, and the opt-in `--enabled-tools` gate is what makes a renderer-less PR 1 safe. Sequence:

1. land the fixes above (plus your two doc catches) on this branch first, so they carry into the split cleanly
2. **PR 1** — go-only substrate, no `WithUIResource`, no rulediff, and the tool description reworded for the interim (you're right that "render ... in the chat" is agent-facing and untrue until the app exists)
3. **PR 2** — the app, stacked on PR 1. i'll pull someone in from dataviz for the "does this feel like grafana" pass
4. **PR 3** — rulediff on its own, alerting owners tagged. agreed this is the one that earns a separate review — it's the only piece where being wrong mutates production alerting

Also agree on *not* splitting the synthesis views — the schema.ts/insight_cell.go mirror drifting across three bundle rebuilds is a worse failure mode than a slightly bigger PR 2.

Both doc catches are right btw: `--dev-cors` doesn't exist (that's `--allowed-origins '*'` — inherited from older text), and the `renderHint.type` framing describes the *output* contract while the tool input is a flat `panel` string. fixing both.

### On dataMode / trust

You and @ioanarm have converged on the real issue, and i agree it should be settled before the UI commits to badge semantics. The `renderedBy`/"declared" rename from her round was a step, but your labels are the better end state: **"agent-supplied"** for frames-backed panels (should be real, we can't know for sure) and **"synthesized"** for the synthesis views. i'll make that the badge + contract wording as part of PR 1/2, and put the caveat in one place in the docs instead of spread across the go, the TS and the docs like it is now.

The server-side re-query idea i'd keep out of this series — it's the right long-term guarantee, but it changes what the tool *is* (render substrate → query-and-render), and as you say it only really works with explicit from/to params, plus the double-query load and drift questions. i'll file it as a follow-up issue so the thread doesn't get lost.

thanos-karachalios:
Phase 1 landed in b6de65e — the fixes that make sense regardless of the split:

- **dropdown clipping** — menus now portal to `<body>` and position fixed from the trigger rect. Turned out to be two bugs stacked: `.viz-box`'s `overflow: hidden` clipped the menu, and `position: fixed` alone didn't fix it because `.viz-toolbar` animates with a `transform`, which quietly makes it the containing block for fixed descendants. Verified placement headlessly with a screenshot harness.
- **download** — no more toast claiming a download that may never have happened (the sandbox swallows the blob-anchor click without throwing, so success is unknowable from inside the iframe). It now attempts the download and always shows the JSON in a dismissable panel for manual save.
- **copy** — falls back to `execCommand("copy")` when the async clipboard API is denied (works under a user gesture in most sandboxes), and to the same selectable-text panel when both are blocked.
- **bugbot's two replay omissions** — refresh now carries `confidence` and derives `rangeHours` back from the recorded `timeRange`, so neither silently resets to defaults.
- **your two doc catches** — `--dev-cors` → `--allowed-origins '*'` (with a dev-only warning), and a note that `renderHint` is output-side while the tool input is a flat `panel` string.

Next up: cutting the split along your lines — PR 1 (go-only substrate) first.

thanos-karachalios:
Split done, along @dafydd-t's lines:

- **#1037 — 1/3 substrate** (Go only, ~1.2k lines, mostly tests) — ready for review. No `WithUIResource`, no result `_meta.ui`, no rulediff; interim tool description says the result isn't rendered visually yet. Both `InsightCellResourceURI` references wait for 2/3, and the opt-in `--enabled-tools` gate is what makes the renderer-less landing safe — both constraints held.
- **#1038 — 2/3 the MCP App** (draft until 1/3 merges) — the `ui/insight-cell/` tree, the wiring, docs; bundle rebuild as its own final commit. This is the one that needs the dataviz "feels like Grafana" pass.
- **#1039 — 3/3 rulediff** (draft) — the alerting-adjacent panel with the read-only argument laid out for alerting owners.

Also settled before the UI commits to badge semantics, as discussed: `dataMode` is now **`"agent-supplied"` | `"synthesized"`** (was `live`/`mock`) — the label never claims verification the substrate can't perform, and the caveat lives in one place in `docs/mcp-apps.md`.

Everything here (including your usability fixes and Bugbot's replay omissions) is carried into the split branches. Converting this PR to draft and keeping it for the discussion history — happy to close it once 1/3 is moving.

PR #1012, opened 2026-07-24
fix(observability): respect OTEL_LOGS_EXPORTER=none to disable OTLP log export

millescasGoto:
When `OTEL_EXPORTER_OTLP_ENDPOINT` is set for traces, `OTLPLogsEndpoint()` was falling back to it and silently enabling OTLP log export — even when the backend does not support LogsService — causing:

```bash
  otel sdk error: rpc error: code = Unimplemented desc = unknown service
  opentelemetry.proto.collector.logs.v1.LogsService
```

`OTLPLogsEndpoint() `now short-circuits to "" when `OTEL_LOGS_EXPORTER=none` (case-insensitive), preventing the logs provider from being created regardless of endpoint configuration.

## Testing

Two new unit tests added:
- `TestOTLPLogsEndpoint_DisabledByExporterNone`
- `TestOTLPLogsEndpoint_DisabledByExporterNoneCaseInsensitive`

All existing tests continue to pass.

## Docs

README updated with the `OTEL_LOGS_EXPORTER=none` escape hatch and a note about the
error it prevents.

### DIFF
--- a/README.md
+++ b/README.md
@@ -525,7 +525,7 @@ volumes:
 Surrounding whitespace (including a trailing newline) is trimmed from the file contents. If both `GRAFANA_SERVICE_ACCOUNT_TOKEN` and `GRAFANA_SERVICE_ACCOUNT_TOKEN_FILE` are set, the inline token takes precedence.
 
 ### Multi-Organization Support
- 
+
 You can specify which organization to interact with using either:
 
 - **Environment variable:** Set `GRAFANA_ORG_ID` to the numeric organization ID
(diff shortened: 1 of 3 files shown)


gerboland (approved):
Thank you!

gerboland:
Hi @millescasGoto thank you for this, but we'll need you to sign our CLA before it can be merged

millescasGoto:
> Hi @millescasGoto thank you for this, but we'll need you to sign our CLA before it can be merged

Thank you, @gerboland, and the sign-in process was done

Issue #524, opened 2026-01-30
403 Forbidden on Prometheus/Loki query tools (datasource resources proxy) while same token works for /api/ds/query and from curl


terryng:
Summary
Prometheus query tools (query_prometheus, and related metadata tools) return 403 Forbidden when mcp-grafana calls the datasource proxy endpoint. The same Grafana service account token and the same URL return 200 when called via curl or wget from a pod in the same cluster. Only requests made by the mcp-grafana process get 403.

Environment
mcp-grafana: Running in Kubernetes, authenticated via GRAFANA_SERVICE_ACCOUNT_TOKEN
Grafana: Behind HTTPS; service account has datasource read/query permissions; dashboards and datasource list work (200)
Endpoint used by MCP: POST /api/datasources/uid/<uid>/resources/api/v1/query
Steps to reproduce
Deploy mcp-grafana with a Grafana service account token that has datasources:read and datasources:query (or equivalent).
Use an MCP client to call query_prometheus (or list_prometheus_label_names, etc.) with a valid datasource UID.
Observe: Grafana returns 403 Forbidden for the request to /api/datasources/uid/<uid>/resources/api/v1/query.
So the failure is specific to how mcp-grafana performs the request (or how Grafana treats that client), not the token or the URL.

What we verified
list_datasources, search_dashboards, and other non-proxy tools work.
Same token and same proxy URL from a laptop (curl) and from a pod in the same cluster (wget) return 200 or 422 (e.g. bad query params), never 403.
Only the mcp-grafana process receives 403 for the resources proxy call.
Possible cause
We suspect this may be related to https://github.com/grafana/grafana/issues/51163 (403 on provisioned datasources via the resources proxy). The MCP server might be hitting a code path or sending headers that trigger the same (or a related) permission check in Grafana.

kimberlyyp:
We hit this same issue on AWS Managed Grafana (workspace `g-284ec0e9c7.grafana-workspace.us-west-2.amazonaws.com`). Adding some findings that narrow down the root cause.

## Root cause: POST not supported on `/resources` proxy in AWS Managed Grafana

The regression was introduced in PR #483 (fixing #467 for Azure Managed Grafana), which changed the Prometheus client base URL from:

```go
// v0.8.2 - works on AWS Managed Grafana
url := fmt.Sprintf("%s/api/datasources/proxy/uid/%s", ...)

// v0.9.0 - broken on AWS Managed Grafana
[...]

PR #1015, opened 2026-07-27
feat: resolve datasources via /api/frontend/settings when the metadata API is forbidden

jinwonjeong-creator:
## What

When the datasource metadata API (`GET /api/datasources*`) is not accessible to the token in use, resolve datasources from `GET /api/frontend/settings` instead, and route the datasource proxy through a path the deployment actually supports.

## Why

Every datasource-backed tool currently fails at the `getDatasourceByUID` pre-flight when the token cannot read datasource metadata:

- **Grafana before 9.0 (7.x/8.x** — still common in large self-hosted/managed fleets): `/api/datasources*` requires the Org Admin role, so viewer/editor tokens get a 403. Worse, the uid-based proxy routes (`/api/datasources/uid/{uid}/resources`, `/api/datasources/proxy/uid/{uid}`) were only added in 9.0 — on 7.x/8.x requests fall through to the numeric `:id` route and return a 500 `"Unable to load datasource meta data"` — so even an admin token wouldn't help.
- **Newer Grafana with RBAC**: minimally-scoped service accounts may lack `datasources:read` while still being permitted to query data.

`GET /api/frontend/settings` is served to any authenticated role and includes each datasource's numeric id, uid, name, type, and non-secret `jsonData` — enough to resolve the datasource and pick a proxy route that exists.

This extends the deployment-compatibility approach of the existing `datasourceFallbackTransport` (#524, Azure vs AWS managed Grafana route differences) to deployments where *neither* uid-based route exists.

## How

- New `tools/datasources_fallback.go`: fetches `/api/frontend/settings`, resolves by uid (or name), records uid→numeric-id in a `sync.Map` keyed by Grafana URL + uid.
- `getDatasourceByUID` / `getDatasourceByName` / `listDatasources`: fall back to frontend settings when the metadata call fails (404 keeps its existing friendly error, no fallback).
- `datasourceProxyPaths` keeps its original signature and behavior — the modern codepath is untouched. Instead, each datasource client constructor (`newPrometheusBackend`, `newLokiClient`, `newGraphiteClient`, `newVictoriaLogsBackend`) checks `fallbackProxyBase` first: when the datasource was resolved through the fallback, the client's **primary base is the numeric-id proxy route** (`/api/datasources/proxy/{id}`) — the uid-based routes don't exist pre-9.0 and answer with version-dependent statuses (verified live: **400 "id is invalid" on 8.5**, 500 on 7.5), so they can't be the primary. The uid-based proxy route remains only as the transport-level fallback for the opposite case: a newer Grafana with an RBAC-restricted token, where the numeric (legacy id) routes are deprecated since 9.0 and [disabled by default since 13.0](https://grafana.com/whats-new/2026-04-14-deprecated-data-source-apis-disabled-by-default/) (404 → retried).
- `datasourceFallbackTransport` now also retries on **404**, since a route missing from a deployment surfaces as 404 there (e.g. the disabled legacy id routes on 13+) rather than the 403/500 already handled. A 404 from the datasource itself just costs one duplicate request, and the fallback path is still only cached on a 2xx.

**No behavior change when the metadata API is accessible** — the fallback only engages on failure, and the happy path makes no extra requests.

## Testing

- **Behavior-matrix unit tests** (`tools/datasources_fallback_behavior_unit_test.go`): real queries driven through `backendForDatasource` → the Prometheus client (and the native Loki client) against a fake Grafana that answers routes per deployment generation — `grafana-7.5` (uid routes → 500), `grafana-8.5` (uid routes → **400 "id is invalid"**, the case from the review), `grafana-13-rbac` (numeric routes → 404), and `modern` (asserts uid routes only and that `/api/frontend/settings` is **never** requested). The 8.5 test reproduces the review failure and was red before this fix.
- Unit tests for the resolution layer: fallback by uid/name, pseudo-datasource filtering + default flag, org/credential cache scoping, uid/name key-space separation, 404-retry in the fallback transport. Full unit suite green.
- **Integration test on real Grafana 8.5** (`tools/legacy_grafana_integration_test.go` + new `grafana-legacy-8` docker-compose service, built from the setup in the review): creates a basic-auth editor user at runtime and exercises `list_datasources`, `get_datasource`, `query_prometheus` (instant `time()`), and `list_prometheus_label_names` end to end. Green against `grafana/grafana:8.5.27`.
- Verified live against a Grafana **7.5.17** instance (n3r, Naver's internal container platform): `query_prometheus` (instant + range), `list_datasources`, `get_datasource` all work; dashboard tools unaffected.


jinwonjeong-creator:
Addressed the Bugbot finding in 9243ab3: the fallback cache key now includes the org id and a hash of the credential material (Grafana tokens are org-scoped), so tenants sharing a Grafana URL in multi-tenant HTTP mode can no longer observe each other's resolved numeric ids. Added `TestFallbackProxyIDs_ScopedPerOrgAndCredential` covering both the different-org and different-token cases. Also signed the commit.

jinwonjeong-creator:
Both new findings addressed in the latest push:

- **Credential scope omits BasicAuth**: the cache-key hash now covers every auth mechanism on `GrafanaConfig` — API key, on-behalf-of tokens, **BasicAuth**, and **ExtraHeaders** (sorted deterministically). Test extended with basic-auth and extra-headers tenants.
- **UID and name cache collide**: uid and name entries now live in separate, kind-prefixed key spaces, and `fallbackProxyBase` consults the uid entry first — mirroring the resolution precedence of `fallbackDatasourceByUID`. Added `TestFallbackProxyIDs_UIDAndNameDoNotCollide` reproducing the name==uid shadowing scenario.

gerboland:
An interesting problem & approach, please allow me time to test it

gerboland:
Hi, ok I had time to test. It is slightly hilarious to me that as an editor user `/api/datasources` is permission denied, but you do indeed have sufficient datasource info from `/api/frontend/settings` to work with. I can confirm `list_datasources` tool works well, and passing a valid `uid` or `name` to `get_datasource` is good too. 

I had problems querying a datasource however, I used this:

```
npx @modelcontextprotocol/inspector --cli http://localhost:8008/mcp --transport http  --method tools/call --tool-name query_prometheus --tool-arg datasourceUid=prometheus endTime="now-1m" expr="time()" queryType="instant"
```

which failed:

```json
{
  "content": [
    {
      "type": "text",
      "text": "querying Prometheus instant: Skip: do not know how to skip: 0, error found in #0 byte of ...||..., bigger context ...||..."
    }
  ],
  "isError": true
}
{"error":{"code":"tool_is_error","message":"Tool 'query_prometheus' returned isError:true."}}
```


Additionally other prometheus calls fail: `list_prometheus_label_names` `list_prometheus_label_values` `list_prometheus_metric_metadata` `list_prometheus_metric_names` `query_prometheus_histogram`
as they all try to use `/api/datasources/proxy/uid/prometheus/*`

Using plain IDs might be safer choice, but it makes this a bigger change again :( I'm not opposed to that, as long as the fallback pattern is clearly separated from the more modern codepath, and it is well tested.




For your context I'm testing locally. To do that I added this to my docker-compose:

```
  grafana-legacy-8:
    image: grafana/grafana:8.5.27@sha256:b43abed8a2b8c5a4055c3729d77ae11c9646c4b9feefb5c136b7c67dcdad21c0
    environment:
      GF_AUTH_ANONYMOUS_ENABLED: "false"
      GF_LOG_LEVEL: debug
      GF_SERVER_ROUTER_LOGGING: "true"
    ports:
      - 3003:3000/tcp
    volumes:
      - ./testdata/provisioning:/etc/grafana/provisioning
      - ./testdata/dashboards:/var/lib/grafana/dashboards
    healthcheck:
      test: ["CMD", "curl", "-sf", "http://localhost:3000/api/health"]
      interval: 5s
      timeout: 3s
      retries: 20
```
And once I launched it, I logged in and manually created an EDITOR user named `editor` with password `editor`. There is a "prometheus" datasource provisioned in there already.

I then launched mcp with
```
GRAFANA_URL="http://localhost:3003" GRAFANA_USERNAME=editor GRAFANA_PASSWORD=editor OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:4317 OTEL_EXPORTER_OTLP_INSECURE=true go run ./cmd/mcp-grafana --transport streamable-http --log-level debug --debug --metrics --enabled-tools search,datasource,incident,prometheus,loki,elasticsearch,influxdb,alerting,dashboard,folder,oncall,asserts,sift,pyroscope,navigation,proxied,annotations,rendering,admin,clickhouse,cloudwatch,athena --address=localhost:8008
```
which then I can interrogate with the MCP inspector tool, e.g. this should work correctly:
```
npx @modelcontextprotocol/inspector --cli http://localhost:8008/mcp --transport http  --method tools/call --tool-name get_datasource --tool-arg uid=prometheus
{
  "content": [
    {
      "type": "text",
      "text": "{\"id\":1,\"isDefault\":true,\"jsonData\":{\"directUrl\":\"http://prometheus:9090\"},\"name\":\"Prometheus\",\"type\":\"prometheus\",\"uid\":\"prometheus\",\"url\":\"/api/datasources/proxy/1\"}"
    }
  ]
}
```

gerboland (changes requested):
see above

jinwonjeong-creator:
Thanks a lot for taking the time to test this on a real 8.5 — you caught a genuine flaw in the route strategy, and your docker-compose snippet made it easy to reproduce.

**Reproduced on your exact setup** (grafana:8.5.27 @ your digest, provisioned datasources, basic-auth editor): your `query_prometheus` call failed with the same `Skip: do not know how to skip: 0` error. Root cause: Grafana 8.5 registers only `/datasources/proxy/:id/*` ([v8.5.27 api.go](https://github.com/grafana/grafana/blob/v8.5.27/pkg/api/api.go)), so `proxy/uid/prometheus/...` falls into `:id`, fails int-parsing, and answers **400 `{"message":"id is invalid"}`** — a status the fallback transport didn't retry, so the non-Prometheus body reached client_golang's JSON parser. (On 7.5 the same route happens to answer 500, which *was* retried — that's why my earlier live testing missed it.)

**Fix — plain IDs, as you suggested, clearly separated:**
- `datasourceProxyPaths` is back to its original signature; the modern codepath is textually unchanged.
- Each datasource client constructor checks `fallbackProxyBase` first: when the datasource was resolved through the frontend-settings fallback, the client's **primary base is the numeric-id route** — no reliance on version-specific error codes, and no wasted first request. The uid proxy route remains only as the transport-level fallback for the opposite case (newer Grafana + RBAC-restricted token, where the legacy id routes are disabled by default since 13.0 and answer 404). The ordering is now uniform across all four call sites — the previous prom/loki argument asymmetry was part of the bug.

**Testing this time:**
- Behavior-matrix unit tests (`datasources_fallback_behavior_unit_test.go`) that drive **real queries through the Prometheus/Loki clients** against a fake Grafana answering per-generation: 7.5 (uid→500), **8.5 (uid→400, your case — red before the fix)**, 13+RBAC (numeric→404), and modern (asserts frontend/settings is never touched). The old tests only asserted path strings, which is exactly how this escaped.
- A `grafana-legacy-8` docker-compose service built from your snippet (8.5.27, :3003) + `legacy_grafana_integration_test.go`, which creates the `editor`/`editor` user at runtime and runs `list_datasources` / `get_datasource` / `query_prometheus` (`time()`) / `list_prometheus_label_names` end to end — your manual verification, automated under `make test-integration`.
- Re-verified your exact inspector call on 8.5.27 (now returns the scalar) and re-verified live against our Grafana 7.5.17 instance.

jinwonjeong-creator:
Addressed Bugbot's follow-up in [`a6fbcaa`](https://github.com/grafana/mcp-grafana/pull/1015/commits/a6fbcaa): when a datasource is referenced by *name*, the uid-based transport fallback was built from the caller's string — on a deployment with the numeric routes disabled it would request `/api/datasources/proxy/uid/{name}` and miss. The fallback cache now stores the resolved uid alongside the numeric id, and both proxy bases are constructed from the resolution result. Added `TestQueryPrometheus_NameReferenceUsesResolvedUID` (name-based query on a numeric-routes-disabled deployment, asserting the retry hits the real uid). Full unit suite green.

@gerboland the PR is ready for another look — the review response is contained in commits `506d7a6` (plain numeric-id routing + the behavior-matrix unit tests and the `grafana-legacy-8` compose service/integration test built from your setup) and `a6fbcaa` above, on top of the state you tested. Happy to squash at merge.

jinwonjeong-creator:
@gerboland Gentle reminder for another look. :)

gerboland:
Thanks for your patience. You've addressed nearly all of my concerns, I've checked these tools:
- list_datasources (with a small caveat, see below)
- get_datasource (problem with 'name' input though, see below)
- update_datasource does fail with permission denied as I'd expect
- create_datasource I'm expecting will also fail eventually
- list_prometheus_label_names & list_prometheus_label_values work
- query_prometheus works now, and query_prometheus_histogram seems ok too
- list_prometheus_metric_names & list_prometheus_metric_metadata are good too.

Problems I've found:

1. `list_datasources` includes the demo "datasource" in its list:
```
    {
      "id": -1,
      "uid": "grafana",
      "name": "-- Grafana --",
      "type": "datasource",
      "isDefault": false
    },
```
which is not actually a valid datasource, so I recommend you filter it from the results (i.e. remove entries with negative ids).

2. `get_datasource` still fails if I use "name" instead of uid:
```
npx @modelcontextprotocol/inspector --cli http://localhost:8008/mcp --transport http  --method tools/call --tool-name get_datasource --tool-arg name=prometheus
```
```json
{
  "content": [
    {
      "type": "text",
      "text": "get datasource by name prometheus: [GET /datasources/name/{name}][403] getDataSourceByNameForbidden {\"message\":\"Permission denied\"}"
    }
  ],
  "isError": true
}
{"error":{"code":"tool_is_error","message":"Tool 'get_datasource' returned isError:true."}}
```

3. I also noticed the `check_datasources_health` tool also fails. Are you able to support that at all? If not, better to return a message saying so, so the agent isn't confused.
```
npx @modelcontextprotocol/inspector --cli http://localhost:8008/mcp --transport http  --method tools/call --tool-name check_datasources_health --tool-arg uids=prometheus
```
```json
{
  "content": [
    {
      "type": "text",
      "text": "list datasources: [GET /datasources][403] getDataSourcesForbidden {\"message\":\"Permission denied\"}"
    }
  ],
  "isError": true
}
{"error":{"code":"tool_is_error","message":"Tool 'check_datasources_health' returned isError:true."}}
```

4. I noticed if you input an incorrect UID name, the response you get:
```
get datasource by name asda: [GET /datasources/name/{name}][403] getDataSourceByNameForbidden {"message":"Permission denied"}
```
differs quite a bit from the "expected":
```
get datasource by name asdas: [GET /datasources/name/{name}] getDataSourceByName (status 404): {}
```
I'd rather 404 be returned to the agent, as it will then understand the error it made.

jinwonjeong-creator:
@gerboland Thank you for yet another thorough round of real-instance testing — these reviews keep uncovering gaps I'd have missed on my own (the 8.5 pseudo-datasource shape and the case-sensitivity difference in particular), and the PR is much more solid for it. All four findings are addressed in [`041e2d5`](https://github.com/grafana/mcp-grafana/pull/1015/commits/041e2d5), each verified against a real 8.5.27 with an editor user using your exact inspector calls:

1. **Pseudo datasource in `list_datasources`** — entries with non-positive ids (`-- Grafana --` is `id: -1, uid: "grafana"` on 8.5) are now filtered at frontend-settings fetch time, so list, resolution, and the routing cache all agree. Your recommendation (drop negative ids) taken as-is.
2. **`get_datasource name=prometheus`** — the datasource's name is "Prometheus"; the fallback matched names exactly while the metadata API is case-insensitive. Name matching is now case-insensitive, and — a companion bug your finding surfaced — the name-keyed routing-cache entries are normalized the same way, so a differently-cased name reference also routes correctly end to end (covered by a new behavior test that queries via `PROMETHEUS`).
3. **`check_datasources_health`** — it can't be supported where the fallback engages (it needs the metadata API plus the uid-based health route, which doesn't exist pre-9.0), so it now returns an explicit message: *"checking datasource health is not supported on this Grafana deployment: the datasource metadata API is not accessible to this token... Datasource discovery still works via the list_datasources tool"*.
4. **Unknown identifiers now surface as not-found** — a sentinel distinguishes "settings read fine, no such datasource" from "settings unavailable", so `name=asda` returns *"datasource with name 'asda' not found. Please check if the datasource exists and is accessible"* instead of the misleading 403.

The 8.5 integration test gained subtests for all four, and the full unit suite plus a live re-check against our 7.5.17 instance are green (bonus: `get_datasource name=release` now resolves our "RELEASE" datasource there).

gerboland:
Hi @jinwonjeong-creator I've reviewed this branch and is working great, thank you.

There is a merge conflict you need to fix before we can merge however.

gerboland (dismissed):
Working well, please fix the conflict & we can merge

jinwonjeong-creator:
Bugbot's latest finding (cross-tenant poisoning of the shared `fallbackEndpoints` cache) is valid, and fixed in [`52274e6`](https://github.com/grafana/mcp-grafana/pull/1015/commits/52274e6).

**Why it appeared**: the transport cache has always been keyed by bare path strings — which was fine while all cached paths were uid-based, since uids rarely collide across tenants. This PR's legacy mode made the primary a *numeric-id* path (`/api/datasources/proxy/1/...`), and numeric ids are identical on every Grafana instance. In multi-tenant HTTP mode, a newer-Grafana tenant whose numeric routes are disabled (404 → uid retry succeeds → cache write) could therefore pin the uid-based path for a Grafana 8.x tenant sharing the same id — whose uid routes answer 400, with no retry on the cache-hit path. Single-tenant/stdio deployments were never affected (one Grafana, no version mixing).

**How it's fixed**: legacy-mode transports now bypass the shared cache entirely (`newLegacyDatasourceFallbackTransport`). The cache buys nothing in that mode — the numeric primary is expected to succeed on the deployments that use it, and the retry only fires in the rare restricted-token-on-newer-Grafana case — so this removes the poisoning vector at zero cost. Modern-path caching is untouched.

**Verification**: added `TestQueryPrometheus_NoCrossTenantCachePoisoning` (a numeric-routes-disabled tenant queries first, then an 8.x tenant with the same datasource id) — confirmed red without the fix, green with it; full unit suite green. Worth noting the underlying cache-key design predates this PR (two tenants provisioning the same uid, e.g. `prometheus`, could interact similarly on the uid paths) — happy to tenant-scope that key in a follow-up if you'd like.

jinwonjeong-creator:
@gerboland This should be ready for another look whenever you have time — thanks again for all the rounds of testing so far.

Changes since your last review:
- **Merged `main` into the branch** to resolve a conflict ([`a0870a3`](https://github.com/grafana/mcp-grafana/pull/1015/commits/a0870a3)): your new `name` filter on `list_datasources` landed on the same line this branch routes the fallback-resolved list through, so they're combined — the name filter now applies to fallback-resolved datasources too.
- **One minor fix** from a Bugbot finding ([`52274e6`](https://github.com/grafana/mcp-grafana/pull/1015/commits/52274e6)): legacy-mode transports now bypass the shared endpoint cache, since numeric-id paths are identical across tenants and one tenant's cached fallback could redirect another's requests in multi-tenant HTTP mode. Covered by a cross-tenant regression test (red before the fix).

Re-verified the merged head end to end: your full inspector command set against a real 8.5.27 with the editor user (query, get_datasource by uid/name, the not-found wording, pseudo-free list — including the new name filter — and the health-tool message), the 8.5 integration test suite, and a live check against our 7.5.17 instance. All green.

gerboland (dismissed):
thank you!

gerboland:
@jinwonjeong-creator  I see your commits are marked as "Unverified" i.e. not signed. We have a company policy that all git commits needs to be signed before they'll land. Can you please resign the commits, then this will land. Sorry I didn't spot this earlier

jinwonjeong-creator:
Bugbot's follow-up ("Legacy retry masks real errors") is a fair catch — fixed in [`ead3ceb`](https://github.com/grafana/mcp-grafana/pull/1015/commits/ead3ceb).

**Why it happened**: legacy mode inherited the modern retry statuses (401/403/500), and the transport always returns the retry response. On pre-9.0 the uid-based routes answer routing noise (400 `"id is invalid"` on 8.x), so a *genuine* 401/403/500 from the working numeric primary — say, the datasource timing out — was replaced by that noise, hiding the real error from the agent.

**Fix**: legacy mode now retries on **404 only** — the single scenario its uid fallback exists for (numeric routes disabled by default since 13.0). A nice side effect: the modern path's retry statuses revert to exactly the upstream set (401/403/500), since the earlier 404 addition existed solely for the legacy scenario — the shared transport behavior is now untouched by this PR.

**Verification**: new regression test (`TestQueryPrometheus_LegacyRealErrorsNotMasked`, red before the fix) asserts a real 500 from the numeric primary reaches the caller unmasked with the uid route never consulted; full unit suite, the Grafana 8.5.27 integration suite, and a live check against our 7.5.17 instance are all green.

sd2k (approved):

PR #1016, opened 2026-07-28
Enrich telemetry to include more tool dimensions

jcolladokuri:
Enrich MCP server telemetry with tool-argument dimensions (operation / resource_type / target)

What

Derives low-level dimensions from tool-call arguments and attaches them to the server's per-call telemetry, split by cardinality:

- Metric (mcp_server_operation_duration_seconds) — two bounded labels, appended only when the argument is present:
  - `mcp.tool.operation` → `mcp_tool_operation` (from the operation arg)
  - `mcp.tool.resource_type` → `mcp_tool_resource_type` (from the type arg, e.g. datasource plugin type)
  - `mcp.tool.phase` -> `mcp_tool_phase` to distinguish phases for multi step tools (i.e. create datasource schema fetching vs actual creation)
- Span — the above plus the high-cardinality mcp.tool.target (uid, else name), kept off the metric to avoid Prometheus series blow-up.

Why

Per-call durations aren't sliceable today by what the call is actually doing for example `alerting_manage_rules` have multiple operations. Additionally, for tools that are multi step like create datasource, there is no way to differentiate a tool call that was just to fetch the schema and one for the second phase that actually creates the datasource.
- target is the correlation key for reconstructing task spans across calls, but only makes sense on traces (unbounded), so it's span-only -- it needs an HTTP transport and tracing enabled

 Note: mcp.tool.target/correlation is span-only — it needs an HTTP transport + tracing enabled


observability/observability.go
```diff
@@ -365,6 +388,122 @@ func toolNameFromMessage(method mcp.MCPMethod, message any) (string, bool) {
...
+		return toolArgDims{}
+	}
+	req, ok := message.(*mcp.CallToolRequest)
+	if !ok || req == nil {
+		return toolArgDims{}
+	}
+	args := req.GetArguments()
+	dims := toolArgDims{
+		operation:    stringArg(args, "operation"),
+		resourceType: stringArg(args, "type"),
```
ioanarm, line 506: We're pulling type/operation off every tool's args by name here. This could potentially bite us on cardinality if some tool has a free-text type, right? Wondering if we should make this opt-in per tool, or at least add a note that these args must be bounded enums. wdyt?

tools/datasources.go
```diff
@@ -23,6 +24,28 @@ const (
...
+// withToolPhase declares a telemetry phase on a tool result via _meta so the
+// observability hooks can record it. Returns r for call-site chaining.
+func withToolPhase(r *mcp.CallToolResult, phase string) *mcp.CallToolResult {
+	if r.Meta == nil {
+		r.Meta = &mcp.Meta{}
+	}
+	if r.Meta.AdditionalFields == nil {
+		r.Meta.AdditionalFields = map[string]any{}
+	}
+	r.Meta.AdditionalFields[observability.ToolPhaseMetaKey] = phase
```
ioanarm, line 47: This puts mcp.tool.phase into the result _meta, so clients see it too. Is that intended, or should it stay server-side only? Not a blocker, just asking.

observability/observability.go
```diff
@@ -450,8 +589,9 @@ func (o *Observability) MCPHooks() *server.Hooks {
 					return
 				}
 				duration := time.Since(startTime.(time.Time))
+				o.enrichSpanWithToolDims(ctx, method, message, result)
```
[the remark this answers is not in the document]
ioanarm: +1 here and it might be worth a line in the PR description that target/correlation needs HTTP + tracing + metrics (or slow-log) enabled, since it no-ops otherwise.

ioanarm (commented):
Thanks! I ran it locally and the enriched metrics work great (schema vs created shows up as separate series). Two small questions: should we document the three new labels in `observability-metrics-and-tracing.md`?

observability/observability.go
```diff
@@ -320,6 +321,28 @@ func (o *Observability) buildOperationAttrs(ctx context.Context, method mcp.MCPM
...
 
+	// Bounded, low-cardinality dimensions derived from the tool-call arguments,
+	// so per-call durations can be sliced by sub-operation and resource type
+	// (e.g. create_datasource by plugin type). Appended only when present, so
+	// tools without these arguments keep their existing metric series identity.
+	// The high-cardinality target (datasource name/uid) is deliberately NOT put
+	// on the metric — it goes on the span instead (see enrichSpanWithToolDims).
+	dims := toolArgDimensions(method, message)
+	if dims.operation != "" {
+		attrs = append(attrs, attribute.String(attrKeyToolOperation, dims.operation))
```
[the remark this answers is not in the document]
jcolladokuri: bounded to using these labels for allowed operations only

observability/observability.go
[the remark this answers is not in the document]
jcolladokuri: added a PR description on the target correlation key, and also made it so that we only return empty hooks if all of metrics, slowLog AND tracing is off.

observability/observability.go
jcolladokuri: good call out! I have made it so that there is an allowlist of tools that are able to attach these dimensions. Made it into its own exported `ToolMetricDimensions` so that we can reuse it in the cloud server!

tools/datasources.go
jcolladokuri: Yes! Since it is non sensitive metadata, it was easier to add it directly from here which has direct context on the phase the tool is in, and read from result

ioanarm (approved):

PR #1017, opened 2026-07-28
Add schema guidance for edit datasources tooling

jcolladokuri:
Before, update was not using any schema guidance and was in general working well, but through more testing we discovered that for fields that had legacy but similar field names (i.e. `traceToLogs` vs `traceToLogsV2`) the agent would sometimes use the legacy field names instead.

This PR adds schema guidance to update tool so that we can guard against these cases and consistently use the right fields.


gerboland (approved):
Looks ok to me, thank you

PR #1025, opened 2026-07-29
pyroscope: default query_pyroscope profile output to a per-function table

aleks-p:
We benchmark LLM agents on profiling-analysis tasks against Pyroscope through different tool surfaces (profilecli, gcx, grafana-mcp). grafana-mcp scored lowest for every model we tested, and the dominant cause is the profile representation `query_pyroscope` returns, not model capability.

`query_pyroscope` returns the Pyroscope render API's DOT call graph (`format=dot`, `max-nodes=100`). That representation goes through two truncations and a name-mangling step before the model sees it.

## What this PR does

- `query_pyroscope` profile output defaults to a per-function table in `pprof -top` style — `flat, flat%, sum%, cum, cum%, function` — built from `SelectMergeStacktraces` and aggregated per fully-qualified function name. The header reports coverage, and when the server collapsed part of the tree into the synthetic `other` bucket, a note reports the unattributable share instead of hiding it.
- DOT stays available via `format="dot"` for call-graph-shaped questions; `cleanupDotProfile` no longer deletes the `other` node, so the header percentages are precise.
- `max_node_depth` description now says what it does (max table rows / graph nodes — a count, not a depth). Renaming it would break existing callers now that unknown arguments are rejected (#997), so the name stays.

Sample output:

```
Total: 2.26s
Showing top 15 out of 826 functions by flat (self) value, accounting for 41.15% of the total
        flat  flat%   sum%          cum   cum%  function
    320.00ms 14.16% 14.16%     320.00ms 14.16%  internal/runtime/syscall.Syscall6
    130.00ms  5.75% 19.91%     130.00ms  5.75%  runtime.futex
     30.00ms  1.33% 29.20%     170.00ms  7.52%  runtime.mallocgc
```

## Benchmark evidence

Full before/after benchmark of this exact change: 19 profiling-analysis tasks x 3 attempts x 2 models, grafana-mcp tool surface only, identical corpus and harness - only the mcp-grafana binary differs (upstream main `20648a1` vs this branch):

| grafana-mcp agent | baseline (DOT) | this PR (table) |
|---|---:|---:|
| claude-opus-5 mean reward | 70.8% | **90.4%** |
| claude-opus-5 pass^3 | 47.4% | **73.7%** |
| claude-opus-5 pass@3 | 89.5% | **100%** |
| claude-sonnet-5 mean reward | 52.7% | **68.9%** |
| claude-sonnet-5 pass^3 | 31.6% | **42.1%** |
| claude-sonnet-5 pass@3 | 57.9% | **84.2%** |

The table build was also cheaper and faster for the same work (opus-5 cell: 12.1M vs 14.8M tokens, $66 vs $82, 73m vs 99m).

Notes:
- The table intentionally ranks by flat; entry-point functions with negligible self time appear only via the `cum` column of listed rows. Call-graph questions are what `format="dot"` remains for.
- A function literally named `other` would be mistaken for the truncation bucket; Pyroscope reserves that name for truncation everywhere else, and the previous code deleted such nodes outright.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Changed
+
+- `query_pyroscope` now returns a per-function table (`pprof -top` style: flat/cum per fully-qualified function name) by default instead of a line-level DOT call graph. The DOT call graph remains available via `format="dot"` and no longer deletes the `other` truncation node ([#1025](https://github.com/grafana/mcp-grafana/pull/1025))
+
 ## [1.0.0] - 2026-07-28
 
 ### Added
--- a/tools/pyroscope.go
+++ b/tools/pyroscope.go
@@ -9,6 +9,8 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
+	"sort"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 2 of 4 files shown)


gerboland (approved):
Wonderful, thank you!

gerboland:
Oh there's an e2e test failing, should be minor fix

aleks-p:
Thanks @gerboland. 

> Oh there's an e2e test failing, should be minor fix

The E2E tests are passing after a re-run, it seems there is some flakiness.

PR #1026, opened 2026-07-29
feat(tools): add ask_assistant tool (opt-in, write-gated)

ioanarm:
## Summary

- Adds an `ask_assistant` tool that forwards a natural-language prompt to Grafana Assistant via the `grafana-assistant-app` plugin's A2A (JSON-RPC / SSE) endpoint and returns the aggregated text reply. Multi-turn conversations are supported by passing the returned `contextId` back into a follow-up call.
- The tool is a **thin transport client only** — all assistant intelligence lives server-side behind the A2A endpoint, so it carries no dependency on the internal `grafana-assistant-app` module. It speaks only the documented JSON-RPC/SSE contract.
- **Disabled by default**, opt-in via `assistant` in `--enabled-tools`, and treated as a **write tool** (the assistant may invoke write tools of its own), so it is skipped under `--disable-write`. Requires the Grafana Assistant plugin on the target instance.
- Server instructions no longer advertise the Assistant capability when write is disabled, since no tool is actually registered in that case.

## Behavior details

- Transient stream-open failures (429/502/503/504, timeouts) are retried with exponential backoff; task-level failures are terminal and not retried.
- Partial replies and the `contextId` are preserved and folded into the returned error when a task fails mid-stream, so MCP clients (which only see `err.Error()` on failure) can still recover them and resume.
- A stream that ends before a terminal task state is reported as incomplete rather than being passed off as a completed answer.
- The SSE stream is bounded with `io.LimitReader(defaultResponseLimitBytes)` (10 MB, the repo's standard ceiling) to prevent unbounded memory growth over the up-to-5-minute call.
- Authentication reuses `BuildTransport` (service-account token / on-behalf-of / extra headers), and every request carries `X-App-Source: mcp-grafana` for usage attribution.

## Testing notes

- **No integration test (`//go:build integration`) is included, deliberately.** A real integration test would require the `grafana-assistant-app` plugin *and* a working LLM in the test environment, neither of which the provisioned CI stack has. The tool is instead covered by thorough unit tests that mock the A2A SSE endpoint across the happy path and every unhappy path.
- `e2e.yml` / `conftest.py` pass an explicit `--enabled-tools` list that already excludes disabled-by-default categories (e.g. `agento11y`), so `assistant` correctly does not need to be added there.

## Test plan

- [x] `go build ./...`
- [x] `go vet ./...`
- [x] `gofmt` clean
- [x] `go test ./tools/ -run 'TestAskAssistant|TestChat|TestAssistantIsRetryable|TestFormatAssistantTimeContext'` (success, multi-turn, task-failed w/ partial output, incomplete-stream, JSON-RPC error, non-200, transient retry, timeout, empty prompt, missing URL)
- [x] `go test ./cmd/mcp-grafana/ -run 'TestBuildInstructions'` (opt-in / disable-flag / write-disabled instruction cases)
- [x] Docs (`docs/sources/...`) synced with the README (tools table, command-line flags, enable/disable categories)
- [ ] Manual: `ask_assistant` single-turn and multi-turn verified via Cursor against a local `grafana-assistant-app` stack


svennergr (dismissed):

ioanarm:
cursor review

gerboland (dismissed):

gerboland (approved):

PR #1033, opened 2026-07-31
Apply --base-path to the whole listener, including /healthz and /metrics

### LINKED ISSUE #1021: Apply `--base-path` to `/healthz` and `/metrics` so a single reverse-proxy rule covers the whole service
## Summary

## Steps to reproduce

```bash
```

Then:

| Request | Result |
| --- | --- |
| `GET /healthz` | works |
| `GET /metrics` | works |

## Current behavior

```go
if obs.MetricsEnabled {
    if obs.MetricsAddress == "" {
    }
    ...
}
```

- `/grafana-mcp/sse` works ✅

## Environment

- **Version:** _0.17.2_

Tarasusrus:
## Summary

`--base-path` exists so a single path-routed reverse-proxy rule can cover the service, but it only reached the MCP handler — and not even reliably:

- the mux mounted the SSE server on an exact-match pattern, so a base path given without a trailing slash (the documented form) never routed `/<base>/sse` at all;
- `/healthz` and `/metrics` were registered at the server root regardless of the prefix;
- for streamable-http the flag was ignored entirely, though `README.md` and the flag reference both document it for that transport.

The base path is now normalized once and every mount derives from it. `/healthz` and `/metrics` stay mounted at the root as well, so probes already pointing there keep working.

## Before / after

Measured on the built binary, `-t sse --base-path … --metrics`:

| base path | `/base/sse` | `/base/healthz` | `/base/metrics` | `/healthz` |
| --- | --- | --- | --- | --- |
| `/my-base` — before | 404 | 404 | 404 | 200 |
| `/my-base/` — before | 200 | 404 | 404 | 200 |
| either — after | 200 | 200 | 200 | 200 |

## Behavior change worth flagging

With `-t streamable-http` and `--base-path` set, the MCP endpoint moves from `/mcp` to `/<base>/mcp`. Today the flag is silently ignored for that transport, so this aligns the code with the documentation, but anyone passing both flags will see the endpoint URL change.

## Validation

- `TestHTTPMuxHonoursBasePath` covers both transports, base path unset / with / without trailing slash, that MCP endpoints do *not* stay reachable at the root, and that an `--endpoint-path` colliding with `/healthz` does not panic on duplicate mux registration.
- `go test -tags unit ./...`, `make lint` — clean.
- Smoke-tested both transports against a running binary.

Fixes #1021


gerboland (changes requested):
In my experience, it is convention to serve internal-only /healthz and /metrics from the root path only, while allowing the application endpoints to use a base prefix.

Those internal endpoints are consumed by internal infrastructure only, and should not be made available externally. It's harder to configure a reverse proxy to not expose them, if they're on the same base path as the application endpoints.

cmd/mcp-grafana/main.go
```diff
@@ -476,6 +477,71 @@ func handleHealthz(w http.ResponseWriter, r *http.Request) {
...
+	}
+	return mux
+}
+
+// newSSEMux mounts the SSE handler and the operational endpoints.
+func newSSEMux(mcpHandler http.Handler, basePath string, metricsHandler http.Handler) *http.ServeMux {
+	// The SSE server routes on the full request path (<base>/sse,
+	// <base>/message), so it needs the subtree pattern: an exact-match mount on
+	// the base path itself never reaches it.
+	return newHTTPMux(normalizeBasePath(basePath)+"/", mcpHandler, metricsHandler)
```
gerboland: `normalizeBasePath(basePath)` was already done on line 644 when the result was passed in. Best to normalize it once and all consumers from there assume it normalized.

cmd/mcp-grafana/main.go
```diff
@@ -476,6 +477,71 @@ func handleHealthz(w http.ResponseWriter, r *http.Request) {
...
+}
+
+// newStreamableHTTPMux mounts the streamable-http handler and the operational
+// endpoints.
+func newStreamableHTTPMux(mcpHandler http.Handler, basePath, endpointPath string, metricsHandler http.Handler) *http.ServeMux {
+	return newHTTPMux(streamableEndpointPath(basePath, endpointPath), mcpHandler, metricsHandler)
```
gerboland: Same happens here for `streamableEndpointPath(basePath, endpointPath)` - it was already assigned to `mcpEndpoint` in the caller's context

cmd/mcp-grafana/main.go
```diff
@@ -476,6 +477,71 @@ func handleHealthz(w http.ResponseWriter, r *http.Request) {
...
+	// A configurable endpoint path can collide with an operational path;
+	// registering the same pattern twice would panic, so keep the MCP mount
+	// and warn instead.
+	if mcpPattern == "/healthz" {
+		slog.Warn("Not mounting health endpoint: path is taken by the MCP endpoint", "path", mcpPattern)
+	} else {
+		mux.HandleFunc("/healthz", handleHealthz)
+	}
+	if metricsHandler != nil {
+		if mcpPattern == "/metrics" {
```
gerboland: Unlikely but if `--endpoint-path` happens to equal `/healthz` or `/metrics`, the code silently skips mounting the real operational handler, logs a slog.Warn, and lets the MCP handler answer on that path instead. That means a misconfigured flag makes health checks vanish with nothing but an easily-missed log line — no startup error. 

I'd expect this to be rejected at flag-parsing time instead: `--endpoint-path` cannot be `/healthz` or `/metrics`

gerboland (changes requested):

gerboland:
@Tarasusrus hi, thanks for the latest push, but it has conflicts & Cursor has raised a good point. I'll review once they are fixed

Tarasusrus:
Rebased onto main and fixed the point Cursor raised. Two commits now.

**Conflicts.** The branch was 52 commits behind, so I squashed my four commits into one before rebasing to keep it to a single resolution. The one worth mentioning is caller auth, which landed meanwhile: `callerToken` is threaded through the mux helpers keeping the wrapping order from main — `withCallerAuth` outside `observability.WrapHandler`, so requests rejected with 401 aren't counted — and `/healthz` and `/metrics` stay outside caller auth so probes and scrapers keep working.

**Cursor's finding.** It was right, and the case is a bit wider than reported. The guard compared the raw `--endpoint-path`, while the mount goes through `path.Join` with `--base-path`, so `healthz`, `/healthz/` and `/foo/../metrics` all resolved to a reserved path and panicked `ServeMux` on the duplicate registration. The check now runs on the resolved mount pattern, plus two things on top:

- `--base-path` can spell an operational path on its own (`--base-path /healthz --endpoint-path /`), so it is part of the check.
- `/metrics` is only mounted when the metrics handler shares the listener, so it is only reserved when `--metrics` is set without `--metrics-address`. Otherwise the path is free, and rejecting it would break a configuration that works today.

I also moved the validation into `main()`, next to `validateServerName` and the Loki guardrail checks, so a bad value prints one line to stderr and exits 2 instead of coming out of `run()` as a panic. That is what you asked for with "rejected at flag-parsing time", and it matches how every other flag here is handled.

While there: applying `--base-path` to streamable-http makes a pre-existing crash reachable for that transport. `ServeMux` reads a pattern as `[METHOD ]/path`, so `--base-path "/my base"` panics with `invalid method "/my"`, and `--endpoint-path "/{mcp}"` registers fine but quietly becomes a wildcard matching any single segment. Both are rejected now. Happy to pull that out into its own PR if you would rather keep this one to the base-path fix.

**One thing I would like your call on.** Reviewing this, `newSSEMux` came up twice: it documents "basePath must already be normalized by the caller" in a comment only. Pass a raw `/foo/` and the pattern becomes `/foo//`, which `ServeMux` accepts without complaint and which no request can reach — SSE would be dead with a clean startup log. The obvious fix is to call `normalizeBasePath` inside `newSSEMux`, since it is idempotent, but that is exactly the call you asked me to drop, so I have not put it back. Instead both the validation and the mount derive the pattern from the raw flag through `normalizeBasePath`, pinned by a test that fails if either side stops doing it. Say the word if you would rather have the defensive call in the helper.

The CHANGELOG entry now also carries an upgrade note for the streamable-http behaviour change, since `--base-path` was previously accepted and ignored by that transport.

Validation: `go test -tags unit ./...` and `make lint` clean. Both guards are mutation-checked — breaking either turns the tests red, including reproducing the exact `ServeMux` panic Cursor predicted — and I smoke-tested the built binary on both transports for routing and for every rejected flag combination.

gerboland:
Sorry for the slow reply. Please resolve conflicts so I can take a look

gerboland:
More conflicts, sorry

Tarasusrus:
Rebased onto main (usage-stats landed meanwhile — `run()` signature and `WithStateLess(true)` carried over). Also dropped the CHANGELOG entry per CONTRIBUTING.md, so this shouldn't conflict on every release anymore. Ready for another look.

gerboland:
Thanks you @Tarasusrus I'm happy with the change. I've edited the PR title & description to update it to the current change, I hope you don't mind. Merging.

gerboland (approved):

PR #1034, opened 2026-07-31
Normalize GRAFANA_URL where it enters the process

### LINKED ISSUE #1032: GRAFANA_URL is normalized too late: schemeless values panic at startup or produce hostless requests
## Summary

`normalizeGrafanaURL` — which trims trailing slashes and supplies a missing scheme — runs at the very end of the context-building chain, in `WithGrafanaConfig`. Three consumers read the raw environment/header value *before* that point, so they disagree with `GrafanaConfig.URL` about what the Grafana URL is.

The result: a schemeless `GRAFANA_URL`, a form the normalizer explicitly supports and unit tests cover, either panics at startup or produces a client whose requests carry no host.

## Steps to reproduce

Reproduced on `main` at 92fc809.

```bash
$ GRAFANA_URL=127.0.0.1:3000 mcp-grafana -t stdio
panic: invalid Grafana URL 127.0.0.1:3000: parse "127.0.0.1:3000": first path segment in URL cannot contain colon
```

```bash
$ GRAFANA_URL=grafana.example.com mcp-grafana -t stdio
level=INFO msg="Using Grafana configuration" url=grafana.example.com
level=WARN msg="Failed to fetch frontend settings" error="Get \"grafana.example.com/api/frontend/settings\": unsupported protocol scheme \"\""
level=ERROR msg="failed to initialize proxied tools for stdio" error="failed to discover MCP datasources: failed to list datasources: Get \"https:///grafana.example.com/api/datasources\": http: no Host in request URL"
```

In the second case `GrafanaConfig.URL` is `https://grafana.example.com` (correct), so tools that build requests from `cfg.URL` work while everything going through the generated OpenAPI client fails — a split tool surface on one configuration value.

## Root cause

Consumers of the un-normalized value:

| Location | Uses raw value for |
| --- | --- |
| `mcpgrafana.go:849` (`ExtractGrafanaInfoFromEnv`) | `url.Parse(u)` → panics on `host:port` |
| `mcpgrafana.go:1292-1297` (`ExtractGrafanaClientFromEnv`) | `NewGrafanaClient(u, …)` → `parsedURL.Host` is empty for a schemeless name |
| `ExtractGrafanaClientFromHeaders` via `extractKeyGrafanaInfoFromReq` | same, for `X-Grafana-URL` |
| `client_cache.go:34` (`cacheKeyFromRequest`) | cache key |

`urlAndAPIKeyFromEnv` (`mcpgrafana.go:56`) trims the trailing slash but does not apply the scheme defaulting, so it is a second, weaker normalizer for the same value.

The cache key consequence is milder but in the same class: `https://grafana.example.com`, `https://grafana.example.com/` and `grafana.example.com` produce three distinct keys, so one instance can hold three cached clients — while `normalizeGrafanaURL` collapses all three to a single value.

## Why the existing tests pass

`mcpgrafana_test.go:784-788` asserts exactly this behavior, including `"127.0.0.1:3000" → "http://127.0.0.1:3000"`, but only against `normalizeGrafanaURL` in isolation. Nothing asserts that the OpenAPI client and the cache key are built from the same value the rest of the code sees, which is where the gap is.

## Suggested fix

Normalize at the boundary — in `urlAndAPIKeyFromEnv` and `extractKeyGrafanaInfoFromReq` — so every consumer derives from one normalized value. The existing `WithGrafanaConfig` call can stay as an idempotent safety net.

Regression coverage worth adding alongside it:
- `GRAFANA_URL=host:port` starts without panicking;
- for a schemeless URL, the client issues requests to the same host `GrafanaConfig.URL` names;
- three spellings of one instance map to one cache key.

I have this working locally and am happy to send a PR if the approach looks right.

Tarasusrus:
## Summary

`normalizeGrafanaURL` trims the trailing slash and supplies a missing scheme, and `TestNormalizeGrafanaURL` pins that behavior — including `"127.0.0.1:3000" → "http://127.0.0.1:3000"`. But it ran only at the end of the chain, in `WithGrafanaConfig`. Consumers that read the environment value earlier (`ExtractGrafanaInfoFromEnv`, `ExtractGrafanaClientFromEnv`, and the cached extractors) saw the raw string, so they disagreed with `GrafanaConfig.URL` about which instance to talk to.

This moves the normalization into `urlAndAPIKeyFromEnv`, where the value enters the process, so the config, the API client and the cache key all derive from one URL.

## Before / after

Measured with `-t stdio` on a build of this branch versus `main`:

| `GRAFANA_URL` | before | after |
| --- | --- | --- |
| `127.0.0.1:3000` | `panic: invalid Grafana URL 127.0.0.1:3000: parse …: first path segment in URL cannot contain colon` | starts; requests go to `http://127.0.0.1:3000` |
| `grafana.example.com` | `Get "https:///grafana.example.com/api/datasources": http: no Host in request URL` | `Get "https://grafana.example.com/api/datasources"` |
| `grafana.example.com/grafana` | hostless request | `https://grafana.example.com/grafana/api/frontend/settings` |
| `http://localhost:3000` | works | unchanged |

In the schemeless cases `GrafanaConfig.URL` was already correct, so tools that build requests from it kept working while everything going through the generated OpenAPI client failed — a tool surface split in half by one configuration value.

## Scope: the header path is deliberately untouched

`X-Grafana-URL` is still taken as-is. `ValidateGrafanaURLMiddleware` rejects a schemeless header today, and having the server guess a scheme for a caller-supplied host would widen what that header is allowed to name — a decision for maintainers, not a side effect of this fix. This PR adds test cases pinning that boundary (`schemeless host rejected`, `schemeless header rejected` → 400).

I also tried normalizing the client-cache key so that spellings of one instance share an entry, and dropped it: the cached client is still constructed from the raw URL, so a normalized key would let a broken client built from one spelling be reused by every other spelling. Normalizing at the source gives the same collapsing for environment-derived URLs without that hazard.

## Validation

New coverage, and what each part is for:

- proves the fix (fails if the normalization is reverted to the old `strings.TrimRight`): schemeless, whitespace-padded and schemeless-subpath spellings of `GRAFANA_URL` all resolve to the same URL in `GrafanaConfig`, in the path the API client actually requests, and in the value the cached extractors key on;
- regression guards (green either way, by design): `normalizeGrafanaURL` is idempotent — it now runs twice, at the boundary and in `WithGrafanaConfig`; environment credentials are still bound to the environment instance when `GRAFANA_URL` is schemeless (same instance keeps the token, a foreign host does not); a blank `GRAFANA_URL` still falls back to the default.

`go test -race -count=1 -tags unit ./...` and `make lint` are clean.

Fixes #1032

### DIFF
--- a/mcpgrafana.go
+++ b/mcpgrafana.go
@@ -53,10 +53,6 @@ const (
 )
 
 func urlAndAPIKeyFromEnv(logger *slog.Logger) (string, string) {
-	// Normalized here, at the point the value enters the process, so that every
-	// consumer — the config in the context, the API client, and the client cache
-	// key — derives from the same URL. Normalizing only in WithGrafanaConfig left
-	// the earlier consumers reading the raw string.
 	u := normalizeGrafanaURL(os.Getenv(grafanaURLEnvVar))
 
 	// Check for the new service account token environment variable first.
--- a/mcpgrafana.go
+++ b/mcpgrafana.go
@@ -53,7 +53,11 @@ const (
 )
 
 func urlAndAPIKeyFromEnv(logger *slog.Logger) (string, string) {
-	u := strings.TrimRight(os.Getenv(grafanaURLEnvVar), "/")
+	// Normalized here, at the point the value enters the process, so that every
+	// consumer — the config in the context, the API client, and the client cache
+	// key — derives from the same URL. Normalizing only in WithGrafanaConfig left
+	// the earlier consumers reading the raw string.
+	u := normalizeGrafanaURL(os.Getenv(grafanaURLEnvVar))
 
 	// Check for the new service account token environment variable first.
 	apiKey := os.Getenv(grafanaServiceAccountTokenEnvVar)
--- a/validate_url_test.go
+++ b/validate_url_test.go
@@ -38,6 +38,12 @@ func TestValidateGrafanaURL(t *testing.T) {
 		// Invalid inputs.
 		{"empty string", "", true},
 		{"slash-only trims to empty", "/", true},
+		// A schemeless value is accepted from GRAFANA_URL, where
+		// normalizeGrafanaURL supplies the scheme, but not from a caller-supplied
+		// header: guessing a scheme for a caller-controlled host would widen what
+		// the header is allowed to name.
+		{"schemeless host rejected", "grafana.example.com", true},
+		{"schemeless host and port rejected", "grafana.example.com:3000", true},
 		{"plain text", "not a url", true},
 		{"invalid percent encoding", "http://%gg", true},
 		{"javascript scheme", "javascript:alert(1)", true},
@@ -84,6 +90,7 @@ func TestValidateGrafanaURLMiddleware(t *testing.T) {
 		{"malformed percent encoding rejected", true, "http://%gg", http.StatusBadRequest, false},
 		{"javascript scheme rejected", true, "javascript:alert(1)", http.StatusBadRequest, false},
 		{"relative path rejected", true, "/relative", http.StatusBadRequest, false},
+		{"schemeless header rejected", true, "grafana.example.com", http.StatusBadRequest, false},
 		{"trim-to-empty slash rejected", true, "/", http.StatusBadRequest, false},
 		{"empty host rejected", true, "http://", http.StatusBadRequest, false},
 		{"CR-LF injection attempt rejected", true, "http://foo\r\nX-Injected: 1", http.StatusBadRequest, false},
--- a/validate_url_test.go
+++ b/validate_url_test.go
@@ -38,10 +38,6 @@ func TestValidateGrafanaURL(t *testing.T) {
 		// Invalid inputs.
 		{"empty string", "", true},
 		{"slash-only trims to empty", "/", true},
-		// A schemeless value is accepted from GRAFANA_URL, where
-		// normalizeGrafanaURL supplies the scheme, but not from a caller-supplied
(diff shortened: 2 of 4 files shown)


mcpgrafana.go
```diff
@@ -53,7 +53,11 @@ const (
 )
 
 func urlAndAPIKeyFromEnv(logger *slog.Logger) (string, string) {
-	u := strings.TrimRight(os.Getenv(grafanaURLEnvVar), "/")
+	// Normalized here, at the point the value enters the process, so that every
+	// consumer — the config in the context, the API client, and the client cache
+	// key — derives from the same URL. Normalizing only in WithGrafanaConfig left
+	// the earlier consumers reading the raw string.
```
gerboland: I don't think this comment is necessary

validate_url_test.go
```diff
@@ -38,6 +38,12 @@ func TestValidateGrafanaURL(t *testing.T) {
 		// Invalid inputs.
 		{"empty string", "", true},
 		{"slash-only trims to empty", "/", true},
+		// A schemeless value is accepted from GRAFANA_URL, where
+		// normalizeGrafanaURL supplies the scheme, but not from a caller-supplied
+		// header: guessing a scheme for a caller-controlled host would widen what
+		// the header is allowed to name.
```
gerboland: I'd also prefer this comment be erased. The code itself make the intent pretty clear

gerboland (changes requested):
Minor changes requested

gerboland (approved):
Thank you!

PR #1041, opened 2026-08-01
fix(run_panel_query): resolve constant and textbox variables from their query field

micdenny:
## Problem

`extractTemplateVariables` reads `current.value`, falling back to `current.text`. Variables of type `constant` and `textbox` keep their value in the `query` field, and Grafana routinely saves them with no `current` object at all — so they resolve to nothing and the raw `$name` survives `substituteTemplateVariables`.

Prometheus and Loki tolerate the leftover token in most cases. SQL datasources do not: the statement reaches the backend plugin with a bare `$name` in it and fails as a syntax error.

Concretely, a dashboard whose templating list is

```json
{ "name": "conversionRate", "type": "constant", "query": "158.87" }
```

sends `... WHERE rate > $conversionRate` to the datasource unchanged.

On the dashboard set I hit this with, 6 of 23 dashboards were affected, all of them using `constant` variables for conversion rates and lookback windows.

## Change

Fall back to the `query` field when `current` yields nothing, restricted to the two types where that field holds a literal value:

- `constant`
- `textbox`

Type `query` is deliberately excluded — there `query` holds a datasource expression such as `label_values(up, job)`, which must never be substituted as a value. An existing `current` still takes precedence.

## Tests

Four cases added to `TestExtractTemplateVariables`:

- constant without `current` resolves from `query`
- textbox without `current` resolves from `query`
- `current` wins over `query` when both are present
- type `query` does not fall back

`go test -tags unit ./tools/` passes.

### DIFF
--- a/tools/run_panel_query.go
+++ b/tools/run_panel_query.go
@@ -410,6 +410,19 @@ func extractTemplateVariables(db map[string]interface{}) map[string]string {
 				}
 			}
 		}
+
+		// Constant and textbox variables hold their value in "query" and are
+		// frequently saved without a "current" at all. Without this fallback
+		// the raw $name survives substitution and reaches the datasource,
+		// which for SQL datasources is a syntax error.
+		if variables[name] == "" {
+			switch safeString(variable, "type") {
+			case "constant", "textbox":
+				if q := safeString(variable, "query"); q != "" {
+					variables[name] = q
+				}
+			}
+		}
 	}
 
 	return variables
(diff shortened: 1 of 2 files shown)


gerboland:
Thanks for the contribution, it looks sensible but I'd appreciate instructions on how I could easily test this change.

sd2k (dismissed):
Thanks 🙏

sd2k:
@micdenny Sorry for the mistake, we have an org-wide policy that commits must be signed and this can't be merged until that's resolved. Please rebase on main and sign the commits and we'll get this in. Thanks again!

micdenny:
Rebased on `main` and the commit is now signed — the `signed-commits` check is green. The re-push dismissed your approval, sorry for the extra round trip. Thanks!

sd2k (approved):
No problem, thanks for signing!

PR #1042, opened 2026-08-01
feat(run_panel_query): support MSSQL panels

### LINKED ISSUE #999: Feature request
Add MSSQL support to run_panel_query or add a query_mssql tool.

micdenny:
Closes #999.

## Why this is small

MSSQL panels reach `/api/ds/query` with the same `rawSql` and `format` fields as BigQuery, and their backend plugin resolves the SQL macros (`$__timeFilter`, `$__timeFrom`, `$__timeGroup`) server-side. `executeBigQueryPanelQuery` was therefore already almost datasource-agnostic — as its own doc comment says. The only BigQuery-specific parts were the datasource type written into the target and the default `format` value:

```go
target["datasource"] = map[string]interface{}{"uid": datasourceUID, "type": BigQueryDatasourceType}
if _, ok := target["format"]; !ok {
    target["format"] = BigQueryFormatTable
}
```

## The one place the two differ

They are not the same stack. BigQuery is sqlds-based; MSSQL is sqleng-based ([`grafana-mssql-datasource`](https://github.com/grafana/grafana-mssql-datasource/blob/main/pkg/mssql/sqleng/sql_engine.go)), and its query model declares:

```go
type QueryJson struct {
	RawSql string `json:"rawSql"`
	...
	Format string `json:"format"`
}
```

So `format` is a string there, not the numeric enum sqlds uses. Sending `1` fails `json.Unmarshal` and the query errors out; omitting `format` altogether hits `default: panic("Unrecognized query model format")` in the plugin's format switch. The default value is consequently resolved per datasource type rather than shared.

## Change

- Generalise the executor into `executeSQLPanelQuery`, parameterised by datasource type, and route both `bigquery` and `mssql` through it.
- `defaultSQLFormat` returns the numeric `SQLFormatTable` for sqlds datasources and the string `MSSQLFormatTable` (`"table"`) for MSSQL.
- `normalizeDatasourceType` maps `mssql` (case-insensitive).
- Rename `BigQueryQueryResult` → `SQLQueryResult` and `BigQueryFormatTable` → `SQLFormatTable`; neither was ever BigQuery-specific. `BigQueryDatasourceType` is kept as-is.
- Add MSSQL entries to the empty-result hints and to the `datasourceType` override description.

BigQuery behaviour is unchanged. The raw target is still preserved so `location`, `project` and `dataset` reach the backend — and, for MSSQL, so do `fill`, `fillInterval`, `fillMode` and `fillValue`, which sqleng reads from the same JSON. `format` is still only defaulted when the panel does not set one, which is the common case for UI-authored panels; the default matters for hand-written and provisioned dashboards that omit it.

## Tests

- `TestExecuteSQLPanelQuery_NilTarget` covers both datasource types
- `TestDefaultSQLFormat` pins the numeric/string split
- `TestExtractPanelInfoMSSQL` asserts `rawSql` extraction and that the raw target survives with its own `format`
- `mssql` and `MSSQL` added to the `normalizeDatasourceType` table
- an `mssql` case added to the hints table

`go test -tags unit ./tools/` passes, `go vet` clean, `gofmt` clean.

## Note

Panels using `constant` template variables will still fail without #1041, which fixes those resolving to nothing and leaving a bare `$name` in the statement. The two changes are independent; this one is enough for panels that only use macros.

### DIFF
--- a/tools/run_panel_query.go
+++ b/tools/run_panel_query.go
@@ -24,7 +24,7 @@ type RunPanelQueryParams struct {
 	End            string            `json:"end" jsonschema:"description=Override end time (e.g. 'now'\\, RFC3339\\, Unix ms)"`
 	Variables      map[string]string `json:"variables" jsonschema:"description=Override dashboard variables (e.g. {\"job\": \"api-server\"})"`
 	DatasourceUID  string            `json:"datasourceUid,omitempty" jsonschema:"description=Override datasource UID"`
-	DatasourceType string            `json:"datasourceType,omitempty" jsonschema:"description=Override datasource type (prometheus\\, loki\\, grafana-clickhouse-datasource\\, cloudwatch\\, influxdb\\, grafana-bigquery-datasource)"`
+	DatasourceType string            `json:"datasourceType,omitempty" jsonschema:"description=Override datasource type (prometheus\\, loki\\, grafana-clickhouse-datasource\\, cloudwatch\\, influxdb\\, grafana-bigquery-datasource\\, mssql)"`
 }
 
 // QueryTimeRange represents the actual time range used for a panel query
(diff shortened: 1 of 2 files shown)


sd2k (approved):
LGTM thanks!

PR #1050, opened 2026-08-05
fix: clarify Loki index stats limitations

1cenix:
## Summary

- add an explicit `caveats` field to native Loki `query_loki_stats` responses
- document that `/loki/api/v1/index/stats` is approximate and excludes data still held in ingesters
- remove the misleading guidance to use stats as an existence check before querying logs
- add a regression test for an all-zero index-stats response

## Why

`query_loki_stats` returned zero streams/chunks/entries/bytes while a bounded `query_loki_logs` query returned log lines. This is consistent with Loki index stats excluding data still in ingesters; an all-zero result must not be interpreted as proof that a selector is empty.

Loki documents this endpoint as approximate and excluding ingester data: https://grafana.com/docs/loki/latest/reference/loki-http-api/#query-ingester-and-store

## Behavior

The numeric stats remain unchanged and advisory. The response now carries the limitation alongside the values, and the tool descriptions direct callers to a bounded log query or `count_over_time()` when they need to verify recent data or count matching lines. No synthetic counts or expensive fallback query are introduced.

## Validation

- `GOTOOLCHAIN=auto go test -tags unit ./tools -count=1`

### DIFF
--- a/tools/loki.go
+++ b/tools/loki.go
@@ -38,10 +38,11 @@ type LabelResponse struct {
 
 // Stats represents the statistics returned by Loki's index/stats endpoint
 type Stats struct {
-	Streams int `json:"streams"`
-	Chunks  int `json:"chunks"`
-	Entries int `json:"entries"`
-	Bytes   int `json:"bytes"`
+	Streams int      `json:"streams"`
+	Chunks  int      `json:"chunks"`
+	Entries int      `json:"entries"`
+	Bytes   int      `json:"bytes"`
+	Caveats []string `json:"caveats,omitempty"`
 }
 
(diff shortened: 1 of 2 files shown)


gerboland (approved):
Yep that's a good clarification

gerboland:
oh sorry @1cenix , there's a conflict, please fix and I'll merge it next time immediately

sd2k:
Hi @1cenix just checking in here, could you please resolve the conflict and sign the CLA and we'll get this merged in? Thanks!

1cenix:
Closing this — opened in error on my side and I'm not pursuing it. Apologies for the review time, and thanks @gerboland / @sd2k.

PR #1051, opened 2026-08-05
fix(tools): declare readOnly/destructive/openWorld hints on every tool

### LINKED ISSUE #1009: Add MCP tool annotations (readOnlyHint / openWorldHint / destructiveHint) to all hosted Cloud MCP tools
# Add MCP tool annotations (readOnlyHint / openWorldHint / destructiveHint) to all hosted Cloud MCP tools

## Summary

The hosted Grafana Cloud MCP server (`https://mcp.grafana.com/mcp`) does not declare the three MCP tool annotations — `readOnlyHint`, `openWorldHint`, `destructiveHint` — on every tool. This blocks listing Grafana Cloud MCP in the OpenAI plugin directory (ChatGPT + Codex): OpenAI's production tool scan requires **all three set explicitly (true/false) on every tool**, and currently fails on ~85 tools.

This is a server-side change, independent of the marketplace packaging PR (grafana/ai-marketplace#17), which is approved and ready to merge separately.

## Requirement

- Every tool the hosted server exposes must set `readOnlyHint`, `openWorldHint`, and `destructiveHint` — no omissions.
- Values must be accurate. In particular, write-capable tools must **not** be marked `readOnlyHint: true`.
- Land the change on whatever powers the **hosted** `mcp.grafana.com` endpoint and redeploy.

## Semantics used below (per MCP spec)

- `readOnlyHint: true` — the tool does not modify state (pure read).
- `destructiveHint` — only meaningful when not read-only; `true` if it can overwrite/delete, `false` if additive.
- `openWorldHint: true` — the tool reaches an external/open world (e.g. the public plugin catalog); `false` if scoped to the user's own Grafana stack and configured datasources.

## Default rule

All pure-read tools → `readOnlyHint: true, destructiveHint: false, openWorldHint: false`.

## Non-default tools (implement exactly)

| Tool | readOnlyHint | destructiveHint | openWorldHint | Rationale |
|---|---|---|---|---|
| `add_activity_to_incident` | false | false | false | additive write |
| `create_annotation` | false | false | false | additive write |
| `create_folder` | false | false | false | additive write |
| `create_incident` | false | false | false | additive write |
| `update_annotation` | false | true | false | overwrites existing |
| `update_dashboard` | false | true | false | can overwrite a dashboard |
| `alerting_manage_rules` | false | true | false | can create/update/delete rules |
| `ask_assistant` | false | true | false | can trigger arbitrary tool use, including writes |
| `grafana_api_request` | false | true | false | raw API passthrough (incl. DELETE) |
| `install_plugin` | false | false | true | installs from the public plugin catalog |
| `search_plugin_information` | true | false | true | reads the public plugin catalog |

## Read-only tools (all → readOnlyHint: true, destructiveHint: false, openWorldHint: false)

`alerting_manage_routing`, `analyze_loki_labels`, `describe_athena_table`, `describe_clickhouse_table`, `describe_infrastructure`, `describe_snowflake_table`, `generate_deeplink`, `get_alert_group`, `get_annotation_tags`, `get_annotations`, `get_assertions`, `get_current_oncall_users`, `get_dashboard_by_uid`, `get_dashboard_panel_queries`, `get_dashboard_property`, `get_dashboard_summary`, `get_datasource`, `get_incident`, `get_oncall_shift`, `get_panel_image`, `get_plugin`, `get_query_examples`, `get_resource_description`, `get_resource_permissions`, `get_role_assignments`, `get_role_details`, `list_alert_groups`, `list_all_roles`, `list_athena_catalogs`, `list_athena_databases`, `list_athena_tables`, `list_clickhouse_tables`, `list_cloudwatch_dimensions`, `list_cloudwatch_metrics`, `list_cloudwatch_namespaces`, `list_datasources`, `list_graphite_metrics`, `list_graphite_tags`, `list_incidents`, `list_loki_label_names`, `list_loki_label_values`, `list_oncall_schedules`, `list_oncall_teams`, `list_oncall_users`, `list_prometheus_label_names`, `list_prometheus_label_values`, `list_prometheus_metric_metadata`, `list_prometheus_metric_names`, `list_pyroscope_label_names`, `list_pyroscope_label_values`, `list_pyroscope_profile_types`, `list_snowflake_tables`, `list_team_roles`, `list_teams`, `list_user_roles`, `list_users_by_org`, `query_athena`, `query_clickhouse`, `query_cloudwatch`, `query_elasticsearch`, `query_graphite`, `query_graphite_density`, `query_influxdb`, `query_loki_logs`, `query_loki_patterns`, `query_loki_stats`, `query_prometheus`, `query_prometheus_histogram`, `query_pyroscope`, `query_snowflake`, `run_panel_query`, `search_dashboards`, `search_folders`, `suggest_loki_alloy_label_config`

## Confirmed with server team (2026-07-23)

- **Assistant tool name:** `ask_assistant` confirmed → `readOnlyHint: false, destructiveHint: true, openWorldHint: false`.
- **`openWorldHint` policy:** reserved for public plugin-catalog tools only — `search_plugin_information` and `install_plugin` → `true`. All other tools stay `false`, including the external cloud-datasource query tools (`query_cloudwatch`, `query_athena`, `query_snowflake`, `query_elasticsearch`, `query_influxdb`, `query_graphite*`), which are scoped to the user's configured datasources.
- **`get_plugin`:** `openWorldHint: false` — treated as an instance-scoped lookup (reads a plugin in the user's Grafana stack, not the public catalog). Flip to `true` only if the implementation actually queries the grafana.com catalog.

## Notes

- **Coverage:** the table + list above cover the ~85 tools OpenAI flagged. Please apply the "every tool, all three hints" rule to the full toolset, including any tools not listed here.
- **Nice-to-have (non-blocking):** add a per-tool `outputSchema` so models parse results better. The scan passes without it.

## Acceptance criteria

Re-running the OpenAI plugin-directory tool scan against `https://mcp.grafana.com/mcp` returns **zero** missing-annotation errors.

cedricziel:
Closes #1009.

## Summary

Every tool now explicitly sets the three MCP tool annotations — `readOnlyHint`, `destructiveHint`, `openWorldHint`. The OpenAI plugin-directory tool scan fails closed on any omission, and before this change 105 of 111 tools were missing at least one hint (most were missing `destructiveHint` and `openWorldHint`; the mcp-go `omitempty` pointer fields drop unset hints from `tools/list` entirely).

Annotation policy, as confirmed in #1009:

- **Pure-read tools** → `readOnlyHint: true, destructiveHint: false, openWorldHint: false`.
- **Additive writes** (`create_*`, `add_activity_to_incident`, `create_datasource`, `create_snapshot`) → `readOnlyHint: false, destructiveHint: false`.
- **Overwriting/deleting writes** (`update_*`, `delete_snapshot`, `alerting_manage_rules`, `grafana_api_request`, `ask_assistant`) → `destructiveHint: true`.
- **`openWorldHint: true` is reserved for the public plugin-catalog tools** (`install_plugin`, `search_plugin_information`, both of which call `grafana.com/api/plugins`). Everything else — including the cloud-datasource query tools — stays `false` since they are scoped to the user's stack and configured datasources.
- Write-gated read-only variants (`grafana_api_request`, `generate_deeplink`, `alerting_manage_rules`, the agento11y eval tools) are annotated per variant, so both `--disable-write` and write-enabled deployments advertise accurate hints.

## Value corrections beyond filling gaps

- `ask_assistant`: `openWorldHint` **true → false** — per the issue, open-world is reserved for public-catalog tools; the assistant is scoped to the user's stack.
- `install_plugin`: `destructiveHint` **true → false** — additive install, per the issue table.
- `find_error_pattern_logs` / `find_slow_requests` (Sift): `readOnlyHint` **true → false** — they POST new investigations to the Sift API, and are already write-gated in `AddSiftTools`. Marking a write-capable tool read-only is exactly what the scan (and the issue) forbids.
- `generate_deeplink` (write variant): `readOnlyHint: false` — with `shorten=true` it creates a Grafana short URL. This deviates from the issue's read-only list, which appears not to account for the short-URL write path; the read-only variant (registered under `--disable-write`) is `readOnlyHint: true`. Happy to revisit if we'd rather match the issue list verbatim.
- `query_clickhouse` / `query_snowflake` / `query_athena` / `query_influxdb`: `readOnlyHint` **true → false**, `destructiveHint` **false → true** (per review) — these pass the caller's raw SQL/InfluxQL/Flux through to `/api/ds/query` unfiltered, so `DELETE`/`DROP` (and Flux `to()`) execute whenever the datasource credentials permit it. This also deviates from the issue's read-only list, which didn't account for the raw-statement passthrough. `query_cloudwatch` stays read-only: it builds a structured `GetMetricData` request from namespace/metric/dimension parameters, with no statement language that could mutate anything. The remaining query tools (Loki, Prometheus, Elasticsearch, Graphite, Pyroscope, Quickwit) use read-only query languages/APIs and stay read-only.

## Regression guard

`cmd/mcp-grafana/annotations_test.go` registers every tool category from `toolEntries()` (the same single source of truth `main.go` uses) in both write modes, lists tools exactly as a client would via `tools/list`, and fails if any tool omits one of the three hints or claims `readOnlyHint: true` together with `destructiveHint: true`. A future tool shipped without hints now fails CI instead of silently delisting the hosted server.

## Notes

- Proxied tools (from app-plugin MCP servers) pass through whatever annotations the plugin declares; they are dynamic per session and out of scope here.
- `outputSchema` (the issue's non-blocking nice-to-have) is not included.

## Testing

- `make test-unit` — all green, including the new annotation test (was failing with 105 violations before the fix).
- `make lint` — clean.

### DIFF
--- a/tools/agento11y.go
+++ b/tools/agento11y.go
@@ -469,6 +469,8 @@ When NOT to use:
 	mcp.WithTitleAnnotation("Manage Agent Observability conversations"),
 	mcp.WithIdempotentHintAnnotation(true),
 	mcp.WithReadOnlyHintAnnotation(true),
+	mcp.WithDestructiveHintAnnotation(false),
+	mcp.WithOpenWorldHintAnnotation(false),
 )
 
 var ManageAgento11yGenerations = mcpgrafana.MustTool(
@@ -489,6 +491,8 @@ When NOT to use:
 	mcp.WithTitleAnnotation("Manage Agent Observability generations"),
 	mcp.WithIdempotentHintAnnotation(true),
 	mcp.WithReadOnlyHintAnnotation(true),
+	mcp.WithDestructiveHintAnnotation(false),
+	mcp.WithOpenWorldHintAnnotation(false),
 )
 
 const manageAgento11yEvaluatorsDescriptionFmt = `%s
@@ -658,6 +662,8 @@ var ManageAgento11yEvaluatorsRead = mcpgrafana.MustTool(
 	mcp.WithTitleAnnotation("Manage Agent Observability evaluators"),
 	mcp.WithIdempotentHintAnnotation(true),
 	mcp.WithReadOnlyHintAnnotation(true),
+	mcp.WithDestructiveHintAnnotation(false),
+	mcp.WithOpenWorldHintAnnotation(false),
 )
 
 var ManageAgento11yEvalRulesRead = mcpgrafana.MustTool(
@@ -667,6 +673,8 @@ var ManageAgento11yEvalRulesRead = mcpgrafana.MustTool(
 	mcp.WithTitleAnnotation("Manage Agent Observability eval rules and guards"),
 	mcp.WithIdempotentHintAnnotation(true),
 	mcp.WithReadOnlyHintAnnotation(true),
+	mcp.WithDestructiveHintAnnotation(false),
+	mcp.WithOpenWorldHintAnnotation(false),
 )
 
 var ManageAgento11yEvalCollectionsRead = mcpgrafana.MustTool(
@@ -676,6 +684,8 @@ var ManageAgento11yEvalCollectionsRead = mcpgrafana.MustTool(
 	mcp.WithTitleAnnotation("Manage Agent Observability saved conversations and collections"),
 	mcp.WithIdempotentHintAnnotation(true),
 	mcp.WithReadOnlyHintAnnotation(true),
+	mcp.WithDestructiveHintAnnotation(false),
+	mcp.WithOpenWorldHintAnnotation(false),
 )
 
 var ManageAgento11yEvaluatorsReadWrite = mcpgrafana.MustTool(
@@ -683,7 +693,9 @@ var ManageAgento11yEvaluatorsReadWrite = mcpgrafana.MustTool(
 	manageAgento11yEvaluatorsDescription(false),
 	manageAgento11yEvaluatorsReadWrite,
 	mcp.WithTitleAnnotation("Manage Agent Observability evaluators"),
+	mcp.WithReadOnlyHintAnnotation(false),
 	mcp.WithDestructiveHintAnnotation(true),
+	mcp.WithOpenWorldHintAnnotation(false),
 )
 
 var ManageAgento11yEvalRulesReadWrite = mcpgrafana.MustTool(
@@ -691,7 +703,9 @@ var ManageAgento11yEvalRulesReadWrite = mcpgrafana.MustTool(
 	manageAgento11yEvalRulesDescription(false),
 	manageAgento11yEvalRulesReadWrite,
 	mcp.WithTitleAnnotation("Manage Agent Observability eval rules and guards"),
+	mcp.WithReadOnlyHintAnnotation(false),
 	mcp.WithDestructiveHintAnnotation(true),
+	mcp.WithOpenWorldHintAnnotation(false),
 )
 
 var ManageAgento11yEvalCollectionsReadWrite = mcpgrafana.MustTool(
@@ -699,7 +713,9 @@ var ManageAgento11yEvalCollectionsReadWrite = mcpgrafana.MustTool(
 	manageAgento11yEvalCollectionsDescription(false),
 	manageAgento11yEvalCollectionsReadWrite,
 	mcp.WithTitleAnnotation("Manage Agent Observability saved conversations and collections"),
+	mcp.WithReadOnlyHintAnnotation(false),
 	mcp.WithDestructiveHintAnnotation(true),
+	mcp.WithOpenWorldHintAnnotation(false),
 )
 
 func AddAgento11yTools(mcp *server.MCPServer, enableWriteTools bool) {
--- a/tools/athena.go
+++ b/tools/athena.go
@@ -526,8 +526,11 @@ Example: SELECT request_time, status FROM my_table WHERE $__timeFilter(request_t
 	queryAthena,
 	mcp.WithTitleAnnotation("Query Athena"),
 	mcp.WithIdempotentHintAnnotation(true),
-	mcp.WithReadOnlyHintAnnotation(true),
-	mcp.WithDestructiveHintAnnotation(false),
+	// The raw SQL is passed through unfiltered, so DML/DDL (INSERT, DROP,
+	// CREATE TABLE AS) executes if the datasource credentials permit it —
+	// not read-only, potentially destructive.
+	mcp.WithReadOnlyHintAnnotation(false),
+	mcp.WithDestructiveHintAnnotation(true),
 	mcp.WithOpenWorldHintAnnotation(false),
 )
 
--- a/tools/athena.go
+++ b/tools/athena.go
@@ -525,7 +525,7 @@ Athena queries are async — Grafana handles polling. Use LIMIT and partition-aw
 Example: SELECT request_time, status FROM my_table WHERE $__timeFilter(request_time) LIMIT 100`,
 	queryAthena,
 	mcp.WithTitleAnnotation("Query Athena"),
-	mcp.WithIdempotentHintAnnotation(true),
+	mcp.WithIdempotentHintAnnotation(false),
 	// The raw SQL is passed through unfiltered, so DML/DDL (INSERT, DROP,
 	// CREATE TABLE AS) executes if the datasource credentials permit it —
 	// not read-only, potentially destructive.
--- a/tools/athena.go
+++ b/tools/athena.go
@@ -241,6 +241,8 @@ var ListAthenaCatalogs = mcpgrafana.MustTool(
 	mcp.WithTitleAnnotation("List Athena catalogs"),
 	mcp.WithIdempotentHintAnnotation(true),
 	mcp.WithReadOnlyHintAnnotation(true),
+	mcp.WithDestructiveHintAnnotation(false),
+	mcp.WithOpenWorldHintAnnotation(false),
 )
 
 type ListAthenaDatabasesParams struct {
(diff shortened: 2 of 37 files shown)


tools/agento11y.go
```diff
@@ -658,6 +662,8 @@ var ManageAgento11yEvaluatorsRead = mcpgrafana.MustTool(
 	mcp.WithTitleAnnotation("Manage Agent Observability evaluators"),
 	mcp.WithIdempotentHintAnnotation(true),
 	mcp.WithReadOnlyHintAnnotation(true),
+	mcp.WithDestructiveHintAnnotation(false),
```
gerboland, line 665: I see "delete_collection" tool in its prompt, so shouldn't this be true? I'm wondering if the readonly hint is correct now...

tools/clickhouse.go
```diff
@@ -240,6 +240,8 @@ Example: SELECT Timestamp, Body FROM otel_logs WHERE $__timeFilter(Timestamp)`,
 	mcp.WithTitleAnnotation("Query ClickHouse"),
 	mcp.WithIdempotentHintAnnotation(true),
 	mcp.WithReadOnlyHintAnnotation(true),
+	mcp.WithDestructiveHintAnnotation(false),
```
gerboland: this could be used destructively, no?

tools/cloudwatch.go
```diff
@@ -318,6 +318,8 @@ Cross-account monitoring: Use accountId to query metrics from a specific source
 	mcp.WithTitleAnnotation("Query CloudWatch"),
 	mcp.WithIdempotentHintAnnotation(true),
 	mcp.WithReadOnlyHintAnnotation(true),
+	mcp.WithDestructiveHintAnnotation(false),
```
gerboland, line 321: destructive?

tools/influxdb.go
```diff
@@ -218,6 +218,8 @@ Flux example:    from(bucket: "metrics") |> range(start: -1h) |> filter(fn: (r)
 	mcp.WithTitleAnnotation("Query InfluxDB"),
 	mcp.WithIdempotentHintAnnotation(true),
 	mcp.WithReadOnlyHintAnnotation(true),
+	mcp.WithDestructiveHintAnnotation(false),
```
gerboland: destructive?

tools/snowflake.go
```diff
@@ -241,6 +241,8 @@ Example: SELECT TIMESTAMP, RECORD['severity_text']::STRING AS LEVEL, VALUE FROM
 	mcp.WithTitleAnnotation("Query Snowflake"),
 	mcp.WithIdempotentHintAnnotation(true),
 	mcp.WithReadOnlyHintAnnotation(true),
+	mcp.WithDestructiveHintAnnotation(false),
```
gerboland: possibly destructive, DELETE can be done

gerboland (changes requested):
Some spots where generic queries can include DELETE clauses should probably be marked destructive? WDYT?

tools/clickhouse.go
cedricziel: Agreed — the handler passes `rawSql` through to `/api/ds/query` with no statement filtering, so a `DELETE`/`DROP` executes if the datasource credentials permit it. Flipped to `readOnlyHint: false, destructiveHint: true` in fa7236d. I applied the same to `query_athena`, which is the identical passthrough pattern but wasn't flagged.

tools/snowflake.go
cedricziel: Agreed — same unfiltered `rawSql` passthrough as ClickHouse. Flipped to `readOnlyHint: false, destructiveHint: true` in fa7236d.

tools/influxdb.go
cedricziel: Agreed — the query string is passed through unfiltered, InfluxQL supports `DELETE`/`DROP`, and Flux can write via `to()`. Flipped to `readOnlyHint: false, destructiveHint: true` in fa7236d.

tools/cloudwatch.go
cedricziel: I kept this one read-only intentionally: unlike the SQL tools there's no query language to smuggle a write through. The handler builds a structured metric query from `namespace`/`metricName`/`dimensions`/`statistic` parameters and sends it to the CloudWatch datasource, which backs it with read-only AWS APIs (`GetMetricData`/`ListMetrics`). There's no path to a mutation here, so `readOnlyHint: true, destructiveHint: false` is accurate. Happy to flip if you see a write path I'm missing.

tools/agento11y.go
cedricziel: This annotation is on the read-only variant, which can't reach any write: `ManageAgento11yEvalRulesReadParams.Operation` is enum-restricted to `list_rules`/`get_rule`/`list_guards`/`get_guard` and `validate()` rejects anything else, and its description says "This variant performs no writes" — `delete_collection`/`delete_rule` only appear in the ReadWrite variant's description. The ReadWrite variant (registered under the same tool name when write tools are enabled, which is what the hosted server serves) is annotated `readOnlyHint: false, destructiveHint: true`. So both variants should be accurate as-is.

tools/athena.go
```diff
@@ -518,7 +526,12 @@ Example: SELECT request_time, status FROM my_table WHERE $__timeFilter(request_t
 	queryAthena,
 	mcp.WithTitleAnnotation("Query Athena"),
 	mcp.WithIdempotentHintAnnotation(true),
-	mcp.WithReadOnlyHintAnnotation(true),
+	// The raw SQL is passed through unfiltered, so DML/DDL (INSERT, DROP,
+	// CREATE TABLE AS) executes if the datasource credentials permit it —
+	// not read-only, potentially destructive.
+	mcp.WithReadOnlyHintAnnotation(false),
+	mcp.WithDestructiveHintAnnotation(true),
```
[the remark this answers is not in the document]
cedricziel: Good catch — fixed in 3ed6ff6: `idempotentHint` is now `false` on all four write-capable query tools (`query_athena`, `query_clickhouse`, `query_snowflake`, `query_influxdb`), so clients won't treat them as retry-safe.

gerboland (approved):
Thanks for double-checking those bits for me!

PR #1052, opened 2026-08-06
fix(http): drop support for undocumented X-Grafana-URL

kolesnikovae:
Drop undocumented `X-Grafana-URL` support from HTTP transports and use the configured Grafana URL consistently.

Deprecate the related validation middleware and remove its README guidance.

**Testing:**
- `go test -count=1 -tags unit . ./tools ./cmd/mcp-grafana`
- `go run github.com/golangci/golangci-lint/v2/cmd/golangci-lint@v2.11.4 run . ./tools ./cmd/mcp-grafana`

### DIFF
--- a/README.md
+++ b/README.md
@@ -996,16 +996,16 @@ grafanaConfig := mcpgrafana.GrafanaConfig{
 contextFunc := mcpgrafana.ComposedStdioContextFunc(grafanaConfig)
 ```
 
-**URL validation when creating clients programmatically:**
+**URL validation:**
 
-When calling `NewGrafanaClient` directly, pre-validate untrusted URLs to avoid a reachable panic:
+When calling `NewGrafanaClient` directly (stdio or programmatic construction), pre-validate URLs to avoid a reachable panic:
 
 ```go
-if err := mcpgrafana.ValidateGrafanaURL(untrustedURL); err != nil {
+if err := mcpgrafana.ValidateGrafanaURL(urlFromHeader); err != nil {
     http.Error(w, err.Error(), http.StatusBadRequest)
     return
 }
-client := mcpgrafana.NewGrafanaClient(ctx, untrustedURL, apiKey, nil)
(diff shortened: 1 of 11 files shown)


gerboland (approved):

ioanarm (approved):
lgtm Anton! 🚀

uniquenode:
This is a breaking change for multi-instance setups. Support for `X-Grafana-URL` and `X-Grafana-API-Key` meant one deployment could serve multiple Grafana backends; a static `GRAFANA_URL` forces one deployment per backend. Could the header path be kept as an optional override?

kolesnikovae:
I'm afraid this exposes a security issue, [CVE-2026-15583](https://github.com/advisories/GHSA-7cfp-pmw3-4j7r) ([NIST](https://nvd.nist.gov/vuln/detail/CVE-2026-15583)). We'll look into the available alternatives. Could you share some details about your setup?

uniquenode:
@kolesnikovae Thank you for your response. We have multiple Grafana instances but only a single Grafana MCP instance. We reuse that Grafana MCP instance across multiple MCP servers by passing different `X-Grafana-URL` and `X-Grafana-API-Key` headers. This lets us interact with multiple Grafana instances through a single Grafana MCP instance, since the endpoint and API key are determined dynamically from the request context. Ideally, we'd be able to opt into this old behavior, avoiding the need to deploy a dedicated Grafana MCP instance for each Grafana instance.

BzSpi:
Hi @kolesnikovae
To answer your question about setups: mine matches @uniquenode's. We run around 150 separate Grafana instances, so one MCP deployment per instance isn't practical for us.

We plan to put the Grafana MCP behind our own "proxy" MCP server, which picks the target instance per request. I haven't found a workaround, because the Grafana URL can now only come from GRAFANA_URL.

I understand the concern behind CVE-2026-15583. Would an opt-in mode be acceptable, where a request that sets the Grafana URL must also bring its own credentials, and the environment credentials are never used for it ?

PR #1053, opened 2026-08-06
feat(auth): add browser session cookie authentication for stdio

ayush22667:
Adds a browser session cookie as a first-class auth mode, for SSO-only Grafana instances where minting a service account token is impractical.

Previously every credential had to be handed to the server (env var, header, or mounted file). The nearest workaround was pasting a cookie into GRAFANA_EXTRA_HEADERS, which goes stale within hours because Grafana rotates grafana_session server-side.

The server now resolves a session from a saved file, then from the grafana_session cookie held by installed browsers (via kooky), and re-acquires it automatically when Grafana rejects a request with 401/403 -- so the session survives rotation without a restart.

Notable details:

- Candidates are validated against /api/user, never /api/health. /api/health is unauthenticated, so a dead cookie passes it and then fails every real call. Redirects are not followed, so a 302 to a login page is not mistaken for success.
- Cookies rank last in AuthRoundTripper (after OBO, API key, basic auth) so they can never shadow an explicit credential, and are attached with AddCookie so a Cookie header set by ExtraHeaders survives alongside them.
- CookieProvider is carried into the stripped GrafanaConfig copies in NewGrafanaClient. Without that the OpenAPI client -- most of the tool surface -- would send no cookie while still compiling.
- Refresh is singleflighted: a burst of concurrent 401s triggers one browser scrape, not one per request.
- Only ever enabled on stdio. On a shared HTTP server the operator's personal cookie would authenticate every client as the operator, so the flag combination is refused at startup and ExtractGrafanaInfoFromHeaders clears the provider defensively.

Verified end to end against Grafana 13.0.2: a real grafana_session cookie authenticates a live list_datasources call, and --debug output shows the cookie redacted. The successful 401 -> refresh -> replay cycle has unit coverage only; proving it live needs a browser rotating a cookie mid-session.

### DIFF
--- a/README.md
+++ b/README.md
@@ -431,6 +431,7 @@ The `mcp-grafana` binary supports various command-line flags for configuration:
 **Grafana Client Options:**
 - `--grafana-timeout`: Time limit for requests made by the Grafana client. Accepts Go duration strings (e.g., `10s`, `500ms`) - default: `10s`
 - `--include-args-in-spans`: Include tool call arguments in OpenTelemetry spans. Only enable in non-production environments or when arguments are known not to contain PII - default: `false`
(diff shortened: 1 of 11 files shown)


ayush22667:
close

PR #1067, opened 2026-08-15
Add opt-in GCF output format for structured tool results (lossless, never-grow)

blackwell-systems:
### Summary

Adds an opt-in output encoding, `--output-format=gcf` (or `GRAFANA_OUTPUT_FORMAT=gcf`),
that serializes structured tool results as a [Graph Compact Format](https://gcformat.com)
generic wire instead of JSON in the model-facing text content block. The list tools
return arrays of uniform records (`list_datasources`, `list_teams`, `list_users_by_org`,
`alerting_manage_rules`, `list_incidents`, …); GCF factors their repeated field names into
a single header, so the per-record cost drops to the values and the result uses fewer
tokens when it crosses the LLM boundary.

It's applied at a single seam: the `MustTool` result serializer in `tools.go`, where every
typed handler return is marshaled. So it covers every record-returning tool with no
per-tool edits and no signature changes.

### Deliberately conservative: it can only ever help

The substitution is per-result and gated twice, so turning it on never regresses a result:

- **Never larger.** If the GCF wire is not smaller than the JSON for a given result, that
  result stays JSON (a never-grow guard). GCF wins on uniform record arrays and ties on
  tiny or high-entropy results, which simply keep JSON.
- **Never lossy.** Every substituted result is decoded and compared back to the original
  before it is used. On any encoding error, decode error, or mismatch (including an integer
  outside GCF's canonical int64 domain, which GCF rejects rather than approximating), the
  result falls back to JSON, so a tool call is never dropped or altered.
- **Opt-in and additive.** Default (`json`) output is unchanged, and `structuredContent`
  (where a tool sets it) is untouched.
- **Zero new footprint.** `gcf-go` is a zero-dependency package, pinned exact.

### Measured on real Grafana output

Captured from a live Grafana (the repo's `docker-compose` stack), real tool output, o200k
tokenizer, every result verified to round-trip back to the original value:

| Tool | records | tokens JSON → GCF |
|------|---------|-------------------|
| `list_datasources` | 57 | **−32.5%** |
| `list_users_by_org` | 31 | **−29.8%** |
| `list_teams` | 20 | **−28.1%** |
| `alerting_manage_rules` | 4 | −10.9% |
| `search_folders` | 3 | −12.4% |
| `search_dashboards` | 61 | ~tie → falls back to JSON via the never-grow guard |

The win is on the uniform-identifier inventory lists an agent hits constantly, and it
scales with record count (production instances are far larger than this test one).
High-entropy results (dashboard search, where each row carries a unique uid/url/title)
don't compress on tokens, so the guard keeps them on JSON. Nothing regresses.

### Why GCF, beyond size

The harder question than byte count is whether a model reads the compact form as
accurately as JSON. GCF is designed and evaluated for exactly that:

- **Generic-profile comprehension: 100% on every frontier model** (Claude
  Opus/Sonnet/Haiku, GPT-5.5, Gemini 2.5 Pro / 3.1 Pro / 3.5 Flash), across 27 runs, 11
  models, 4 providers; equal to or better than JSON on every model.
  https://gcformat.com/guide/benchmarks
- The grammar was reverse-engineered from attention-level research into how BPE
  tokenization shapes what a transformer can represent, then validated by controlled
  from-scratch training at 1.3B scale, measured causally rather than by an LLM judge, so it
  holds on models that have never seen the format. Four papers (2026, under review):
  - GCF: A Token-Optimized Wire Format for Structured LLM Interactions — https://doi.org/10.5281/zenodo.20579817
  - Tokenizer-Attention Coupling — https://doi.org/10.5281/zenodo.20925910
  - Stranded Attention — https://doi.org/10.5281/zenodo.21158886
  - Developmental Atlas of Attention Head Specialization — https://doi.org/10.5281/zenodo.21205389

### Already adopted

GCF is in production across observability, developer tooling, and network automation:

- **[Chrome DevTools MCP](https://github.com/ChromeDevTools/chrome-devtools-mcp)** (Google) merged it as an experimental output format.
- **[Speakeasy](https://speakeasy.com)** (customers include Google, Verizon, Mistral AI) ships it in their `oq` CLI.
- **[OmniRoute](https://omniroute.online)** vendored it into its AI gateway.
- **[NetClaw](https://github.com/automateyournetwork/netclaw)** benchmarked GCF against TOON on real network data and replaced TOON across its MCP servers.
- **[mcp-cisco-support](https://github.com/sieteunoseis/mcp-cisco-support)** (listed on Cisco Code Exchange) reports 28.5% fewer tokens on real Cisco Support API data.

…and many others across the MCP and agent ecosystem.

### The change

- `--output-format` flag + `GRAFANA_OUTPUT_FORMAT` env (`json` default, `gcf` opt-in),
  threaded through `GrafanaConfig` like the other server settings.
- One `encodeGCF` helper (never-grow guard + fail-safe round-trip) hooked into the
  `MustTool` JSON seam.
- Tests: unit coverage of the never-grow guard, losslessness, and exact int64
  preservation (a value above 2^53 is never float-rounded); a robustness table
  over nested/delimiter/quote/unicode/numeric/null/empty shapes; a fuzz target
  asserting the never-grow + lossless invariant on arbitrary JSON; the full
  `MustTool` handler path; and an integration e2e that runs the real
  `list_datasources` tool with GCF enabled against a live Grafana. README
  documents the flag.

GCF is open source and MIT-licensed. Format, spec, SDKs, and benchmarks: https://gcformat.com
Happy to adjust the flag shape or defaults, and to re-run the numbers against any Grafana
you point me at. Reproducible harness on request; independent verification welcome.

### DIFF
--- a/README.md
+++ b/README.md
@@ -445,6 +445,9 @@ Caller authentication is enforced only when `--server-auth-token` is set. When i
 - `--grafana-timeout`: Time limit for requests made by the Grafana client. Accepts Go duration strings (e.g., `10s`, `500ms`) - default: `10s`
 - `--include-args-in-spans`: Include tool call arguments in OpenTelemetry spans. Only enable in non-production environments or when arguments are known not to contain PII - default: `false`
 
+**Response Encoding:**
+- `--output-format`: Serialization for structured tool results in the model-facing text block: `json` (default) or `gcf`. Also settable via the `GRAFANA_OUTPUT_FORMAT` environment variable. See [Response Format](#response-format-gcf).
+
 **Observability:**
 - `--metrics`: Enable Prometheus metrics endpoint at `/metrics`
 - `--metrics-address`: Separate address for metrics server (e.g., `:9090`). If empty, metrics are served on the main server
@@ -574,6 +577,18 @@ volumes:
 ```
 
 Surrounding whitespace (including a trailing newline) is trimmed from the file contents. If both `GRAFANA_SERVICE_ACCOUNT_TOKEN` and `GRAFANA_SERVICE_ACCOUNT_TOKEN_FILE` are set, the inline token takes precedence.
+
+### Response Format (GCF)
+
+By default, structured tool results are serialized as JSON in the text content block the model reads. Setting `--output-format=gcf` (or `GRAFANA_OUTPUT_FORMAT=gcf`) instead offers each result as [Graph Compact Format](https://gcformat.com), a token-optimized encoding that factors the repeated field names out of uniform record arrays (the shape returned by list tools like `list_datasources`, `list_teams`, `list_users_by_org`, `alerting_manage_rules`, and `list_incidents`).
+
+The substitution is applied **per result and only when it helps**:
+
+- **Never larger.** If GCF is not smaller than the JSON for a given result, that result stays JSON. On real Grafana output this yields roughly 28-33% fewer tokens on uniform inventory lists (datasources, teams, users) at realistic sizes, and a no-op on tiny or high-entropy results.
+- **Never lossy.** Every substituted result is decoded and compared back to the original before it is used; on any encoding error (including an integer outside GCF's canonical int64 domain) the result falls back to JSON, so a tool call is never dropped or altered.
+- **Opt-in and additive.** The default (`json`) is unchanged, and `structuredContent` (where a tool sets it) is untouched.
+
+GCF is MIT-licensed and its `gcf-go` encoder is zero-dependency. It is designed and evaluated for how accurately models read the compact form, not just for size; format, spec, SDKs, and benchmarks are at [gcformat.com](https://gcformat.com).
 
 ### Multi-Organization Support
 
--- a/cmd/mcp-grafana/main.go
+++ b/cmd/mcp-grafana/main.go
@@ -140,6 +140,10 @@ type grafanaConfig struct {
 
 	// timeout is the time limit for requests made by the Grafana client.
 	timeout time.Duration
+
+	// outputFormat selects how structured tool results are serialized for the
+	// model: "json" (default) or "gcf".
+	outputFormat string
 }
 
 func (dt *disabledTools) addFlags() {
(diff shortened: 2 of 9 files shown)


sd2k:
Hi, thanks for the PR! I think we will wait until this format is more established and perhaps included in some models' training sets before adding it. Happy to revisit in a few months or so in such a case!

blackwell-systems:
Thanks, and I fully respect the decision, no push to reverse it.

Worth flagging one thing on the "wait for training sets" reasoning, because the data speaks directly to it: training models on GCF turns out to be unnecessary. That's the exact variable the research isolates, and it cuts the other way. The comprehension results aren't LLM-judged (which could just be measuring familiarity), they're measured causally from attention structure and controlled training: models comprehend clean-delimiter structure ~2.3x better on data never seen during training, because BPE merge decisions permanently shape which attention heads can attend to boundaries at all. Downstream, GCF isn't in any current model's training data, yet holds 100% comprehension on every frontier model across 11 models from 4 providers, no accuracy cost versus JSON, at 50-69% fewer tokens. That robustness holds under adversarial conditions too: comprehension stays at 90.7% on adversarial code graphs and 100% on standard workloads.

The four companion papers (currently under review) lay out the format and the mechanism:
- GCF: A Token-Optimized Wire Format for Structured LLM Interactions — https://doi.org/10.5281/zenodo.20579817
- Tokenizer-Attention Coupling: How BPE Merge Decisions Permanently Shape Transformer Internal Organization — https://doi.org/10.5281/zenodo.20925910
- Stranded Attention: BPE Tokenization Permanently Constrains Transformer Structural Capacity — https://doi.org/10.5281/zenodo.21158886
- Developmental Atlas of Attention Head Specialization — https://doi.org/10.5281/zenodo.21205389

Benchmarks and harness are public if it's useful to sanity-check: https://gcformat.com/guide/benchmarks

Either way, it's opt-in and off by default, so a no-op for existing users if it ever lands. Happy to leave it with you and reopen whenever you'd like another look.

PR #1073, opened 2026-08-18
fix(metrics): make client cache and session OTel meters injectable

### LINKED ISSUE #1072: OTel metrics (client cache, session, proxied-tools discovery) are silently no-op in the hosted deployment
## Problem

`client_cache.go`'s `clientCacheMetrics`, `session.go`'s `sessionMetrics`, and `proxied_tools.go`'s `discoveryMetrics` (added in #1071) all construct their OTel instruments via `otel.GetMeterProvider().Meter(...)` — i.e. they read whatever the *global* MeterProvider happens to be at construction time.

grafana-assistant-app (which imports this repo as a library and runs the hosted MCP server) has an `init()` in `api/internal/tracing/http.go` that unconditionally does:

```go
otel.SetMeterProvider(noop.MeterProvider{})
```

This is deliberate — it dodges a known otelhttp memory-leak bug (https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5190) — and runs before any mcp-grafana code executes, in the same process as the hosted MCP server. As a result, every `Add()`/`Record()` call these three metric structs make is silently discarded in that deployment. They only work in the standalone `mcp-grafana` CLI when run with `--metrics` (which calls `observability.Setup` in `observability/observability.go`, the only place in this repo that installs a real MeterProvider).

grafana-assistant-app doesn't use OTel metrics at all for its own instrumentation — it uses `prometheus/client_golang` + `promauto` directly against `prometheus.DefaultRegisterer` (e.g. `api/internal/cloudmcp/metrics.go`), scraped at `/metrics` via `promhttp.HandlerFor`. Nothing in mcp-grafana feeds that registry today.

## Proposed fix

Make the meter injectable instead of pulled from the ambient global:
- Add a way to pass a `metric.Meter` (or a `metric.MeterProvider`) into `NewToolManager` / the client cache / `SessionManager` constructors, defaulting to `otel.GetMeterProvider().Meter(...)` for backward compatibility (so the standalone CLI/`observability.Setup` path keeps working unchanged).
- In grafana-assistant-app, construct a private, non-global `sdkmetric.MeterProvider` wired to `prometheus.DefaultRegisterer` (mirroring the existing `sigilMeter` pattern in `api/cmd/dash-api/main.go`, which explicitly avoids calling `otel.SetMeterProvider()` to preserve the noop workaround) and pass its `Meter` into mcp-grafana's constructors from `hosted/mcp.go`'s `NewMCPServer`.

This fixes all three metric groups (client cache, session, discovery) at once, since they all follow the same pattern.

## Context

Surfaced while reviewing #1071 (proxied-tools discovery reliability fix), which added `discoveryMetrics` following the exact existing convention — only to discover that convention doesn't actually reach a scrapeable registry in the real hosted deployment. Deliberately scoped out of #1071 to keep that PR small; tracking here as a follow-up.

sd2k:
## Summary

Fixes #1072.

- `clientCacheMetrics` (`client_cache.go`) and `sessionMetrics` (`session.go`) previously built their OTel instruments via `otel.GetMeterProvider().Meter(...)`, reading whatever the *global* MeterProvider happens to be at construction time.
- In the hosted deployment (grafana-assistant-app), the global MeterProvider is deliberately reset to a noop (to dodge an [otelhttp memory-leak bug](https://github.com/open-telemetry/opentelemetry-go-contrib/issues/5190)) before mcp-grafana code runs, so these metrics were silently discarded there.
- Adds `WithClientCacheMeterProvider` and `WithSessionMeterProvider` functional options so a caller can inject an explicit `metric.MeterProvider`. Both default to `otel.GetMeterProvider()` when unset, so existing behavior is unchanged for callers that don't opt in.
- Adds `Observability.MeterProvider()` so the standalone CLI (and other embedders) can pass the provider from `observability.Setup` through explicitly, rather than relying on the implicit global. Wired into `cmd/mcp-grafana/main.go`.

## Out of scope

`proxied_tools.go`'s `discoveryMetrics`, mentioned in #1072, was added in #1071, which is still an unmerged draft PR — it doesn't exist on `main` yet, so there's nothing to fix there right now. It should adopt the same `WithXxxMeterProvider` pattern once #1071 lands.

The grafana-assistant-app side (constructing a private `sdkmetric.MeterProvider` wired to `prometheus.DefaultRegisterer` and passing it into `NewMCPServer`) lives in a separate repo and is not part of this PR.

## Testing

- Added `TestClientCache_MetricsUseInjectedMeterProvider` and `TestSessionManager_MetricsUseInjectedMeterProvider`, each using an `sdkmetric.ManualReader`-backed provider to assert the injected provider (not the global one) actually receives the recorded instruments.
- `go build ./...`, `go vet ./...`, `make test-unit`, and `make lint` all pass.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (doc comments on new exported options/getter)
- [x] Linting passes

### DIFF
--- a/client_cache.go
+++ b/client_cache.go
@@ -77,8 +77,32 @@ type clientCacheMetrics struct {
 	size    metric.Int64Gauge   // Current number of cached clients
 }
 
-func newClientCacheMetrics() clientCacheMetrics {
-	meter := otel.GetMeterProvider().Meter(clientCacheMeterName)
+// clientCacheConfig holds configuration gathered from ClientCacheOptions.
+type clientCacheConfig struct {
+	meterProvider metric.MeterProvider
+}
+
+// ClientCacheOption configures a ClientCache created via NewClientCache.
+type ClientCacheOption func(*clientCacheConfig)
+
+// WithClientCacheMeterProvider sets the metric.MeterProvider used to create
+// the client cache's OTel instruments. If unset (or passed as nil), the
+// cache falls back to otel.GetMeterProvider(), matching the pre-existing
+// behavior. Callers embedding mcp-grafana as a library and running with a
+// non-global MeterProvider (e.g. because the process resets the global
+// provider to a noop for unrelated reasons) should pass their own provider
+// here so these metrics actually reach a scrapeable registry.
+func WithClientCacheMeterProvider(mp metric.MeterProvider) ClientCacheOption {
+	return func(c *clientCacheConfig) {
+		c.meterProvider = mp
+	}
+}
+
+func newClientCacheMetrics(mp metric.MeterProvider) clientCacheMetrics {
+	if mp == nil {
+		mp = otel.GetMeterProvider()
+	}
+	meter := mp.Meter(clientCacheMeterName)
 
 	lookups, _ := meter.Int64Counter("mcp.client_cache.lookups",
 		metric.WithDescription("Total number of client cache lookups"),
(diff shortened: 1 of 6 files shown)


sd2k:
cursor review

svennergr (approved):

PR #1081, opened 2026-08-19
feat(oncall): return the last alert payload from get_alert_group

### LINKED ISSUE #814: Feature Request: Expose last_alert payload in get_alert_group response

sd2k:
`get_alert_group` now returns the group's most recent alert along with its raw payload. Alert payloads carry integration-specific fingerprints — Sentry's `payload.data.event.hashes[0]`, Alertmanager's `payload.alerts[0].fingerprint` — which let an agent recognise that an alert group is a recurrence of one it has already seen. Before this, the tool returned `alerts_count` and no way to reach the payload without another call.

Closes #814. Supersedes #828.

## Attribution

Requested in #814 by @OshriBay and first implemented in #828 by @Genmin — thanks to both. This is a reimplementation on a fresh branch rather than a rebase, because #828 is blocked on an unsigned CLA, which cannot be resolved in place. The code and tests here follow that PR closely; the differences are noted under Implementation.

## Implementation

`amixr-api-go-client`'s `AlertGroup` has no `LastAlert` field, and its `Alert` types the payload as a fixed `AlertPayload` struct that would discard exactly the source-specific keys this feature is for. The public API path therefore issues the `alert_groups/{id}/` request directly through `client.NewRequest`/`client.Do` instead of `aapi.AlertGroupService`, decoding into a local struct that mirrors the API response and holds the payload as `map[string]any`. Path construction, including `url.PathEscape`, mirrors what the library's own `GetAlertGroup` does, and the response is mapped field by field onto `OnCallAlertGroup` so everything the tool returned before is unchanged.

Two deliberate departures from #828:

- `LastAlert` hangs off the shared `OnCallAlertGroup` type in `oncall_types.go` rather than a new `GetAlertGroupResult` wrapper. The JSON on the wire is identical — the wrapper embedded `OnCallAlertGroup`, so its fields were inlined either way — but this keeps one alert-group type in the package and matches how that file already unifies the proxy and public API shapes. `list_alert_groups` is unaffected: it builds its results explicitly and `last_alert` is `omitempty`.
- The tool description now says what the payload is useful for, so an agent knows the fingerprints are there without having to fetch a group first.

There is no new parameter. #814 proposed an opt-in `includeLastAlertPayload` flag; per the review on #828, always returning the payload is better than paying for a flag and its description in every conversation, since a user asking about an alert group usually wants to see what fired. #828 was already updated to drop the flag before it stalled.

## Deliberate omissions

- **The IRM proxy path returns no `last_alert`.** When on-behalf-of tokens are present, `get_alert_group` goes through the IRM plugin proxy to the internal `alertgroups/{pk}/` endpoint, whose response (`onCallAlertGroupInternal`) has no `last_alert` field — so in Grafana Cloud this feature is currently a no-op and the tool returns the regular shape. Wiring it up needs confirmation of whether that internal endpoint exposes the payload at all, and under what shape; it is a follow-up, not something to guess at here. `last_alert` is `omitempty`, so the proxy response is byte-identical to today's.
- **No `list_alerts` tool.** #814 mentions one as a separate proposal for reaching a group's full alert history. Out of scope.

## Testing

- `make test-unit` — full unit suite green, including four new tests in `tools/oncall_unit_test.go`: nested payload survives the round trip (asserting the Sentry `hashes[0]` value specifically), `last_alert` absent means the key is omitted rather than `null`, the API error path is wrapped with the alert group ID, and the group ID is escaped into the path.
- `make lint-jsonschema`, `make lint-openapi`, `golangci-lint run`, `gofmt -l .` — all clean.
- Listed `tools/list` over stdio and inspected the registered `get_alert_group` definition: input schema is unchanged, so no caller sees a new or altered parameter.
- Not tested against a live OnCall instance. The integration suite runs against shared services with fixtures other tests assert on, and the cloud suite needs credentials; the payload shape is exercised through an `httptest` server that also stands in for the plugin settings lookup.

## Docs

The `get_alert_group` rows in `README.md` and `docs/sources/reference/mcp-tools-table.md` are unchanged — those tables carry a one-line summary and required scope, both still accurate, and the tables are column-aligned so widening one cell would rewrite every row.

### DIFF
--- a/tools/oncall.go
+++ b/tools/oncall.go
@@ -5,6 +5,7 @@ import (
 	"encoding/json"
 	"fmt"
 	"net/http"
+	"net/url"
 	"reflect"
 	"strings"
 
@@ -545,15 +546,40 @@ func getAlertGroup(ctx context.Context, args GetAlertGroupParams) (*OnCallAlertG
 	return amixrGetAlertGroup(ctx, args.AlertGroupID)
 }
 
+// oncallAlertGroupPublic mirrors the public OnCall API's alert group response.
+//
+// It exists because aapi.AlertGroup omits last_alert, and aapi.Alert types its
+// payload as a fixed struct that would discard integration-specific fields. The
+// request below therefore bypasses aapi.AlertGroupService and decodes into this
+// type instead.
+type oncallAlertGroupPublic struct {
+	ID             string            `json:"id"`
+	IntegrationID  string            `json:"integration_id"`
+	AlertsCount    int               `json:"alerts_count"`
+	State          string            `json:"state"`
+	CreatedAt      string            `json:"created_at"`
+	ResolvedAt     string            `json:"resolved_at"`
+	AcknowledgedAt string            `json:"acknowledged_at"`
+	Title          string            `json:"title"`
+	Permalinks     map[string]string `json:"permalinks"`
+	LastAlert      *OnCallLastAlert  `json:"last_alert"`
+}
+
 func amixrGetAlertGroup(ctx context.Context, alertGroupID string) (*OnCallAlertGroup, error) {
 	client, err := oncallClientFromContext(ctx)
 	if err != nil {
 		return nil, fmt.Errorf("getting OnCall client: %w", err)
 	}
 
-	alertGroupService := aapi.NewAlertGroupService(client)
-	ag, _, err := alertGroupService.GetAlertGroup(alertGroupID)
+	// Mirrors aapi.AlertGroupService.GetAlertGroup's path construction.
+	path := fmt.Sprintf("alert_groups/%s/", url.PathEscape(alertGroupID))
+	req, err := client.NewRequest(http.MethodGet, path, nil)
 	if err != nil {
+		return nil, fmt.Errorf("creating OnCall alert group request %s: %w", alertGroupID, err)
+	}
+
+	var ag oncallAlertGroupPublic
+	if _, err := client.Do(req, &ag); err != nil {
 		return nil, fmt.Errorf("getting OnCall alert group %s: %w", alertGroupID, err)
 	}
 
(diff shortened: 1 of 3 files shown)


gerboland (approved):
lgtm

PR #1085, opened 2026-08-20
feat(tools): add --disable-query, and gate raw-SQL query tools behind --disable-write

sd2k:
## Summary

Two related changes to how tool exposure is gated.

**1. A read-only-mode hole.** `query_clickhouse`, `query_snowflake`, `query_athena` and `query_influxdb` send the query to the datasource unfiltered. The code says so, and all four carry `WithReadOnlyHintAnnotation(false)`:

```go
// tools/clickhouse.go
// The raw SQL is passed through unfiltered, so a DELETE/DROP executes if
// the datasource credentials permit it — not read-only, potentially destructive.
```

None of them were gated behind `--disable-write`, so a server started in read-only mode could still `DROP TABLE` through a datasource. This PR treats those four as write tools as well as query tools, and adds `--enable-query` to put them back for operators whose datasource credentials are themselves read-only.

**2. A new `--disable-query` flag** that removes every tool executing a query against a datasource, while keeping the metadata and discovery tools. For an assistant that should be able to explore what exists — datasources, dashboards, metric names, labels, table schemas — without running potentially expensive or data-revealing queries.

## Behaviour

| Flags | Safe query tools<br>(`query_prometheus`, `query_loki_logs`, `run_panel_query`, …) | Raw-SQL query tools<br>(`query_clickhouse`, `query_snowflake`, `query_athena`, `query_influxdb`) | Write tools |
| --- | --- | --- | --- |
| _(none)_ | registered | registered | registered |
| `--disable-write` | registered | **not registered** ⚠️ *changed* | not registered |
| `--disable-write --enable-query` | registered | registered | not registered |
| `--disable-query` | not registered | not registered | registered |
| `--disable-query --enable-query` | not registered | not registered | registered |

`--disable-query` is the strongest setting and wins over `--enable-query` (the server logs a warning if both are passed). `--enable-query` on its own is a no-op, and never re-enables a non-query write tool.

⚠️ **Behaviour change:** anyone running `--disable-write` with `clickhouse`, `snowflake`, `athena`, or `influxdb` enabled loses those four query tools unless they add `--enable-query`. That's the point of the change, but it's worth a release note.

## Changes

- `--disable-query`: gates `query_prometheus`, `query_prometheus_histogram`, `query_loki_logs`, `query_loki_patterns`, `query_elasticsearch`, `query_quickwit`, `query_influxdb`, `query_clickhouse`, `query_snowflake`, `query_athena`, `query_graphite`, `query_graphite_density`, `query_cloudwatch`, `query_pyroscope`, `run_panel_query`. Every `list_*` / `describe_*` sibling stays registered.
- `--enable-query`: overrides the write gate on the four raw-SQL tools only.
- `mutatingQueryCategories` in `cmd/mcp-grafana/main.go` is the single list of categories whose query tools can write; `queryToolsEnabled(category)` resolves all three flags and drives both registration and instructions.
- Server instructions follow registration: `elasticsearch`, `quickwit`, `influxdb`, and `runpanelquery` contain only query tools and drop out of the capability list entirely when gated; partially gated categories get a "Query execution is disabled" description instead of one promising PromQL/LogQL.
- Docs: README (read-only mode + new "Query-Free Mode" section), both `docs/sources/configure/` pages, tools-table reference, CONTRIBUTING convention, PR-template checklist.

`run_panel_query` is deliberately **not** in the mutating class: if a dashboard panel holds a `DELETE`, the problem is upstream of this server.

## Testing

- `make lint` and `go test -tags unit ./...` pass. First push was fully green on CI (25/25, including Test Cloud and all three Python E2E suites).
- `cmd/mcp-grafana/main_test.go`: safe query tools survive `--disable-write`; raw-SQL ones don't; `--enable-query` restores exactly those four and no other write tool; `--enable-query` alone is a no-op; `--disable-query` beats `--enable-query`; both disable flags together leave only metadata tools. Plus `buildInstructions` cases for each state.
- Manual smoke test over stdio across all five flag combinations, checking `tools/list` for `query_prometheus`, `query_clickhouse`, `query_influxdb`, `run_panel_query`, `list_clickhouse_tables`, and `update_dashboard`.

## Notes for reviewers

- `query_loki_stats` and `analyze_loki_labels` are **not** gated: both send a selector to the datasource, but they read the index and return stream/chunk/byte counts, never log content. The line is "returns data", so only `query_loki_logs` and `query_loki_patterns` are gated on the Loki side.
- `--disable-query` gates the query tools; it doesn't police every path to a datasource. `grafana_api_request` (already GET-only under `--disable-write`) and `get_panel_image` are deliberately unaffected — called out in the docs.
- **Breaking for direct library consumers:** twelve exported `tools.Add*Tools` functions gain a bool parameter, the same shape the write gate uses. In-repo, `cmd/mcp-grafana` is the only caller.
- No new tools, so no token-cost change.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated
- [x] Linting passes


sd2k:
cursor review

sd2k:
cursor review

sd2k:
cursor review

sd2k:
cursor review

ioanarm (approved):

PR #1086, opened 2026-08-20
fix(observability): install a global TextMapPropagator so distributed traces connect

### LINKED ISSUE #1084: Distributed traces break over HTTP transports: no global TextMapPropagator is set
### What happened?

When mcp-grafana runs with an HTTP transport (streamable-http/sse) and OTLP tracing enabled, it doesn't participate in W3C trace-context propagation over HTTP, so distributed traces break on both sides of the server:
  - Inbound: an incoming traceparent/tracestate header (e.g. injected by an upstream proxy) is ignored — mcp-grafana starts a brand-new trace instead of continuing the caller's.
  - Outbound: requests to the Grafana API carry no traceparent, so Grafana starts yet another unrelated trace.

  **Expected:** mcp-grafana continues an inbound W3C traceparent and injects one on outbound Grafana calls ; caller → mcp-grafana → Grafana is one connected trace.
  **Actual:** three disjointed traces.

### Steps to reproduce

1. Run with tracing over HTTP:
```bash
     OTEL_EXPORTER_OTLP_ENDPOINT=http://<collector>:4317 OTEL_EXPORTER_OTLP_INSECURE=true \
     mcp-grafana -t streamable-http --address 127.0.0.1:8000
```
2. Send a request carrying a W3C traceparent (any tool call reaching the Grafana API):
```bash
     curl -H 'traceparent: 00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01' http://127.0.0.1:8000/mcp ...
```
3. Inspect the span mcp-grafana exports for this request.
 
**Observed:** the exported span has a fresh trace ID (not the inbound one), and the outbound Grafana request has no traceparent.

Likely missing `otel.SetTextMapPropagator`/`otelhttp.WithPropagators` on handlers/transports. Or missing `go.opentelemetry.io/contrib/propagators/autoprop` to honour `OTEL_PROPAGATORS`.

`GRAFANA_FORWARD_HEADERS` set to `traceparent,tracestate` works partially but skips mcp-grafana in a trace. mcp-grafana still emits its own distinct trace.

### mcp-grafana version

v1.1.0

### Grafana version

12.3.10, self-hosted

### MCP client

Claude Code

### Transport

streamable-http

### How did you install it?

Docker image

### Operating system and CPU architecture

macOS, arm64

### Logs

```text

```

sd2k:
## Summary

Fixes #1084. Under an HTTP transport, mcp-grafana did not take part in W3C trace-context propagation, so one logical request produced three disjointed traces (caller, mcp-grafana, Grafana).

The root cause is a single missing call: `otel.SetTextMapPropagator` was never made, leaving OTel's no-op default installed. Both `otelhttp` wrappers were already wired up — `observability.WrapHandler` on the server side, `otelhttp.NewTransport` in `mcpgrafana.BuildTransport` — but each consults the global propagator, so the handler ignored an inbound `traceparent` and the client transport injected nothing.

## Changes

- `observability.Setup` now installs a propagator via `contrib/propagators/autoprop`, which honours the standard `OTEL_PROPAGATORS` env var and defaults to `tracecontext,baggage`.
  - Installed **unconditionally**, not only when `OTEL_EXPORTER_OTLP_ENDPOINT` is set: with no tracer provider the extracted span context still rides through the non-recording spans and is injected outbound, so a server with tracing off passes the caller's trace context through to Grafana instead of dropping it.
- `ExtraHeadersRoundTripper` no longer overwrites trace-context headers the propagator has already injected. A `traceparent` copied from the incoming request via `GRAFANA_FORWARD_HEADERS` names the *caller's* span, so applying it would re-parent Grafana's spans onto the caller and cut mcp-grafana out of the middle of the trace — the half-working state described in the issue. The protected header names come from the propagator's own `Fields()`, so they follow `OTEL_PROPAGATORS`; forwarded values still apply when nothing was injected.
- Docs: new "Trace context propagation" section covering both directions, `OTEL_PROPAGATORS`, the `_meta` channel used under stdio, and the `GRAFANA_FORWARD_HEADERS` interaction; plus a cross-reference from the README's forwarding section. CHANGELOG entry under Unreleased → Fixed.
- Dependency: `go.opentelemetry.io/contrib/propagators/autoprop` v0.68.0, pinned to the contrib line matching the otel v1.43.0 already in `go.mod` so nothing else moves.

## Testing

`make lint` clean, `go vet ./...` clean, `go test -tags unit ./...` passing.

New unit tests:
- `observability/propagation_test.go` — `Setup` installs a propagator that extracts and injects W3C trace context; `OTEL_PROPAGATORS` is honoured (`b3`, `none`); `WrapHandler` parents its server span on an inbound `traceparent`, with the pre-fix no-op behaviour pinned as a contrast case.
- `trace_propagation_test.go` — `BuildTransport` injects a `traceparent` continuing the caller's trace (and does not under `WithoutOtel()`); extra/forwarded trace headers, whether from config or per-request context, do not override an injected value, but do apply when nothing was injected.

Also verified end-to-end with the real binary against a stub Grafana that logs request headers. With tracing enabled and an inbound `traceparent` for trace `0af7651916cd43dd8448eb211c80319c`:

```
inbound:   00-0af7651916cd43dd8448eb211c80319c-b7ad6b7169203331-01
outbound:  GET /api/search traceparent="00-0af7651916cd43dd8448eb211c80319c-ea72dab70faa6043-01"
```

Same trace, new span ID — mcp-grafana is now a real hop in the middle rather than a separate trace. On `main` the outbound `traceparent` is absent entirely.

## Notes for reviewers

Out of scope, flagged rather than fixed: `_meta`-based propagation in `extractTraceContext` still takes precedence over the HTTP-level parent when a tool call carries both. Trace IDs match in practice (both come from the same client), but the tool span parents to the client's span rather than nesting under our own server span. Making the HTTP parent win would require distinguishing "server span with a remote parent" from "server span that is a trace root", which isn't recoverable from a `SpanContext` alone — so I left the existing behaviour untouched and documented the channel instead.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated
- [x] Linting passes


sd2k:
cursor review

gerboland (approved):
nice

PR #1088, opened 2026-08-20
fix(navigation): emit Explore deeplinks in the panes format

### LINKED ISSUE #300: Does the generate_deeplink tool work with version 12.1?
All the links i generate are invalid, if i manually go do the same via Explore, the links i get look nothing like what comes from the tool for the same query. 

I can get it to work if i post an example of a working link, but until i do that it looks like it's generating for a different version of grafana.

DrFaust92:
Fixes #300.

## Problem

`generate_deeplink` builds Explore links with the pre-10.2 `left` parameter:

```go
params.Set("left", string(leftJSON))
```

Grafana 10.2 moved Explore URL state to `panes` (keyed by pane id) plus `schemaVersion`. Links using `left` still load, but they go through a migration that is lossy for datasource-specific query models, so the pane opens with the query silently dropped.

A Cloud Monitoring (`stackdriver`) PromQL query shows it plainly. This goes in:

```json
{"refId":"A","queryType":"promQL","promQLQuery":{"expr":"...","projectName":"...","step":"300s"}}
```

and Grafana rewrites it to:

```json
{"refId":"A","queryType":"timeSeriesList","timeSeriesList":{"promQLQuery":{"expr":"..."},"projectName":"","filters":[],"view":"FULL"}}
```

`queryType` is reset to the default and `promQLQuery` ends up nested inside `timeSeriesList`, where nothing reads it — the editor opens in Builder mode with every field empty and "No data". #300 also reports Loki links failing outright with `TypeError: Cannot read properties of undefined (reading 'datasourceInstance')`.

## Change

Emit the current format:

```
/explore?schemaVersion=1&panes={"a":{"datasource":"<uid>","queries":[...],"range":{...}}}
```

Each query also gets a `datasource` reference, as the pane format expects. A caller-supplied one is preserved rather than overwritten, because it carries the plugin `type` that some query editors need to select the right editor mode — that is what makes the Cloud Monitoring case above work.

Verified against Grafana 12.x: the same PromQL query that previously opened an empty Builder pane now round-trips and executes.

## Notes for reviewers

- The pane id is a fixed `"a"`. Grafana uses short random ids in the UI; a stable key is fine for a single-pane link, but say the word if you would rather it be randomised.
- Resolving the datasource *type* from the uid (via the client already in context) would be a better default than the uid-only fallback, and would remove the need for callers to pass `datasource` themselves. Left out to keep this focused — happy to add it here or separately.
- `shortenURL`'s test still uses a `left=` URL as its input fixture. That is unrelated to the format this tool emits, so I left it alone.

## Tests

`tools/navigation_test.go` updated for the new shape, plus two additions: one asserting `left` is never emitted, and one asserting a datasource-specific query model (`queryType: "promQL"` with a `stackdriver` datasource) survives into the URL intact. `go test ./tools/` passes.

### DIFF
--- a/tools/navigation.go
+++ b/tools/navigation.go
@@ -379,13 +379,6 @@ func buildDashboardTargetURL(baseURL string, dashboardUID *string, preview *Deep
 	return target, nil
 }
 
-// toGrafanaTimeParam converts a time value to a format Grafana understands
-// in URL query parameters. Grafana's Scenes parseUrlParam uses hardcoded
-// string length checks and only recognizes ISO 8601 at exactly 24 chars
-// (with milliseconds, e.g. "2026-04-28T12:45:00.000Z"). Shorter ISO 8601
-// strings like "2026-04-28T12:45:00Z" (20 chars) are silently ignored.
-// This function converts RFC 3339 timestamps to epoch milliseconds, which
-// is universally supported. Relative strings and epoch values pass through.
 // supportsExplorePanes reports whether a Grafana version reads Explore state
 // from `panes`. Grafana 10.2 introduced it alongside schemaVersion; earlier
 // versions ignore both and read `left`.
@@ -458,14 +451,16 @@ func buildExplorePanesParams(args GenerateDeeplinkParams) (url.Values, error) {
 		// Queries carry their own datasource reference in the pane format.
 		// Preserve a caller-supplied one — it may include the plugin type,
 		// which some query editors need to select the right editor mode —
-		// and only fall back to the uid when absent.
+		// and only fall back to the uid when absent, null or empty. A null one
+		// must not be propagated: Grafana would resolve it to the default
+		// datasource and render the wrong data, which is worse than an empty pane.
 		queries := make([]map[string]interface{}, 0, len(args.Queries))
 		for _, q := range args.Queries {
 			enriched := make(map[string]interface{}, len(q)+1)
 			for k, v := range q {
 				enriched[k] = v
 			}
-			if _, ok := enriched["datasource"]; !ok {
+			if v, ok := enriched["datasource"]; !ok || v == nil || v == "" {
 				enriched["datasource"] = map[string]string{"uid": *args.DatasourceUID}
 			}
 			queries = append(queries, enriched)
@@ -527,6 +522,13 @@ func exploreTimeRange(tr *TimeRange) map[string]string {
 	return rangeObj
 }
 
+// toGrafanaTimeParam converts a time value to a format Grafana understands
+// in URL query parameters. Grafana's Scenes parseUrlParam uses hardcoded
+// string length checks and only recognizes ISO 8601 at exactly 24 chars
+// (with milliseconds, e.g. "2026-04-28T12:45:00.000Z"). Shorter ISO 8601
+// strings like "2026-04-28T12:45:00Z" (20 chars) are silently ignored.
+// This function converts RFC 3339 timestamps to epoch milliseconds, which
+// is universally supported. Relative strings and epoch values pass through.
 func toGrafanaTimeParam(value string) string {
 	if _, err := strconv.ParseInt(value, 10, 64); err == nil {
 		return value
--- a/tools/navigation.go
+++ b/tools/navigation.go
@@ -46,6 +46,10 @@ const (
 	// explorePaneID keys the single pane this tool generates. Grafana uses
 	// short random ids in the UI; any stable key works for a one-pane link.
 	explorePaneID = "a"
+	// explorePanesMinMajor and explorePanesMinMinor are the first Grafana
+	// version that reads Explore state from `panes`.
+	explorePanesMinMajor = 10
+	explorePanesMinMinor = 2
 )
 
 type TimeRange struct {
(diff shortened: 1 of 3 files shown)


sd2k:
This looks good but I think it will break things for users on Grafana 9.x since it doesn't understand `panes`. https://github.com/grafana/mcp-grafana/pull/1091 adds a way to extract the Grafana version from within a tool; once it is merged, could you update this PR to fall back to the old logic if the version is <10.x?

sd2k:
Ah, also we have a policy that commits need to be signed, could you please rebase/amend commits and sign them? Thanks!

DrFaust92:
Rebased and re-signed — signatures verify now.

Also added the version gate now that #1091 has landed: `left` below 10.2, `panes` from 10.2 on. Two things worth a look:

- I gated at **10.2** rather than 10.x — that's where `panes`/`schemaVersion` actually landed, so 10.0/10.1 get the legacy format.
- **Unknown version emits `panes`.** `GrafanaVersion` fails soft, so an unreadable settings endpoint looks identical to an ancient Grafana, and gating closed would downgrade every instance that can't report a version. Happy to flip it if you'd rather bias the other way.

This also covers the Bugbot finding, which flagged the same thing.

tools/navigation.go
```diff
@@ -392,6 +386,147 @@ func buildDashboardTargetURL(baseURL string, dashboardUID *string, preview *Deep
...
+	}
+	if len(parts) < 2 {
+		return major, 0, true
+	}
+	minor, ok := leadingInt(parts[1])
+	if !ok {
+		return 0, 0, false
+	}
+	return major, minor, true
+}
```
sd2k, line 419: nit: this can probably be moved out of navigation.go, but we can do this later.

tools/navigation.go
```diff
@@ -392,6 +386,147 @@ func buildDashboardTargetURL(baseURL string, dashboardUID *string, preview *Deep
 // strings like "2026-04-28T12:45:00Z" (20 chars) are silently ignored.
 // This function converts RFC 3339 timestamps to epoch milliseconds, which
 // is universally supported. Relative strings and epoch values pass through.
```
sd2k: This doc comment has become orphaned

tools/navigation.go
```diff
@@ -392,6 +386,147 @@ func buildDashboardTargetURL(baseURL string, dashboardUID *string, preview *Deep
...
+		// Preserve a caller-supplied one — it may include the plugin type,
+		// which some query editors need to select the right editor mode —
+		// and only fall back to the uid when absent.
+		queries := make([]map[string]interface{}, 0, len(args.Queries))
+		for _, q := range args.Queries {
+			enriched := make(map[string]interface{}, len(q)+1)
+			for k, v := range q {
+				enriched[k] = v
+			}
+			if _, ok := enriched["datasource"]; !ok {
```
sd2k:
```suggestion
			if _, ok := enriched["datasource"]; !ok || v == nil || v == "" {
```

to avoid `datasource: null` being propagated through, since that would result in Grafana using the wrong datasource which could result in wrong data being shown, rather than just an empty panel (which is preferable).

sd2k (changes requested):
A couple of comments inline - most important is the `datasource: null` one.

Could you also update the README to mention `?panes` instead of just `?left` on line 220? Thanks!

tools/navigation.go
DrFaust92: Applied, with one adjustment: `v` isn't in scope at that point (it belongs to the inner `for k, v := range q` loop, which has already ended), so I bound it in the `if` instead:

```go
if v, ok := enriched["datasource"]; !ok || v == nil || v == "" {
    enriched["datasource"] = map[string]string{"uid": *args.DatasourceUID}
}
```

Same semantics as your suggestion. I also added a table test covering both the `null` and `""` cases, and wrote your reasoning into the comment above it so the intent survives — an empty pane beats the wrong data.

tools/navigation.go
DrFaust92: Good catch — the `toGrafanaTimeParam` doc comment got left behind when the function moved down the file, so it had attached itself to `supportsExplorePanes`. Moved it back above its own function.

tools/navigation.go
DrFaust92: Agreed. Leaving it here for now as you suggested, happy to pull it out in a follow-up.

DrFaust92:
Thanks @sd2k — all four points addressed, rebased onto `main`, and all three commits are signed and verified.

- **`datasource: null`** — done, with a small adjustment noted inline: `v` wasn't in scope where the suggestion landed, so it's bound in the `if` instead. Added a table test covering `null` and `""`.
- **Orphaned doc comment** — the `toGrafanaTimeParam` block had drifted onto `supportsExplorePanes` when the function moved; put back where it belongs.
- **README line 220** — now shows the `?schemaVersion=1&panes={...}` form and notes that Grafana below 10.2 gets the legacy `?left={...}` instead.
- **Moving the helper out of `navigation.go`** — left as-is per your note, happy to do it in a follow-up.

The version gate from the earlier round is unchanged: `left` below 10.2, `panes` from 10.2 on, with unknown/unparseable versions treated as 10.2+ (`mcpgrafana.GrafanaVersion` fails soft, so an unreachable settings endpoint is indistinguishable from an ancient Grafana — defaulting to the format the vast majority understand). That also resolves the Bugbot finding above, which was raised against the first commit before the gate existed.

`go build ./...`, `gofmt`, and `go test ./tools/` all pass on the rebased tree.

sd2k (approved):

sd2k:
Thanks!

PR #1091, opened 2026-08-21
feat(mcpgrafana): expose the target Grafana version to tools

sd2k:
## Summary

Adds `GrafanaVersion(ctx)`, so tools can find out which version of Grafana they're targeting. There was no way to do this before — the nearest thing was the dashboard capability check, which infers "new enough" from the API versions `dashboard.grafana.app` serves rather than reading a version.

The server already fetches `/api/frontend/settings` for `appUrl` and `namespace`, and that response carries `buildInfo`. So this is mostly parsing one more field out of a response we were already throwing away, plus a cache and an accessor.

## Changes

- Parse `buildInfo.version` in `doFetchFrontendSettings` and surface it as `frontendSettings.Version`.
- Add `GrafanaVersion(ctx) string`, fetched lazily on first use and cached per Grafana URL behind `singleflight`, mirroring the existing `publicURL` and `namespace` helpers.

Two deliberate choices worth a look in review:

- **Cache key is the URL alone**, not `URL|orgID` like `namespaceCache`. An instance's version doesn't vary by org, so partitioning by org would just mean redundant fetches. There's a test pinning this.
- **Only non-empty results are cached**, so a transient settings failure is retried rather than pinning the version to unknown for the process lifetime. Same reasoning as `publicURLCache`.

## A caveat on how this should be used

It fails soft: empty string when the endpoint is unavailable, unauthorised, or omits `buildInfo`. That means **unknown is indistinguishable from old** — a caller lacking permission to read the settings endpoint gets the same empty string as an ancient Grafana.

So this is not safe for gating anything that must fail closed. `fetchDashboard` is the cautionary case: silently routing to the legacy API on a v2-capable Grafana would read and write a lossy conversion. Capability detection via `KubernetesClient.GroupVersions` stays the right tool there. This is for advisory uses — error hints, tool descriptions, telemetry. The doc comment says so at the call site.

No caller is added in this PR; it's the mechanism only.

## Testing

Six subtests covering the happy path, settings without `buildInfo`, a 401, caching (asserting one HTTP call), retry-after-failure, and the org-independence of the cache key.

Locally: `go build`, `go vet`, `gofmt`, `make lint-jsonschema`, `make lint-openapi`, and `go test -race -count=2` on the new tests all pass.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (godoc; no README tool-table row needed — this adds no tool)

### DIFF
--- a/mcpgrafana.go
+++ b/mcpgrafana.go
@@ -983,6 +983,9 @@ type frontendSettings struct {
 	// computed server-side by Grafana's namespacer (e.g. "default", "org-2",
 	// or "stacks-123" on Grafana Cloud). Empty if not reported.
 	Namespace string
+	// Version is the Grafana version reported by buildInfo.version
+	// (e.g. "12.1.0"). Empty if not reported.
+	Version string
 }
 
 // doFetchPublicURL performs the actual HTTP request to fetch the public URL.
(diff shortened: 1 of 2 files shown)


gerboland (approved):
+1

PR #1095, opened 2026-08-21
feat(tools): add OTel counters for Loki cost guardrail decisions

### LINKED ISSUE #1093: Add OTel counters for Loki cost guardrail decisions
**Summary**

The Loki cost guardrail added in #1031 reports itself only through log lines. Add OTel counters for the decisions it makes, so embedders can alert on the guardrail rather than grep for it.

**Why**

The guardrail is designed to be rolled out in `shadow` first, then promoted to `enforce` once the would-block population looks right. That promotion is a judgement call that needs numbers, and today the only source is warn-level logs:

```
loki guardrail would block query   reasons="..." selectors="..."
loki guardrail blocked query       reasons="..." selectors="..."
loki guardrail found no parseable stream selector; passing query through (fail-open)
loki guardrail byte estimate failed (fail-open)  error="..."
```

Three problems with logs as the only signal:

1. **`reasons` is free text with the selector interpolated into it.** Grouping by which check tripped means pattern-matching on a distinguishing phrase (`no selective label matcher`, `exceeds the maximum allowed`, `above the ... budget`) rather than reading a label. That distinction matters a lot: a would-block population dominated by the selectivity check will *not* be softened by raising the byte or range budgets, because that check is unconditional. Getting the wrong answer here means promoting to `enforce` and causing exactly the breakage the shadow period was meant to prevent.
2. **The fail-open paths are invisible in aggregate.** Every `no parseable stream selector` is a query the guardrail silently admitted because the hand-rolled LogQL scanner did not recognise its shape, and every `byte estimate failed` is a budget check that did not happen. Without a rate for these, a quiet dashboard reads as "nothing would be blocked" when it may mean "the guardrail is not actually looking". A high unparseable rate is a signal to improve the scanner, which is a different fix from tuning bounds.
3. **Alerting on log volume is second-class** next to the counters this library already exports, and the log lines change text between modes (`would block` → `blocked`), so anything built on them breaks at the moment of promotion.

**Proposed instruments**

Following the existing `mcp.<subsystem>.<event>` convention (`mcp.sessions.reaped`, `mcp.client_cache.hits`, `mcp.discovery.probe_success`):

| Counter | Incremented when |
|---|---|
| `mcp.loki_guardrail.admitted` | Query passed every enabled check |
| `mcp.loki_guardrail.would_block` | Checks failed, mode is `shadow`, query ran anyway |
| `mcp.loki_guardrail.blocked` | Checks failed, mode is `enforce`, query rejected |
| `mcp.loki_guardrail.fail_open` | Guardrail could not evaluate and admitted the query |

Bounded attributes, kept low-cardinality because these become Prometheus labels downstream:

- `reason` on the block counters — one of `selector`, `range`, `bytes`. A query can trip more than one check; either increment per reason or report the first, but say which in the docs.
- `cause` on `fail_open` — one of `unparseable` (no parseable stream selector) and `estimate_failed` (index/stats unavailable). Worth distinguishing: the first is a scanner gap, the second is a Loki availability blip.
- Possibly `backend` (`loki` / `victorialogs`), since the byte check is skipped entirely on VictoriaLogs and lumping the two together makes the admitted count misleading.

Deliberately **not** proposed: the selector or the LogQL as attributes. Both are unbounded, and line filters can carry sensitive literals — which is why the existing code logs only the extracted selectors at warn and keeps the full query at debug.

**Open question: where does the MeterProvider come from?**

The other instrumented components take a provider through a constructor option (`WithSessionMeterProvider`, `WithToolManagerMeterProvider`, `WithClientCacheMeterProvider`). The guardrail has no constructor — it runs inside a tool handler and reads its settings from `GrafanaConfigFromContext`. Two options:

- A `MeterProvider` field on `GrafanaConfig`, mirroring the `Logger` field already added there for embedders, with a `MeterProviderOrDefault()` accessor next to `LoggerOrDefault()`. Per-request config, so instruments would need building once behind a `sync.Once` rather than per call.
- A package-level setter in `tools`, called at startup.

The first is more consistent with how embedders already inject a logger; the second avoids the once-vs-per-request awkwardness. No strong opinion, but it should be settled before the counters go in, since the same question will come up for any future tool-level instrumentation.

Worth noting the constraint that motivates this: an embedder whose process installs a noop global MeterProvider (as the hosted Grafana Cloud MCP server does, to avoid an otelhttp leak) drops every recording unless a provider is passed explicitly. So "just use `otel.GetMeterProvider()`" is not sufficient — the injection path is the substance of this issue, not an implementation detail.

**Related**

- #1031 — the guardrail this instruments
- #1030 — original request
- grafana/grafana-assistant-app#9618 — the log-derived monitoring this would replace
- grafana/grafana-assistant-app#9617 — the shadow-to-enforce rollout the numbers are for

sd2k:
Closes #1093

## Summary

The Loki cost guardrail (#1031) reported itself only through log lines. This adds four OTel counters for the decisions it makes, so the `shadow` → `enforce` promotion can be judged from numbers instead of grepped log text.

| Counter | Incremented when |
|---|---|
| `mcp.loki_guardrail.admitted` | Query passed every enabled check |
| `mcp.loki_guardrail.would_block` | Checks failed, mode is `shadow`, query ran anyway |
| `mcp.loki_guardrail.blocked` | Checks failed, mode is `enforce`, query rejected |
| `mcp.loki_guardrail.fail_open` | Guardrail could not evaluate and admitted the query |

The four partition the guarded population exactly once, so their sum is the number of calls the guardrail evaluated.

**Attributes** (bounded — these become Prometheus labels downstream):

- `reason` on the block counters: `selector`, `range`, `bytes`
- `cause` on `fail_open`: `unparseable` (scanner gap) or `estimate_failed` (Loki index/stats blip)
- `backend` on all four: `loki`, `victorialogs`, `unknown`

Neither the selector nor the LogQL is exported — both are unbounded, and line filters can carry sensitive literals.

## Decisions taken on the issue's open questions

**Reason attribution.** A query can trip more than one check; it is counted **once**, attributed to the check that ran first (`selector` → `range` → `bytes`). This keeps the invariant `admitted + would_block + blocked + fail_open == guarded queries`, so `sum(rate(mcp_loki_guardrail_would_block_total[5m]))` is directly the size of the population a promotion would break, and `sum by (reason)` is the breakdown. First-check attribution is also the actionable answer: the selectivity check is unconditional, so a query labelled `selector` would not be admitted by raising either budget. Documented in both README and the observability page.

**MeterProvider injection.** Took the `GrafanaConfig` option from the issue, as the more consistent one: a new `MeterProvider` field with a `MeterProviderOrDefault()` accessor next to `LoggerOrDefault()`. `main.go`'s `run()` sets it from `obs.MeterProvider()`. Instruments are built once per process behind a `sync.Once` (the config is per-request); the caveat — a process handing different requests different providers records everything against the first — is documented at the `sync.Once`, and no embedder does that today. Injection rather than `otel.GetMeterProvider()` is the point: an embedder installing a noop global provider drops every recording otherwise.

## Changes

- `mcpgrafana.go`: `GrafanaConfig.MeterProvider` + `MeterProviderOrDefault()`.
- `tools/loki_guardrail_metrics.go` (new): the four instruments, the bounded attribute value sets, and `guardrailBackendLabel`.
- `tools/loki_guardrail.go`: reason kinds are now carried alongside the human-readable message (`guardrailReason`), and the evaluation returns a `guardrailOutcome` that distinguishes a clean admission from a fail-open instead of collapsing both into an empty reason list. Recording happens at the one decision point in `guardLokiQuery`.
- `cmd/mcp-grafana/main.go`: wire `obs.MeterProvider()` into the config.
- Docs: a new "Loki cost guardrail metrics" section on the observability page covering how to read the counters during a rollout, plus pointers from the README metrics table and the CLI flags reference.

## Testing

`tools/loki_guardrail_metrics_test.go` drives `guardLokiQuery` against a manual-reader `MeterProvider` and asserts the exact counter and attribute set for: clean admission; shadow vs. enforce vs. unknown-mode (fail-closed, counts as `blocked`); each `reason` value; `selector` winning over a co-occurring `range`; both fail-open causes; a partial-total-over-budget that must count as a block rather than a fail-open; mode `off` recording nothing; the injected provider actually receiving the recordings; and a pin on the attribute keys so no unbounded attribute creeps in.

The existing guardrail unit tests were updated for the new return type and strengthened to assert reason kinds and fail-open causes rather than only prose.

`make lint` and `go test -tags unit ./...` pass.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated
- [x] Linting passes

### DIFF
--- a/README.md
+++ b/README.md
@@ -1126,6 +1126,19 @@ When using the SSE or streamable HTTP transports, enable Prometheus metrics with
 
 **Note:** Metrics are only available when using SSE or streamable HTTP transports. They are not available with the stdio transport.
 
+When the [Loki cost guardrail](#cli-flags-reference) (`--loki-guardrail-mode`) is enabled, four more counters record its decisions:
+
+| Metric | Type | Description |
+|--------|------|-------------|
+| `mcp_loki_guardrail_admitted_total` | Counter | Queries that passed every enabled check (labels: `backend`) |
+| `mcp_loki_guardrail_would_block_total` | Counter | Queries that failed a check in `shadow` mode and ran anyway (labels: `backend`, `reason`) |
+| `mcp_loki_guardrail_blocked_total` | Counter | Queries rejected in `enforce` mode (labels: `backend`, `reason`) |
+| `mcp_loki_guardrail_fail_open_total` | Counter | Queries the guardrail could not evaluate and admitted (labels: `backend`, `cause`) |
+
+`reason` is one of `selector`, `range`, `bytes`; `cause` is one of `unparseable`, `estimate_failed`; `backend` is one of `loki`, `victorialogs`, `unknown`. A query that trips several checks is counted once, labelled with the check that ran first (`selector`, then `range`, then `bytes`), so the four counters partition the guarded population. See [Observability](docs/sources/developer/observability-metrics-and-tracing.md#loki-cost-guardrail-metrics) for how to read them during a `shadow` → `enforce` rollout.
+
+Library embedders should set `GrafanaConfig.MeterProvider` (the metrics counterpart of `GrafanaConfig.Logger`): the guardrail runs inside a tool handler, so it has no constructor option, and a process that installs a noop global `MeterProvider` would otherwise drop every recording.
+
 #### Slow-request logging
 
 The `--slow-request-threshold` flag emits a structured log event whenever an MCP request (tool invocation, list, resource read, etc.) exceeds the given duration. It is useful for diagnosing slow queries and tool calls without drowning in the full debug log.
--- a/cmd/mcp-grafana/main.go
+++ b/cmd/mcp-grafana/main.go
@@ -670,6 +670,13 @@ func run(transport, addr, basePath, endpointPath string, logLevel slog.Level, dt
 		slog.Info("OTLP trace export configured", "endpoint", observability.OTLPTracesEndpoint())
 	}
 
+	// Instrumentation that lives inside tool handlers (the Loki cost
+	// guardrail) has no constructor to take a meter provider option, so it
+	// reads one off the GrafanaConfig instead. Set explicitly rather than
+	// relying on otel.GetMeterProvider() so the counters land on this
+	// process's provider.
+	gc.MeterProvider = o.MeterProvider()
+
 	// Create a client cache for HTTP-based transports to avoid per-request
 	// transport allocation (see https://github.com/grafana/mcp-grafana/issues/682).
 	var clientCache *mcpgrafana.ClientCache
--- a/docs/sources/configure/command-line-flags.md
+++ b/docs/sources/configure/command-line-flags.md
@@ -121,6 +121,8 @@ When caller authentication is enabled, the `Authorization` header is reserved fo
 - `--loki-guardrail-max-bytes`: Maximum bytes a single `query_loki_logs` call may scan, estimated via Loki's index/stats API. Defaults to 100 GiB; `0` disables the byte-budget check. Falls back to `GRAFANA_LOKI_GUARDRAIL_MAX_BYTES`.
 - `--loki-guardrail-max-range`: Maximum effective time range for a single `query_loki_logs` call, including range-vector durations. Defaults to `24h`; `0` disables the range check. Falls back to `GRAFANA_LOKI_GUARDRAIL_MAX_RANGE`.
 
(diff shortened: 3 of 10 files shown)


sd2k:
cursor review

ioanarm (approved):
nice! 🚀

Issue #1092, opened 2026-08-21
Consolidate the tool surface into action-based read/write tool pairs


sd2k:
> Placeholder / umbrella issue. The shape of the proposal matters more than the specifics below; individual categories can be split into their own issues as work starts.

## The problem

The tool surface has grown past what clients handle well. Measured against the current `main` (2026-08-21), the hosted Grafana Cloud MCP server advertises:

| | Tools | JSON schema | ≈ tokens |
|---|---|---|---|
| Full surface (write enabled) | 110 | 178 KB | ~44,600 |
| Read-only grant (scope-filtered) | 82 | 102 KB | ~25,500 |

(Token figures are a rough 4-chars-per-token estimate over the serialized `mcp.Tool` values, so treat them as an order of magnitude, not a measurement.)

That is ~45k tokens of tool definitions before the model has read a single word of the user's question. #569 raised this at ~16k tokens in February; the surface has since roughly tripled. Beyond the context cost, many clients degrade in tool-selection accuracy well before 100 tools, and some cap the list outright.

## Proposal

Consolidate related tools into **action-based tools**, and split each into a **read tool and a write tool as two distinct names**:

```
dashboards_read   { action: "get" | "search" | "summary" | "panel_queries" | "property", ... }
dashboards_write  { action: "create" | "update" | "delete", ... }
```

rather than today's roughly one-tool-per-endpoint layout. `alerting_manage_rules` and the `agento11y_manage_*` family already do the action-parameter half of this, so the pattern is established — the new part is the read/write split.

## Why split the names rather than keep one gated tool

The current consolidated tools register **one tool name from two variants** — `ManageRulesRead` and `ManageRulesReadWrite` both register as `alerting_manage_rules`, with `enableWriteTools` selecting which. That shape has a real downstream cost: it is invisible to any consumer that classifies tools by name.

The hosted Cloud MCP server has to gate tools on OAuth scope (`grafana:read` / `grafana:query` / `grafana:write`), and it derives the write set by registering the library twice — once with write off, once with write on — and diffing the tool names. Same-name variants produce an empty diff, so every one of them has to be hand-listed as write-capable, with a cross-check test against the `readOnlyHint` annotation to catch new ones. That list currently carries `alerting_manage_rules`, `alerting_manage_silences`, `generate_deeplink`, `grafana_api_request` and five `agento11y_manage_*` tools, and it grows with every new consolidated tool upstream adds.

[...]

PR #1098, opened 2026-08-21
feat(tools): tool-consolidation foundation + admin_read (#1092)

sd2k:
## Summary

Foundation for the tool-surface consolidation proposed in #1092 (design: https://github.com/grafana/mcp-grafana/issues/1092#issuecomment-5370582352). This PR is now a clean, self-contained unit: the shared validation/delegation machinery, the tool-name-reference drift test, and `admin` as the read-only reference domain. The read/write reference domain (`sift`) is a separate stacked PR, #1104, branched off this one — see below for why.

This is a **draft**, not ready for review/merge — it establishes conventions for the sibling PRs covering the other consolidated domains.

### 1. Shared foundation

- **`tools/operation_validation.go`** — `OperationValidator`, a reusable helper for the "discriminate on an `operation` param" pattern that `alerting_manage_rules` and the `agento11y_manage_*` tools each currently hand-roll via `switch p.Operation { ... }`. Covers: unknown-operation reporting, per-operation required fields (`Require`), at-most-one field groups (`MutuallyExclusive`/`MutuallyExclusiveFor`), and exactly-one field groups (`ExactlyOne`/`ExactlyOneFor`).
- **Sentinel-based read/write delegation**, generalized from `agento11y_manage_evaluators`'s hand-rolled version (`tools/agento11y_eval_types.go:106`, `_handlers.go:201-221`): a shared `ErrUnknownOperation` sentinel, `OperationValidator.Check()` (bare sentinel instead of a formatted message), `UnknownOperationError` (final formatting), and `DelegateValidation`/`DelegateDispatch` (the "try the primary, fall through only on the sentinel" idiom). `tools/operation_validation_test.go` documents **two** reference shapes — read the doc comments, they matter:
  - `TestDelegation_SingleNameDualSchema` — one tool name, schema toggles between read-only and read+write (`alerting_manage_rules`, `agento11y_manage_*`). Delegation here removes real duplicated dispatch logic. **Only relevant if you're refactoring an existing single-name dual-schema tool — not for a new split-name domain.**
  - `TestDelegation_SplitNameDisjointOperations` — the **new** split-name convention, which every domain actually being consolidated uses (`admin_read`; and `sift_read`/`sift_write` in #1104). The two tools' operations are disjoint by design — a write tool's jsonschema enum never offers a read operation — so delegation's job is a better error message on a malformed/out-of-schema operation value. **It must never let a write tool accept a sibling read tool's operation as valid for itself** — that would silently grant it capability it doesn't advertise, exactly what the split-name convention exists to prevent. **This is the shape to copy.**
- **`tools/tool_name_references_test.go`** — a drift test guarding the ~60 tool-name mentions embedded in tool descriptions and generated hint text. Registers the full write-enabled tool set plus every `GenerateEmptyResultHints` branch, and flags tool-name-shaped tokens following a verb (use/via/call/run/invoke/prefer/calling/created-with, in various inflections) or wrapped in backticks/"the X tool". Caught one pre-existing stale reference on current `main` (`loki_label_analyzer.go`, fixed). Its verb coverage was broadened after a real miss surfaced while briefly consolidating `annotations` (see below) — **this test narrows a manual grep before/after a rename, it does not replace one**; every domain PR's checklist should include both.
- **Registration pattern** for read/write-split domains:
  ```go
  func AddXTools(mcp *server.MCPServer, enableWriteTools bool) {
      XRead.Register(mcp)
      if enableWriteTools { XWrite.Register(mcp) }
  }
  ```
  For read-only domains (no write half), no `enableWriteTools` param at all — see `alerting_manage_routing` for the pre-existing precedent, and don't invent a write half that doesn't exist (`admin_read` has none; don't add `admin_write`).

### 2. Reference domain: `admin` (9 tools -> 1, read-only)

Consolidates all nine (read-only) admin tools into `admin_read` with an `operation` param. No `admin_write` — none of the underlying APIs support mutation. Split into `tools/admin_types.go` / `admin_handlers.go` / `admin.go`.

**BREAKING CHANGE**: the 9 admin tool names no longer exist.

Measured for `--enabled-tools admin`: 9 tools / 4,627 bytes -> 1 tool / 3,137 bytes (**-32.2%**).

### 3. `annotations` was consolidated here, then reverted — read this before picking a domain

`annotations_read`/`annotations_write` (4 tools -> 2) was built and measured **negative**: +15.1% schema bytes write-enabled, +18.4% read-only — fewer tools but a bigger, more expensive schema, because the original 4 tools were already tightly scoped with almost no field overlap to eliminate. Per the repo owner: **consolidation only pays off when several tools with overlapping parameters actually collapse together; a domain that costs tokens for no benefit doesn't get consolidated just because the mechanical work is straightforward.** The annotations consolidation (commit `42db07b`) was reverted (commit `c3d29c6`) and is no longer part of this PR's diff — it's mentioned here only so the history and reasoning aren't lost. The `delete_annotation` capability it added along the way (wiring up `DeleteAnnotationByIDWithParams`, already vendored in `grafana-openapi-client-go` but never exposed as a tool) went with the revert — it was a genuine new capability, not repackaging, so it's filed separately as #1103 to be judged on its own merits.

**Implication for the remaining domains: measure early, and treat a negative result as a real, reportable answer — not a failure to fix by pushing through.** If your domain's tool count drops but its schema bytes don't, say so and stop; don't ship it. See #1104 (`sift`) for a domain that measured positive, and the correction posted on #1092 (comment 5380203127) for the numbers behind this rule.

## Stacked PR

#1104 (`sift_read`/`sift_write`, branched off this PR's branch) is the read/write reference domain, built after `annotations` was reverted here. **Do not merge #1104 before this PR.**

## Base branch note

**Sequencing decision (from the repo owner): the #1092 consolidation proceeds on `main` first; the mark3labs -> official-SDK migration (`gosdk-migration-server`, #1076 -> #1078 -> #1082) rebases over this work later**, not the other way around — that stack has failing checks and merge conflicts on all three PRs and none of them are reviewed yet. This PR stays based on `main`; #1104 stacks on this one.

For whoever does that later rebase, here's the precise per-file SDK-coupling list from this PR:
- `tools/admin.go` — the `mcp.WithXAnnotation(...)` calls (mark3labs) become package-level `mcpgrafana.WithXAnnotation(...)` on the migrated branch; `AddAdminTools`'s `*server.MCPServer` parameter becomes `*mcp.Server` (official SDK). Same for every other domain's `AddXTools` — all 33 on the migrated branch take `*mcp.Server`.
- `tools/tool_name_references_test.go` — its `srv.ListTools() map[string]*ServerTool` call has no direct equivalent on the official SDK's `*mcp.Server`. On the migrated branch this needs `mcp.NewInMemoryTransports()` + a client session's `ListTools`/`Tools(ctx, nil)` iterator instead, mirroring the pattern already in that branch's `tools/agento11y_eval_test.go` (`listAgento11yEvalTools` helper).
- Worth flagging now rather than at rebase time: on the migrated branch, `AddOnCallTools` has *dropped* its `enableWriteTools` parameter entirely (unrelated to this PR, but oncall is a later #1092 consolidation domain — that's a latent shape conflict between the two efforts for whoever does that domain).
- (#1104 adds `tools/sift.go` to this same list, with the same `mcp.WithXAnnotation`/`*server.MCPServer` coupling.)

Everything else — `tools/operation_validation.go` (+ its test), `tools/admin_types.go`/`admin_handlers.go`, and all docs/Python changes — is SDK-agnostic and should carry over untouched.

## Follow-ups filed

- #1099 — shared time-range and pagination param structs, deliberately deferred out of this PR (existing per-domain description text carries real, non-generic API semantics a shared struct's static jsonschema tag can't preserve without becoming lossy).
- #1103 — expose `delete_annotation` (wiring up the already-vendored `DeleteAnnotationByIDWithParams`) as its own change, split out of the reverted annotations consolidation.

## Convention gaps found while building `agento11y` (8→3, #1108)

Neither `admin` (9→1) nor `sift`/`snapshot`/`incident` (all clean 1:1 read/write pairs) exercised these, because none of them merge more than two source tools into one destination, or produce more than two output tools. `agento11y` does both (8 tools → `agento11y_read` / `agento11y_evals_read` / `agento11y_evals_write`), and surfaced two real gaps in the conventions this PR set:

1. **Shared Go field structs are unsound across an n-way merge unless type *and* semantics match, not just the field name.** The precedent this PR and #1104/#1106/#1107 established (see `alerting_manage_rules`'s `FolderUID`, or `annotations`' reverted create/update split) was "don't share a field across operations where its *role* differs, even if the name is identical" — established for 2-way merges. Merging 3+ source tools' fields into one struct raises the odds that a field name lines up across *some* of the sources by coincidence rather than by shared meaning, which the 2-way guidance doesn't explicitly warn against. Rule of thumb for the next n-way merge: before embedding or sharing a field, check it against every source operation being merged, not just the two most obviously related ones.
2. **No documented file-layout convention for a domain that produces more than two output tools.** `<domain>_types.go` / `<domain>_handlers.go` / `<domain>.go` (this PR, `#1104`, `#1106`, `#1107`) assumes one types file and one handlers file cover exactly one read tool and one write tool. `agento11y`'s three output tools needed a convention for splitting (or not splitting) those files further — not resolved here; whoever picks up a domain with 3+ output tools next should establish it and report back rather than reinventing it per-domain.

## Notes for the sibling PRs (other domains)

- **Measure before you commit to a domain.** If tool count drops but bytes don't (or barely do), that's a legitimate "don't consolidate this" answer — report it and stop, per the `annotations` finding above.
- Copy `tools/operation_validation.go`'s usage pattern from `tools/admin_types.go`'s `validate()` for a read-only domain, or #1104's `tools/sift_types.go`'s `SiftWriteParams.validate()` for a read+write domain with disjoint operations — the shape every actually-consolidated wave-1 domain has. Only reach for `TestDelegation_SingleNameDualSchema`'s shape if you're refactoring an existing single-name dual-schema tool like `alerting_manage_rules` itself.
- Copy the file layout: `<domain>_types.go` (params + `validate()`), `<domain>_handlers.go` (dispatch + business logic), `<domain>.go` (description + `MustTool` vars + `AddXTools`).
- Run `TestToolNameReferencesAreValid` before and after your rename — **and also run a plain `grep` for the old tool name across non-test `.go` files**, since the matcher has documented, real blind spots.
- `make lint` runs `lint-jsonschema`, which will catch unescaped commas in `enum=...,enum=...` discriminator tags.
- Check for hand-maintained duplicates beyond `README.md`: `docs/sources/reference/mcp-tools-table.md` (tools table) and `docs/sources/configure/command-line-flags.md` (the `--disable-write` tool list) both mirror README content and go stale independently. `tests/disable_write_test.py` has two `write_tools`/`read_tools` lists, each duplicated across two test functions (4 lists total) — update all of them for any renamed write/read tool.
- Check whether your domain already has unit tests before assuming a cloud/integration-tagged suite is enough coverage — `sift` (#1104) had none at all before that PR added them.
- Branch your domain's PR off this one (`feat/tool-consolidation-foundation`), same as #1104 does, and mark it stacked/draft until this PR lands.

## Test plan

- [x] `go vet ./...`
- [x] `make lint` (jsonschema, openapi, gofmt, golangci-lint)
- [x] `make test-unit` — all packages pass, including `TestOperationValidator_*`, `TestDelegate*`, and `TestAdminRead*`
- [x] `go vet -tags cloud ./tools/...` compiles (admin's cloud-tagged integration suite)
- [x] Manually dumped `tools/list` for `--enabled-tools admin`, with and without `--disable-write`, to measure the delta above
- [ ] Not run: `tests/admin_test.py` / e2e suite (requires a live Grafana + LLM credentials) — updated `expected_tools` to the new tool names but unverified against a live instance; `tests/disable_write_test.py` similarly unverified live


sd2k:
Two convention gaps surfaced while consolidating `agento11y` (8 tools → 3, #1108) that this PR's conventions didn't cover, because every domain so far merges ≤2 tools into a result. Writing them up here in a form that can be lifted straight into this PR's body, per the repo owner's request.

---

### Convention addition: field-struct embedding is unsound under an n-way merge

`sift` and `admin` each merged tools that had no pre-existing shared field-group types, so a merged domain's params struct could just declare fields fresh. `agento11y`'s eval sub-domains already had their own embeddable field-group structs from an earlier refactor (`agento11yEvaluatorFields`, `agento11yEvalRuleFields`, `agento11yEvalCollectionFields`, `agento11yExperimentFields`, `agento11yTestSuiteFields`), and naively embedding all of them into one flat struct for a 5-into-2 merge is unsound in two distinct ways:

- **Ambiguous promoted fields.** `Limit`/`Cursor` exist in every one of the five field-groups. Embedding multiple structs with the same field name at the same depth compiles, but the field becomes unaddressable unqualified, and — more importantly — `encoding/json` silently drops depth-ambiguous fields from (de)serialization rather than erroring. This does not show up as a compile error; it shows up as a field quietly not round-tripping.
- **Type conflicts.** `Tags` is `map[string]string` in one sub-domain (saved-conversation bookmark tags) and `*[]string` in two others (experiment/test-suite tag lists) — a straight type conflict that *does* fail to compile, but only once you've already built the embed.

**The convention:** for an n-way merge (n > 2) where the source tools already have field-group structs, don't embed them into the new flat struct. Hand-flatten every field once, and share one Go field across sub-domains only where both the **type** and the **semantics** (not just the name) match — rename otherwise. In `agento11y_evals_types.go`, `Name`/`Description`/`Definition`/`Metadata` ended up shared across sub-domains this way, while eval-collections' `source` (telemetry/manual) and experiments' `source` (collection/external) — same field name pre-merge, unrelated enums — became `source` and `experiment_source` respectively.

### Convention addition: no file-layout precedent for a 5-into-2 (or 3-into-1) merge

The `<domain>_types.go` / `<domain>_handlers.go` / `<domain>.go` split holds for a 2-into-1 merge, but doesn't say what to do when five *already-separate* implementation files (each with their own wire types, Client methods, and per-tool params/dispatch) need to become one merged tool's params/dispatch/registration layer, while the sub-domain-specific wire types and Client methods are still worth keeping in their own files for readability.

**The convention used in #1108:** keep each sub-domain's existing `<subdomain>_types.go`/`_handlers.go` files, trimmed down to wire types, shared field-group structs (kept for reuse by the merged struct's constructor methods, not embedded — see above), and `Client` methods/business logic. Add new `<domain>_read_types.go`/`_handlers.go`/`.go` (or `<domain>_evals_types.go`/`_handlers.go`/`.go`) files on top, containing only the new flat merged params struct, its `validate()`, and the dispatch function that fans out to each sub-domain's existing (now-internal) dispatch helper.

---

Also worth a one-line addition to the file-layout section regardless of merge size: a brace-depth-counting flattener, not a regex, is needed if you ever need to programmatically flatten `X: X{...}` field-group literals in test files during a port — a non-greedy-regex version breaks silently on any slice/map literal inside the group (e.g. `Tag: []string{"critical", "", "nightly"}` — the regex grabs the slice's own closing brace instead of the struct's). Only relevant if a future domain's tests get ported mechanically rather than by hand, but cheap to note.

Reference: #1108 (agento11y consolidation), where all three of the above came up.

PR #1100, opened 2026-08-21
fix(auth): reload token file for stdio requests

### LINKED ISSUE #987: stdio transport does not re-read GRAFANA_SERVICE_ACCOUNT_TOKEN_FILE after startup, unlike HTTP/SSE
## Summary

`GRAFANA_SERVICE_ACCOUNT_TOKEN_FILE` is documented (README, CHANGELOG for #935) as being read fresh on every request so that token rotation (e.g. a Kubernetes Secret volume mount updated in place by kubelet) is transparent with no server restart needed. That is true for the HTTP and SSE transports, but not for stdio: the stdio transport builds one static context (and therefore one Grafana API client, with whatever token was valid at that moment) at process start, via `mcp-go`'s `StdioServer.Listen()`, and reuses it for every tool call for the rest of the process's life.

## Root cause

- `cmd/mcp-grafana/main.go`, stdio branch:
  ```go
  case "stdio":
      srv := server.NewStdioServer(s)
      cf := mcpgrafana.ComposedStdioContextFunc(gc)
      srv.SetContextFunc(cf)
  ```
- `ComposedStdioContextFunc` chains `ExtractGrafanaInfoFromEnv` + `ExtractGrafanaClientFromEnv`, not the cached, per-request `extractGrafanaClientCached` that `ComposedHTTPContextFunc`/`ComposedSSEContextFunc` use.
- `mark3labs/mcp-go`'s `server/stdio.go`, `(s *StdioServer) Listen()`:
  ```go
  // Set a static client context since stdio only has one client
  ...
  if s.contextFunc != nil {
      ctx = s.contextFunc(ctx)
  }
  ...
  for i := 0; i < s.workerPoolSize; i++ {
      s.workerWg.Add(1)
      go s.toolCallWorker(ctx)
  }
  ```
  `contextFunc` (and therefore the Grafana client resolved from `GRAFANA_SERVICE_ACCOUNT_TOKEN_FILE`) is built exactly once, before the tool-call worker pool starts, and that same context is reused indefinitely.

## Impact

Any long-lived stdio deployment (an MCP client that spawns mcp-grafana as a subprocess and keeps it running for hours/days, e.g. an agent runtime with a persistent session) will silently start 401ing on every authenticated Grafana call after the underlying token rotates, with no recovery short of restarting the subprocess. This is easy to miss because:

- the file on disk is correct the whole time (a `curl` using the same file succeeds)
- nothing in mcp-grafana logs or errors distinguishes "credential genuinely invalid" from "process is holding a credential that's since been rotated out"
- the README's rotation-transparency claim doesn't scope itself to HTTP/SSE, so it's natural to assume stdio behaves the same way

## Reproduction

1. Run `mcp-grafana -t streamable-http -address 127.0.0.1:PORT` against a token file.
2. Initialize an MCP session, call `list_datasources`, confirm success.
3. Overwrite the token file on disk with an invalid value, without restarting the process.
4. Call `list_datasources` again, same session: fails with `401 Invalid API key`, proving per-request re-read.
5. Restore the valid token in the file, still without restarting: call succeeds again immediately.

Repeating steps 1-5 with `-t stdio` instead: after step 3, the tool call keeps succeeding (using the stale in-memory client), because the file is never re-consulted after startup. (We reproduced steps 1-5 in HTTP mode directly; the stdio half follows directly from the `Listen()` source above and matches what we independently observed happening to a real ~12h-rotated token in a long-lived production stdio deployment.)

## Suggested fix (either would resolve this)

- **Preferred**: have the stdio transport rebuild/re-resolve its Grafana client per tool call too (mirroring `extractGrafanaClientCached`'s per-request cache-by-token-value approach), rather than baking one client into a static context. `mcp-go`'s "stdio only has one client" assumption is about the *transport session*, not about the credential behind it staying constant for that session's lifetime; a rotated token is a legitimate mid-session credential change even over stdio.
- **Minimum**: scope the README/docs claim explicitly to HTTP/SSE, so operators running mcp-grafana over stdio in a long-lived process don't assume rotation is handled for them.

## Environment

- mcp-grafana version: 0.17.0 (also read the `main` branch source as of 2026-07-08; the relevant code paths are unchanged)
- Transport: stdio, spawned as a long-lived subprocess by an agent runtime, `GRAFANA_SERVICE_ACCOUNT_TOKEN_FILE` pointed at a Kubernetes Secret volume mount rotated by External Secrets Operator every 12h

Tethys0:
## What does this PR do?

Fixes #987.

Long-lived stdio servers currently build the Grafana API client once at startup. When `GRAFANA_SERVICE_ACCOUNT_TOKEN_FILE` is rotated, subsequent tool calls keep the old token because the OpenAPI client's static authorization header is reused.

This change adds a transport layer for file-based service-account authentication that reads the token and replaces the `Authorization` header for each outgoing Grafana request. Inline `GRAFANA_SERVICE_ACCOUNT_TOKEN` and the deprecated `GRAFANA_API_KEY` retain precedence, while token-file deployments now pick up rotated credentials without restarting the stdio process.

## Verification

- TDD RED: the focused rotation test initially failed because `tokenFileRoundTripper` did not exist.
- TDD GREEN: focused rotation tests pass.
- `make test-unit` passes.
- `go vet ./...` passes.
- JSON-schema, OpenAPI, and gofmt portions of `make lint` pass; the Go lint portion could not run locally because `golangci-lint` is not installed.

The tests cover both the transport directly and the real generated Grafana client path, including a token-file rewrite between requests.

### DIFF
--- a/client_cache.go
+++ b/client_cache.go
@@ -361,8 +361,8 @@ func extractGrafanaClientCached(cache *ClientCache) httpContextFunc {
 		if config.OrgID == 0 {
 			logger.Warn("No org ID found in request headers or environment variables, using default org. Set GRAFANA_ORG_ID or pass X-Grafana-Org-Id header to target a specific org.")
 		}
-
 		u, apiKey, basicAuth, _ := extractKeyGrafanaInfoFromReq(req, logger)
+		ctx = markRequestScopedAuth(ctx, req)
 		key := cacheKeyFromRequest(u, apiKey, basicAuth, config.OrgID, req)
 
 		grafanaClient := cache.GetOrCreateGrafanaClient(key, func() *GrafanaClient {
--- a/mcpgrafana.go
+++ b/mcpgrafana.go
@@ -1325,9 +1325,6 @@ func NewGrafanaClient(ctx context.Context, grafanaURL, apiKey string, auth *url.
 							IdleConnTimeout:       90 * time.Second,
 						}
 					}
-					if tokenFile := serviceAccountTokenFileFromEnv(); tokenFile != "" {
-						base = &tokenFileRoundTripper{path: tokenFile, underlying: base}
-					}
 					// Use BuildTransport but skip APIKey/BasicAuth auth
 					// (handled by the OpenAPI client). OBO tokens still need
 					// transport-level injection since the OpenAPI client
@@ -1348,6 +1345,13 @@ func NewGrafanaClient(ctx context.Context, grafanaURL, apiKey string, auth *url.
 					wrapped, err := BuildTransport(&oboConfig, base)
 					if err != nil {
 						panic(fmt.Errorf("failed to build transport: %w", err))
+					}
+					if tokenFile := serviceAccountTokenFileFromEnv(); tokenFile != "" && tokenFileMatchesAPIKey(tokenFile, apiKey) {
+						wrapped = &tokenFileRoundTripper{
+							path:          tokenFile,
+							fallbackToken: apiKey,
+							underlying:    wrapped,
+						}
 					}
 					transportField.Set(reflect.ValueOf(wrapped))
 					logger.Debug("HTTP tracing, user agent tracking, and timeout enabled for Grafana client", "timeout", timeout)
--- a/mcpgrafana.go
+++ b/mcpgrafana.go
@@ -206,6 +206,28 @@ func apiKeyFromHeaders(req *http.Request) string {
 	return req.Header.Get(grafanaAPIKeyHeader)
 }
 
+type requestScopedAuthContextKey struct{}
+
+func requestHasGrafanaCredentials(req *http.Request) bool {
+	if apiKeyFromHeaders(req) != "" || req.Header.Get("Authorization") != "" {
+		return true
+	}
+	username, password, _ := req.BasicAuth()
+	return username != "" || password != ""
+}
+
+func markRequestScopedAuth(ctx context.Context, req *http.Request) context.Context {
+	if !requestHasGrafanaCredentials(req) {
+		return ctx
+	}
+	return context.WithValue(ctx, requestScopedAuthContextKey{}, true)
+}
+
+func requestScopedAuthIsProtected(ctx context.Context) bool {
+	protected, _ := ctx.Value(requestScopedAuthContextKey{}).(bool)
+	return protected
+}
+
 // grafanaConfigKey is the context key for Grafana configuration.
 type grafanaConfigKey struct{}
 
@@ -572,6 +594,9 @@ func (t *ExtraHeadersRoundTripper) RoundTrip(req *http.Request) (*http.Response,
 	if len(headers) > 0 {
 		propagated := propagatedHeaderFields()
 		for k, v := range headers {
+			if tokenFileAuthIsProtected(clonedReq.Context()) && isTokenFileCredentialHeader(textproto.CanonicalMIMEHeaderKey(k)) {
+				continue
+			}
 			// Never overwrite a trace-context header the OTel propagator has
 			// already injected for this request. A traceparent forwarded from
 			// the incoming request (GRAFANA_FORWARD_HEADERS) names the caller's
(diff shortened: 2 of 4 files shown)


sd2k:
Thanks for this — the goal is clearly right (long-lived stdio clients must pick up rotated token files), the precedence rules are carefully thought through, and the test coverage in token_file_transport_test.go is genuinely thorough. My main feedback is
  architectural rather than line-level: I think the design is solving a problem it creates for itself, and simplifying the root cause makes several of the smaller issues disappear.

  The hard part of this PR is disambiguating, at RoundTrip time, whether an Authorization header is the stale startup token we should replace or a genuine request-scoped credential we must not touch. We answer that by string-comparing the header against a
  remembered fallbackToken:

  if authorization == "Bearer "+strings.TrimSpace(t.fallbackToken) { ... }

  That single ambiguity is what forces most of the surrounding complexity:

  - Two parallel context markers for nearly the same concept — tokenFileAuthContextKey (token_file_transport.go) and requestScopedAuthContextKey (mcpgrafana.go).
  - A four-branch RoundTrip enumerating file-readable × header-present × matches-fallback.
  - A three-term rotation predicate in NewGrafanaClient where tokenFileMatchesAPIKey re-reads the file just to decide whether the feature is enabled.
  - The TLS unwrap/rewrap dance in BuildTransport, which exists only because the token-file layer sits mid-stack and has to be peeled off and reapplied around the TLS transport.

  The reason the header is ambiguous at all is that go-openapi injects cfg.APIKey as Authorization: Bearer <key> before our transport runs — so by round-trip time the file token and a real request token are indistinguishable, and we reverse-engineer the
  difference instead of removing it.

  Suggested direction

  Don't inject the startup token into cfg.APIKey when a rotating file is configured. Leave cfg.APIKey empty and let the file transport be the sole authority for the file credential. RoundTrip then collapses to:

  func (t *tokenFileRoundTripper) RoundTrip(req *http.Request) (*http.Response, error) {
      if hasAuth(req) { // request brought its own credential
          return t.underlying.RoundTrip(req)
      }
      // file is the only source of auth — read it fresh each time
      ...set Authorization from file, fall back if unreadable...
  }

  This removes the fallbackToken comparison, tokenFileMatchesAPIKey, the requestScopedAuthContextKey marker (a present header speaks for itself), and the tokenFileAuthIsProtected guards in AuthRoundTripper/ExtraHeadersRoundTripper — those layers were only
  ever competing with the injected cfg.APIKey we've now stopped injecting. It should delete more than it adds.

  Worth considering alongside this: stdio and HTTP have opposite needs being served by one mechanism. stdio needs rotation and never has request-scoped credentials; HTTP already builds a client per request from headers, so it re-reads the file at construction
  and barely needs the rotating transport. Separating the two removes most of the reconciliation logic.

  Smaller findings (several resolved by the above)

  1. extractGrafanaClientCached doesn't mark request-scoped auth — client_cache.go:370. Unlike its sibling ExtractGrafanaClientFromHeaders (mcpgrafana.go:1461), the cached extractor doesn't call markRequestScopedAuth. It's currently masked because
  ExtractGrafanaInfoFromHeaders runs first in both composed context funcs, but if that ordering ever changes, a request's own X-Grafana-Service-Account-Token gets silently replaced by the server's file token. No test covers the cached path. (Goes away
  entirely under the refactor, since a present header would win by construction.)
  2. Behavior regression: non-*http.Transport base + TLSConfig now errors — mcpgrafana.go:822. Previously this fell back to cloning http.DefaultTransport; now it returns an error that NewGrafanaClient turns into a panic. A config using a custom BaseTransport
  (the roundTripFunc test pattern / hosted deployment) plus TLS that worked before will crash. If this is intentional it deserves a changelog note; otherwise the fallback should be preserved.
  3. tokenFileMatchesAPIKey returns true for an empty key without checking readability — token_file_transport.go:48. With an unreadable file and no token/API key set, rotation is "enabled" with an empty fallback, so every RoundTrip fails with a read error
  instead of the pre-PR unauthenticated behavior. The truthy return for an unreadable file is surprising. (Also disappears if the file transport isn't gated on matching the key.)
  4. os.ReadFile on every RoundTrip — token_file_transport.go:67. Justified for stdio rotation, but in cached HTTP mode the cache key already changes on rotation, so the per-request read is largely redundant under load.
  5. Double canonicalization — mcpgrafana.go:597 / token_file_transport.go:34. isTokenFileCredentialHeader is called with an already-canonical key and re-canonicalizes it (and the three constants) on every call. Minor.

gerboland:
I disagree with @sd2k , if you change the token file, then you should relaunch the mcp server to pick up the new state.

If you have a usecase I've not considered, please re-open and share it.

PR #1104, opened 2026-08-22
feat(sift): consolidate 5 sift tools into read/write pair (#1092)

sd2k:
## Summary

**Stacked on #1098** (base branch: `feat/tool-consolidation-foundation`, not `main`) — do not merge before #1098 lands. This is the read/write reference domain for the #1092 tool-surface consolidation, replacing `annotations` (which measured negative and was reverted in #1098 — see that PR for the full story and the reasoning this domain avoids repeating).

`sift_read` (`list_investigations`, `get_investigation`, `get_analysis`) + `sift_write` (`find_error_pattern_logs`, `find_slow_requests`), replacing `list_sift_investigations`, `get_sift_investigation`, `get_sift_analysis`, `find_error_pattern_logs`, `find_slow_requests`. Split into `tools/sift_types.go` / `sift_handlers.go` / `sift.go`, following the file layout #1098 establishes.

## Why sift, and why it's expected to measure positive

`find_error_pattern_logs` and `find_slow_requests` had **identical** param shapes before consolidation — both exactly `{Name, Labels, Start, End}`. `SiftWriteParams` needs no per-operation field split at all: `Name`/`Labels` are unconditionally required for the whole tool, declared `jsonschema:"required"` directly rather than through `OperationValidator.Require` (which exists for fields required for *some* but not all of a tool's operations). This is the real parameter overlap `annotations` didn't have, and it's why this domain was picked to replace it as the exemplar.

## Disjoint operations, validation-layer delegation only — same shape as #1098's `admin_read`

`sift_write`'s operation set (`find_error_pattern_logs`, `find_slow_requests`) is disjoint from `sift_read`'s (`list_investigations`, `get_investigation`, `get_analysis`) by design — `sift_write`'s jsonschema enum never offers a read operation. `SiftWriteParams.validate()` uses `DelegateValidation`/`ErrUnknownOperation` (from #1098) to recognize "that's a `sift_read` operation, not mine" as a more specific error than a generic unknown-operation message — but it never falls through to accepting it as valid. `sift_write` does **not** dispatch to `sift_read`'s handlers. See `tools/sift_types.go`'s doc comments and `tools/operation_validation_test.go`'s `TestDelegation_SplitNameDisjointOperations` (in #1098) for the pattern this follows.

## Test coverage: sift had none before this

`tools/sift.go` had zero unit tests before this PR — only a 139-line `cloud`-tagged integration suite (`sift_cloud_test.go`) requiring a live Grafana + Sift plugin. Added `tools/sift_unit_test.go`: httptest-mocked coverage of every handler, both dispatchers, and both params' `validate()`, including the disjoint-operations rejection case.

One cost worth flagging: `createSiftInvestigation`'s poll loop has a real 5-second ticker with no test seam, so 4 of the new tests each cost ~5s of wall-clock time (~20s added to `make test-unit` for this package). Kept redundant end-to-end coverage to a minimum (one dispatch-level test proving routing, not one per operation) to limit the cost rather than eliminating it — changing the ticker itself felt out of scope for a repackaging PR.

Also ported `sift_cloud_test.go` to the new types/entrypoints and added an entrypoint-level (`siftRead`/`siftWrite`) subtest there.

## Measured token/byte delta

`--enabled-tools sift`:

| | Tools | Bytes (write-enabled) | Bytes (read-only) |
|---|---|---|---|
| Before | 5 | 3,478 | 1,743 |
| After | 2 | 3,125 (**-10.1%**) | 1,397 (**-19.9%**) |

Positive in both configurations. (For contrast: `annotations`, reverted in #1098, measured +15.1%/+18.4% — worse in both. The difference is the real parameter overlap described above.)

Measurement command:
```bash
printf '%s\n%s\n' '{"jsonrpc":"2.0","id":1,"method":"initialize","params":{"protocolVersion":"2024-11-05","capabilities":{},"clientInfo":{"name":"x","version":"1"}}}' '{"jsonrpc":"2.0","id":2,"method":"tools/list","params":{}}' | ./dist/mcp-grafana -t stdio --enabled-tools sift 2>/dev/null | jq -c 'select(.id==2)|.result.tools' | wc -c
```

## BREAKING CHANGE

`get_sift_investigation`, `get_sift_analysis`, `list_sift_investigations`, `find_error_pattern_logs`, and `find_slow_requests` no longer exist as tool names. Clients must switch to `sift_read` (operations: `list_investigations`, `get_investigation`, `get_analysis`) and `sift_write` (operations: `find_error_pattern_logs`, `find_slow_requests`).

## Consumer updates

`README.md` (tool table + Sift Investigations section + Read-Only Mode list), `docs/sources/reference/mcp-tools-table.md`, `docs/sources/configure/command-line-flags.md`, `tests/disable_write_test.py` (both `write_tools`/`read_tools` lists, in both test functions).

## Test plan

- [x] `go vet ./...`
- [x] `make lint` (jsonschema, openapi, gofmt, golangci-lint)
- [x] `make test-unit` — all packages pass (runtime for `tools` package up ~20s from the new `sift` coverage, see above)
- [x] `go vet -tags cloud ./tools/...` compiles (sift's cloud-tagged integration suite)
- [x] Manually dumped `tools/list` for `--enabled-tools sift`, with and without `--disable-write`, to measure the deltas above
- [ ] Not run: `sift_cloud_test.go` and `tests/disable_write_test.py` (both require live Grafana credentials) — updated but unverified against a live instance


sd2k:
Closing, we'll probably remove Sift tools instead as Sift is deprecated.

PR #1111, opened 2026-08-24
fix(alerting): encode rule matchers in the JSON shape Grafana expects

### LINKED ISSUE #1110: matchers filter on alerting_manage_rules is silently ignored
### What happened?

Calling alerting_manage_rules with a matchers filter (e.g. space_name="Ribbon-BRS-01.S.01") returns the same unfiltered results as calling it with no matcher at all. No error, it just doesn't filter.

The tool sends the matcher to Grafana as JSON with capitalized field names, e.g. `{"Type":0,"Name":"space_name","Value":"..."}`. Grafana's API silently ignores this because it expects lowercase fields instead: `{"name":"space_name","value":"...","isRegex":false}`. When I sent the correctly-shaped JSON directly to Grafana API, filtering worked correctly.

### Steps to reproduce

Call this tool: `{"name":"alerting_manage_rules","arguments":{"rule_limit":200,"matchers":["space_name=\"Ribbon-BRS-01.S.01\""],"operation":"list","rule_type":"alerting","states":"firing"}}`

It returns every rule instead of just ones matching space_name.

### mcp-grafana version

v1.1.0

### Grafana version

cloud

### MCP client

claude code

### Transport

streamable-http

### How did you install it?

I installed using `uv` but that is not an option in the github issue template dropdown list.

### Operating system and CPU architecture

linux

sd2k:
## Summary

Fixes #1110. The `matchers` filter on `alerting_manage_rules` was silently ignored — passing a matcher returned every rule instead of only the matching ones, with no error.

## Root cause

Each `matcher` query parameter was built with `json.Marshal` on a Prometheus `labels.Matcher`, which has no JSON tags. That produced capitalized, numeric-typed JSON:

```json
{"Type":0,"Name":"space_name","Value":"..."}
```

Grafana's Prometheus rules API (`/api/prometheus/grafana/api/v1/rules`) unmarshals each matcher into an alertmanager `pkg/labels.Matcher`, which reads the lowercase `name`/`value` fields plus `isRegex`/`isEqual` booleans. The capitalized fields deserialize to an empty matcher, so the filter did nothing.

## Changes

- Encode matchers via a dedicated `ruleMatcherJSON` struct that emits the fields Grafana actually reads: `{"name":...,"value":...,"isRegex":...,"isEqual":...}`.
- Map the Prometheus match type onto `isRegex`/`isEqual`, covering all four operators (`=`, `!=`, `=~`, `!~`).
- Skip nil matchers defensively.
- Update existing client tests to assert the correct wire shape.

## Testing

- New unit tests cover all four match-type mappings and multiple matchers.
- Updated the two existing tests that had encoded the buggy format.
- `make lint` and `go test ./tools/...` pass locally.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (inline comments corrected)
- [x] Linting passes

### DIFF
--- a/tools/alerting_client.go
+++ b/tools/alerting_client.go
@@ -103,9 +103,33 @@ type GetRulesOpts struct {
 	RuleLimit    int      // Maximum number of rules to return (max 200)
 	LimitAlerts  int      // Maximum number of alert instances per rule (max 200)
 	// Matchers filters alert instances by labels. Each matcher is JSON-encoded
-	// as a Prometheus matcher object (e.g. {"type":0,"name":"severity","value":"critical"}).
+	// in the shape Grafana expects (e.g. {"name":"severity","value":"critical","isRegex":false,"isEqual":true}).
 	// Multiple matchers are AND-ed together.
 	Matchers []*labels.Matcher
+}
+
+// ruleMatcherJSON is the JSON shape Grafana's Prometheus rules API expects for
+// each `matcher` query parameter. Grafana unmarshals it into an
+// alertmanager pkg/labels.Matcher, which reads the lowercase name/value fields
+// plus the isRegex/isEqual booleans. Prometheus' own labels.Matcher marshals
+// with capitalized field names and a numeric Type instead, which Grafana
+// silently ignores — so we cannot json.Marshal the matcher directly.
+type ruleMatcherJSON struct {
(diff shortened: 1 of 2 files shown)


sd2k:
cursor review

gerboland (approved):

PR #1112, opened 2026-08-24
feat(run_panel_query): support PostgreSQL datasources

### LINKED ISSUE #1101: Add PostgreSQL datasource support to `run_panel_query`
## What problem are you trying to solve?

`run_panel_query` cannot execute dashboard panels backed by Grafana's built-in PostgreSQL datasource.

With the current PostgreSQL plugin ID, the panel query fails with an error such as:

```text
datasource type 'grafana-postgresql-datasource' is not supported by run_panel_query
```

Older and provisioned Grafana installations may expose the same datasource as `postgres`. Neither identifier is currently normalized or dispatched by `run_panel_query`.

The tool can retrieve the dashboard, select the panel target, and extract its `rawSql`, but it rejects the datasource before executing the query. This means an MCP client can inspect a PostgreSQL-backed dashboard without being able to retrieve the values shown by its panels.

This is particularly limiting for dashboards that combine observability data with operational data stored in PostgreSQL. An agent can query the Prometheus or Loki panels but cannot complete the same dashboard-level analysis for panels containing rankings, reconciliation results, execution quality, data-quality checks, or other SQL-backed data.

The standalone SQL tooling discussed in #146 and #652 addresses a different workflow. It does not replace the panel-aware behavior of `run_panel_query`, which loads a saved panel target and applies its datasource, time range, and dashboard-variable context.

## What would you like to see?

Extend the existing `run_panel_query` tool to support Grafana's PostgreSQL datasource identifiers:

- `grafana-postgresql-datasource`
- `postgres`

This should use the existing SQL panel-query path that already supports BigQuery and MSSQL, rather than introducing a new MCP tool or changing the input/output shape of `run_panel_query`.

The implementation should:

- normalize both PostgreSQL datasource identifiers to the same internal type;
- route PostgreSQL panel targets through Grafana's `/api/ds/query` endpoint;
- preserve the complete saved panel target, including PostgreSQL-specific fields;
- retain the existing dashboard-variable and frontend macro handling;
- allow Grafana's PostgreSQL backend to expand SQL macros such as `$__timeFilter`, `$__timeFrom`, `$__timeTo`, and `$__timeGroup`;
- support the panel's table or time-series format;
- return results through the existing `SQLQueryResult` representation;
- retain the existing per-panel error and partial-success behavior;
- list PostgreSQL in the tool schema and description;
- provide useful PostgreSQL-specific hints for empty results;
- add unit tests for type normalization, dispatch, target preservation, macro handling, request construction, result parsing, and empty-result hints.

If the repository's integration-test environment provides a PostgreSQL datasource, an integration test covering a representative panel query through `/api/ds/query` would also be valuable.

The BigQuery request and implementation in #922 and #930 provide a close precedent. The current shared `executeSQLPanelQuery` path used for BigQuery and MSSQL appears to provide the appropriate extension point.

## How are you working around it today?

The current workaround is to retrieve the dashboard or panel definition, extract the SQL target, and then either:

1. manually reconstruct a request to Grafana's `/api/ds/query` endpoint; or
2. execute the SQL through a separate database or SQL client.

Both approaches bypass `run_panel_query` and require the caller to reproduce panel-specific behavior such as target selection, time-range handling, dashboard-variable substitution, Grafana macro semantics, and datasource authentication.

## Would you be willing to implement this?

Yes, with some guidance

sd2k:
## Summary

Closes #1101.

`run_panel_query` could execute panels for Prometheus, Loki, ClickHouse, CloudWatch, InfluxDB, BigQuery and MSSQL, but rejected Grafana's PostgreSQL datasource before running the query:

```text
datasource type 'grafana-postgresql-datasource' is not supported by run_panel_query
```

Older/provisioned installations expose the same datasource as the legacy identifier `postgres`, which was also unhandled. This PR routes PostgreSQL panels through the **existing** SQL panel-query path already used for BigQuery and MSSQL — no new tool, no change to the input/output shape.

## Changes

- `normalizeDatasourceType` maps both `grafana-postgresql-datasource` and `postgres` to a single internal `postgres` type.
- `runSinglePanelQuery` dispatches `postgres` to `executeSQLPanelQuery`, passing the **resolved** type through so the datasource object sent to Grafana matches the panel's declared identifier.
- `defaultSQLFormat` returns the string `"table"` for PostgreSQL, matching its `sqleng` backend (like MSSQL), whose query model rejects a numeric format enum.
- The complete saved panel target is preserved; dashboard variables and the frontend `$__interval`/`$__range` macros are substituted locally, while SQL macros (`$__timeFilter`, `$__timeFrom`, `$__timeTo`, `$__timeGroup`) are left for the backend plugin to expand.
- Empty results surface PostgreSQL-specific hints.
- The tool schema and description now list PostgreSQL (and MSSQL, which was previously missing from the description).

## Testing

- `make test-unit` — all green.
- New unit tests cover type normalization, dispatch, target preservation, variable + macro handling, request construction, result parsing, and empty-result hints — the request-construction/result-parsing tests drive the real path against a mock `/api/ds/query` server for both the modern and legacy datasource identifiers.
- `gofmt`, `golangci-lint`, and the jsonschema/openapi linters pass.

No PostgreSQL datasource exists in the integration-test environment (neither did BigQuery/MSSQL), so coverage is unit-only, matching that precedent.

## Checklist

- [x] Tests added/updated
- [x] Documentation updated (tool description; the docs tool table is generic and needs no change)
- [x] Linting passes

### DIFF
--- a/tools/run_panel_query.go
+++ b/tools/run_panel_query.go
@@ -24,7 +24,7 @@ type RunPanelQueryParams struct {
 	End            string            `json:"end" jsonschema:"description=Override end time (e.g. 'now'\\, RFC3339\\, Unix ms)"`
 	Variables      map[string]string `json:"variables" jsonschema:"description=Override dashboard variables (e.g. {\"job\": \"api-server\"})"`
 	DatasourceUID  string            `json:"datasourceUid,omitempty" jsonschema:"description=Override datasource UID"`
-	DatasourceType string            `json:"datasourceType,omitempty" jsonschema:"description=Override datasource type (prometheus\\, loki\\, grafana-clickhouse-datasource\\, cloudwatch\\, influxdb\\, grafana-bigquery-datasource\\, mssql)"`
+	DatasourceType string            `json:"datasourceType,omitempty" jsonschema:"description=Override datasource type (prometheus\\, loki\\, grafana-clickhouse-datasource\\, cloudwatch\\, influxdb\\, grafana-bigquery-datasource\\, mssql\\, grafana-postgresql-datasource\\, postgres)"`
 }
 
 // QueryTimeRange represents the actual time range used for a panel query
@@ -226,6 +226,11 @@ func runSinglePanelQuery(ctx context.Context, params singlePanelQueryParams) (*P
 		results, err = executeSQLPanelQuery(ctx, datasourceUID, panelData, query, params.Start, params.End, vars, BigQueryDatasourceType)
 	case "mssql":
 		results, err = executeSQLPanelQuery(ctx, datasourceUID, panelData, query, params.Start, params.End, vars, MSSQLDatasourceType)
+	case "postgres":
+		// PostgreSQL exposes two datasource identifiers (grafana-postgresql-datasource
+		// and the legacy postgres); pass the resolved type through so the datasource
+		// object sent to Grafana matches what the panel actually declared.
+		results, err = executeSQLPanelQuery(ctx, datasourceUID, panelData, query, params.Start, params.End, vars, datasourceType)
 	default:
 		return nil, fmt.Errorf("datasource type '%s' is not supported by run_panel_query; use the native query tool (e.g. query_prometheus\\, query_loki_logs\\, query_clickhouse\\, query_cloudwatch\\, query_influxdb) directly", datasourceType)
 	}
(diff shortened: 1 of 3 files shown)


sd2k:
cursor review

gerboland (approved):
lgtm

PR #1119, opened 2026-08-25
feat(transport): support scoped SOCKS5 proxy via GRAFANA_SOCKS5_PROXY

### LINKED ISSUE #983: Support a scoped SOCKS5 proxy setting
Hey! I'd like to suggest one improvement.

## Suggested behavior

Allow `mcp-grafana` to use a SOCKS5 proxy configured directly in the MCP server env block, without relying on global `HTTP_PROXY` / `HTTPS_PROXY`.

```json
{
  "mcpServers": {
    "grafana": {
      "command": "uvx",
      "args": ["mcp-grafana"],
      "env": {
        "GRAFANA_URL": "http://localhost:3000",
        "GRAFANA_SERVICE_ACCOUNT_TOKEN": "<your service account token>",
        "GRAFANA_SOCKS5_PROXY": "socks5://127.0.0.1:1080"
      }
    }
  }
}
```

The proxy should apply only to Grafana requests made by `mcp-grafana`. If unset, current behavior should remain unchanged.

Open to a different name if preferred, for example `GRAFANA_PROXY_URL`.

## Benefits

- Keeps proxy configuration scoped to `mcp-grafana` instead of affecting the whole MCP client or shell session.
- Makes SOCKS5 proxy usage explicit in the `mcpServers.grafana.env` block.
- Avoids impacting other MCP servers or tools that should not use this proxy.
- Preserves current behavior when the SOCKS5 proxy variable is not set.

nk-tedo-001:
Closes #983.

Adds a `GRAFANA_SOCKS5_PROXY` environment variable that routes this server's Grafana traffic through a SOCKS5 proxy. The setting is scoped to mcp-grafana: it does not modify `HTTP_PROXY`/`HTTPS_PROXY`, and when it is unset nothing changes.

## How it works

- `GRAFANA_SOCKS5_PROXY` accepts `socks5://` and `socks5h://` URLs (Go's net/http treats the two schemes identically; the proxy resolves hostnames). Credentials in the URL work.
- The URL is validated at startup and the process refuses to start on an invalid value. Validation errors never include the raw value, since it may embed proxy credentials.
- `BuildTransport` applies the proxy to a clone of the innermost `*http.Transport`, below the TLS layer, so every client built through it (OpenAPI, datasource backends, rendering, proxied MCP, incident, OnCall, k8s) picks it up. No new dependencies: `http.Transport.Proxy` supports SOCKS5 natively.
- A `BaseTransport` that is not an `*http.Transport` cannot carry a proxy, so that combination is an explicit error instead of a silent replacement.

## Fail-closed behavior

A few call sites used to fall back to a default transport when building the custom transport failed (the incident clients, the cached incident client, OnCall, and the reflection path in `NewGrafanaClient`). With a proxy configured, that fallback would send traffic directly, so these paths now return an error instead of bypassing the proxy. Without a proxy their behavior is unchanged.

`proxiedToolSetKey` now includes the proxy URL, since two sessions with different proxies must not share a client. Logs only see a redacted present/absent flag.

## Testing

- Unit tests cover URL validation, `BuildTransport` clone semantics and TLS interaction, the base-transport conflict, fail-closed incident clients, the cache key, and log redaction through a real `slog` JSON handler.
- A behavioral test drives `NewGrafanaClient` against an `httptest` server plus a fake SOCKS listener and asserts every request (the public-URL fetch and OpenAPI calls) is dialed through the proxy, never directly.
- `go test -tags unit ./...`, `go build ./...`, and `golangci-lint` v2.11.4 (the CI version) pass locally.

### DIFF
--- a/README.md
+++ b/README.md
@@ -636,6 +636,32 @@ You can add arbitrary HTTP headers to all Grafana API requests using the `GRAFAN
 }
 ```
 
+### SOCKS5 Proxy
+
+You can route all requests this server makes to Grafana through a SOCKS5 proxy using the `GRAFANA_SOCKS5_PROXY` environment variable. The proxy is scoped to this server's Grafana traffic: it does not modify the global `HTTP_PROXY`/`HTTPS_PROXY` variables, and when set it overrides their proxy selection for Grafana transports only, without affecting other MCP servers or your shell session. When unset, behavior is unchanged.
+
+The URL must use the `socks5://` or `socks5h://` scheme (Go treats them identically: hostname resolution is delegated to the proxy) and may include credentials, e.g. `socks5://user:pass@127.0.0.1:1080`.
+
+**Example:**
+
+```json
+{
+  "mcpServers": {
+    "grafana": {
+      "command": "mcp-grafana",
+      "args": [],
+      "env": {
+        "GRAFANA_URL": "http://localhost:3000",
+        "GRAFANA_SERVICE_ACCOUNT_TOKEN": "<your token>",
+        "GRAFANA_SOCKS5_PROXY": "socks5://127.0.0.1:1080"
+      }
+    }
+  }
+}
+```
+
+An invalid proxy URL is a startup error, and if building a proxied connection fails at runtime the server fails closed rather than silently sending Grafana traffic directly.
+
 ### Forwarding Headers from the Client (SSE/Streamable-HTTP Only)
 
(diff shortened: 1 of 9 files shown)


sd2k (approved):
Nice thank you, looks good. I have a few refactors that a review revealed but I'll apply those separately. Thanks again!

nk-tedo-001:
> Nice thank you, looks good. I have a few refactors that a review revealed but I'll apply those separately. Thanks again!

Thanks for the quick review! One question - how often are releases published?

sd2k:
There's no fixed cadence, I intend to do one pretty soon but have a few other PRs I would like to get merged first. Certainly this week!

Issue #569, opened 2026-02-11
Reduce the size of the MCP (16K tokens)


JohnieLee:
Currently the MCP is ~16K tokens.  That's just short of 10% of available context for Claude.  

Total: 15,496 Tokens
```
MCP tools · /mcp
     └ mcp__ide__getDiagnostics: 112 tokens
     └ mcp__ide__executeCode: 183 tokens
     └ mcp__grafana__add_activity_to_incident: 175 tokens
     └ mcp__grafana__create_alert_rule: 593 tokens
     └ mcp__grafana__create_annotation: 222 tokens
     └ mcp__grafana__create_folder: 153 tokens
     └ mcp__grafana__create_graphite_annotation: 138 tokens
     └ mcp__grafana__create_incident: 326 tokens
     └ mcp__grafana__delete_alert_rule: 95 tokens
     └ mcp__grafana__fetch_pyroscope_profile: 545 tokens
     └ mcp__grafana__find_error_pattern_logs: 218 tokens
     └ mcp__grafana__find_slow_requests: 205 tokens
     └ mcp__grafana__generate_deeplink: 321 tokens
     └ mcp__grafana__get_alert_group: 99 tokens
     └ mcp__grafana__get_alert_rule_by_uid: 136 tokens
     └ mcp__grafana__get_annotation_tags: 102 tokens
     └ mcp__grafana__get_annotations: 336 tokens
     └ mcp__grafana__get_assertions: 268 tokens
     └ mcp__grafana__get_current_oncall_users: 125 tokens
     └ mcp__grafana__get_dashboard_by_uid: 130 tokens
     └ mcp__grafana__get_dashboard_panel_queries: 201 tokens
     └ mcp__grafana__get_dashboard_property: 250 tokens
     └ mcp__grafana__get_dashboard_summary: 115 tokens
     └ mcp__grafana__get_datasource_by_name: 115 tokens
     └ mcp__grafana__get_datasource_by_uid: 115 tokens
     └ mcp__grafana__get_incident: 91 tokens
     └ mcp__grafana__get_oncall_shift: 119 tokens
     └ mcp__grafana__get_panel_image: 474 tokens
     └ mcp__grafana__get_sift_analysis: 169 tokens
     └ mcp__grafana__get_sift_investigation: 165 tokens
     └ mcp__grafana__list_alert_groups: 475 tokens
     └ mcp__grafana__list_alert_rules: 424 tokens
[...]

PR #1134, opened 2026-08-29
Add delete_annotation tool

### LINKED ISSUE #1103: Expose annotation deletion (DeleteAnnotationByIDWithParams)
## Background

While consolidating the annotations domain for #1092 (see #1098, since reverted — that consolidation measured negative on tokens and was dropped, unrelated to this), I noticed Grafana's annotation-delete API is already fully wired up in the vendored client but was never exposed as an MCP tool:

- `grafana-openapi-client-go/client/annotations.Client.DeleteAnnotationByID` / `.DeleteAnnotationByIDWithParams`
- Today's four annotation tools (`get_annotations`, `get_annotation_tags`, `create_annotation`, `update_annotation`) cover list/tag/create/update, but there's no way to delete an annotation via MCP.

## Proposal

Add a `delete_annotation` tool (or equivalent), taking an annotation `id` (`int64`, required) and calling `c.Annotations.DeleteAnnotationByIDWithParams(annotations.NewDeleteAnnotationByIDParamsWithContext(ctx).WithAnnotationID(id))`.

This is a genuine new capability, not a repackaging of existing tools, so it should be judged and reviewed as its own change — sizing, RBAC scope (`annotations:write`, consistent with `create_annotation`/`update_annotation`), and whether it belongs behind `--disable-write` (yes, it's a write operation) can all be decided independently of any tool-surface consolidation work.

## Reference implementation

A working version of this (as `annotations_write`'s `delete` operation) existed briefly in #1098 before that PR's annotations consolidation was reverted; the diff is still visible in that PR's history (commit `42db07b`) if useful as a starting point, though it was written as part of a now-abandoned consolidated tool rather than a standalone one.

Tarasusrus:
## Summary

Grafana's annotation-delete API is already wired up in the client — `annotations.Client.DeleteAnnotationByIDWithParams` — but no MCP tool ever exposed it. `get_annotations`, `get_annotation_tags`, `create_annotation` and `update_annotation` covered list/tag/create/update, so deleting an annotation was the one operation callers could not reach.

This adds `delete_annotation`, shaped as described in #1103:

- takes the annotation `id` (`int64`, required)
- calls `c.Annotations.DeleteAnnotationByIDWithParams(annotations.NewDeleteAnnotationByIDParamsWithContext(ctx).WithAnnotationID(id))`
- registered behind `--disable-write` alongside `create_annotation` / `update_annotation`
- annotated destructive and non-idempotent, matching `delete_snapshot`

Two details worth flagging, both found while validating rather than assumed:

**RBAC action.** The issue suggested `annotations:write`, consistent with create/update, but Grafana gates this endpoint on a separate action. On Grafana 13.0.2, `DELETE /api/datasources/../api/annotations/<id>` answers `403 {"message":"You'll need additional permissions to perform this action. Permissions needed: annotations:delete"}`. The RBAC tables therefore list `annotations:delete`, not `annotations:write` — otherwise an operator building a least-privilege role from those tables would get a 403 on every call.

**Missing `id`.** `required` in the generated schema is advisory; nothing rejects a `tools/call` that omits the field, and the zero value would send `DELETE /api/annotations/0`, which Grafana answers with the permission error above — an access problem in the wording, a bad argument in reality. The tool rejects a non-positive id before the request goes out, the same way `delete_snapshot` guards its key.

## Token cost

The repo is token-sensitive (#569), so measured rather than guessed — `tools/list` dumped from binaries built before and after:

| | tools | `tools/list` bytes |
| --- | --- | --- |
| before | 70 | 88,732 |
| after | 71 | 89,155 |

+422 bytes (~106 tokens), +0.48% of the tool surface. The input schema is one required integer, about as small as a write tool gets. Happy to fold this into an operation-style tool instead if you would rather not spend a top-level slot — #1103 says "or equivalent", and the annotations consolidation in #1098 was reverted for measuring negative, so a standalone tool seemed the shape you wanted.

## Validation

- `go test -tags unit ./...` and `make lint` clean.
- Unit tests: the request is a `DELETE` to `/api/annotations/{id}`; a failure is wrapped as `delete annotation: …`; a zero or negative id is rejected without any request reaching Grafana.
- Integration test: creates an annotation, deletes it, then asserts `GetAnnotationByID` no longer finds it — so it proves the row is gone rather than trusting the tool's own success reply. Run against the Grafana 13.0.2 service from `docker-compose.yaml`; all `TestAnnotationTools` subtests pass.
- Mutation-checked: making `deleteAnnotation` report success without calling the API fails the integration test ("annotation N is still retrievable after deletion"); swapping the verb fails the unit test; dropping the error wrap fails the error test; disabling the id guard fails the guard test.

Docs updated everywhere a tool is listed: the README feature list, the README RBAC table, `docs/sources/reference/mcp-tools-table.md`, the `--disable-write` lists in the README and `docs/sources/configure/command-line-flags.md`, `tests/disable_write_test.py`, and a CHANGELOG entry.

Fixes #1103


sd2k (dismissed):
Nice thanks! There is a conflict with main unfortunately but I'll merge once that's sorted.

Tarasusrus:
Conflict with main resolved (rebased). Ready for re-review.

sd2k (approved):
Thanks!

PR #1135, opened 2026-09-01
feat(loki): add optional matcher to list_loki_label_names/values

### LINKED ISSUE #382: Loki: Add support for query filters in list_loki_label_values
Currently, the list_loki_label_values function retrieves label values without allowing any query-based filtering.

For example, it’s not possible to fetch label values (e.g., service) limited to a specific namespace.

According to the [Loki API documentation](https://grafana.com/docs/loki/latest/reference/loki-http-api/#query-label-values), the /label/{name}/values endpoint supports an optional query parameter that can be used for such filtering.

Adding support for the query parameter in list_loki_label_values would make it possible to narrow down label values based on specific criteria (like namespace, job, or other labels).

Ashfaqbs:
## Summary

Fixes #382

`list_loki_label_names` and `list_loki_label_values` previously listed labels/values across an entire Loki (or VictoriaLogs) datasource with no way to scope the search — e.g. there was no way to ask "what `service` values exist within a given `namespace`" without listing every value and filtering client-side.

Adds an optional `matcher` parameter, forwarded to the underlying API on both backends:
- **Loki**: the `query` parameter on `/labels` and `/label/<name>/values` (a LogQL stream selector, e.g. `{namespace="prod"}`)
- **VictoriaLogs**: the `query` parameter on `/select/logsql/field_names` and `/select/logsql/field_values` (a LogsQL filter), falling back to the existing "match everything" expression when `matcher` is empty so behavior is unchanged by default

This is a parameter addition to two existing tools (their purpose/shape is unchanged), so per CONTRIBUTING.md it doesn't require a new-tool proposal.

## Context

There's a stale PR (#380) that attempted the same thing against an earlier version of this file; a maintainer closed it for staleness in March and explicitly invited a redo ("feel free to pick this up again though, we'd be happy to review!"). This PR is a fresh implementation against the current codebase — the Loki tools have since moved to a `lokiBackend` interface (native Loki + VictoriaLogs), so the change threads `matcher` through both backend implementations rather than the old single `Client.fetchData` method #380 touched.

## Test plan

- [x] Added unit tests for both backends: `TestLokiClient_FetchData_PassesMatcherAsQueryParam` / `_OmitsQueryParamWhenMatcherEmpty` (native), `TestVictoriaLogsBackend_ListLabelNames_PassesMatcher` / `TestVictoriaLogsBackend_ListLabelValues_PassesMatcher` (VictoriaLogs), plus updated the two existing VictoriaLogs tests to assert the "match everything" fallback when `matcher` is empty.
- [x] Verified RED→GREEN: `git stash` on the source-only changes fails the build with unmatched call sites, confirming the tests exercise the new parameter; restoring the fix goes green.
- [x] Full unit suite (`go test -tags unit ./tools/...`) passes.
- [x] `go build ./...` and `go vet ./...` clean under both `unit` and `integration` build tags.
- [x] `gofmt -l` clean on all changed files.
- [x] Added an integration-tagged test case (`get loki label values narrowed by matcher` in `loki_test.go`) alongside the existing ones — **not run locally**, since it requires a real Loki instance; flagging this explicitly rather than claiming it passed.
- [x] CHANGELOG.md updated under `[Unreleased]`.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.1.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- Optional `matcher` parameter on `list_loki_label_names` and `list_loki_label_values` to narrow label discovery to a subset of streams (a LogQL stream selector on Loki, LogsQL on VictoriaLogs) ([#382](https://github.com/grafana/mcp-grafana/issues/382))
+
 ## [1.3.0] - 2026-08-28
 
 ### Added
--- a/tools/loki.go
+++ b/tools/loki.go
@@ -158,9 +158,15 @@ func (c *Client) makeRequest(ctx context.Context, method, urlPath string, params
 	return bytes.TrimSpace(bodyBytes), nil
 }
 
-// fetchData is a generic method to fetch data from Loki API
-func (c *Client) fetchData(ctx context.Context, urlPath string, startRFC3339, endRFC3339 string) ([]string, error) {
+// fetchData is a generic method to fetch data from Loki API. matcher, when
+// non-empty, is forwarded as the "query" parameter to narrow the label
+// search to streams selected by a LogQL stream selector (e.g. `{app="foo"}`),
+// per Loki's /labels and /label/<name>/values API.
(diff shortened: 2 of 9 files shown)


Ashfaqbs:
Fixed the failing integration test in 6fb59f0 — `{container="grafana"}` was an exact match against a container-name assumption that doesn't hold (Compose generates `<project>-grafana-<n>`, not the bare service name). Switched to `{container=~".*grafana.*"}`, matching the regex convention every other matcher literal in this file already uses. Was disclosed as not run locally in the original PR description (needs a real Loki instance) — CI catching it and this fix is exactly why that disclosure was there.

sd2k (approved):
Looks great, thank you!

PR #1137, opened 2026-09-02
feat: add healthz listener appart with metrics for mux, close #1090

### LINKED ISSUE #1090: `/healthz` cannot be bound to a different address than the MCP endpoint, unlike `/metrics`

We run `mcp-grafana` (streamable-http) behind an Envoy sidecar in the same Kubernetes pod. Envoy terminates OIDC/JWT auth for callers and proxies requests to `mcp-grafana` over `127.0.0.1`. We want to set `--address 127.0.0.1:<port>` so the MCP endpoint is only accessible via Envoy.

However, `/healthz` is currently registered on the same mux/address as the main endpoint (`cmd/mcp-grafana/main.go`):

```go
mux.Handle(endpointPath, withCallerAuth(callerToken, ...))
mux.HandleFunc("/healthz", handleHealthz)
```

Restricting `--address` to loopback also restricts `/healthz` to loopback. Kubernetes `httpGet` probes cannot reach it because the `kubelet` probes the pod IP from the host network namespace and cannot target `127.0.0.1` inside the container network namespace. As a result, we cannot lock down the MCP endpoint without breaking our liveness/readiness probes.

This limitation affects anyone running `mcp-grafana` in Kubernetes behind a trusted reverse-proxy sidecar.

A way to keep `/healthz` reachable independently of `--address`, similar to how `--metrics-address` already decouples `/metrics`. For example:

- Add a `--healthz-address` flag, symmetric to `--metrics-address`.
- Or serve `/healthz` on the `--metrics-address` listener when it's set (that listener is already exempt from Host/Origin validation), instead of adding a third flag.

We're open to other approaches.

One related point: `/healthz` is currently subject to `--allowed-hosts`/`--allowed-origins` validation only because it shares the same `mux` that `DNSRebindingProtectionMiddleware` wraps — there's no check specific to `/healthz`. `--metrics-address` avoids this the same way: it runs on a separate `http.Server` (`runMetricsServer`) that's never wrapped by that middleware, so it's exempt because it's a different listener, not because of a special case in the middleware. `/healthz` should probably follow the same rule if it moves to its own address. We wanted to call this out so it's an explicit decision, not an accidental side effect.

We can't separate `/healthz` from `--address`, so we plan to avoid `/healthz` entirely: keep `--address` at `127.0.0.1` for `/mcp`, move `/metrics` to its own address via `--metrics-address`, and point the readiness/liveness probe at `/metrics` instead of `/healthz`. `/healthz` is the endpoint meant for this, but `--metrics-address` is the only option available today.

Yes

leoswing:
## What this changes

Allow `/healthz` to bind separately from the MCP endpoint via `--healthz-address`, matching `--metrics-address`, so Kubernetes probes still work when MCP is loopback-only behind a sidecar.

Fixes #1090 

## How you tested it

Unit tests for route placement (`go test -tags unit ./cmd/mcp-grafana/`). Local binary: kubelet-style Host on MCP /healthz is 403; with `--healthz-address` the probe returns 200. Existing `--metrics` / `--metrics-address` paths unchanged.

## Checklist

- [x] CLA signed (required to merge)
- [x] README tool table updated, with RBAC permissions and scopes, if this adds a tool
- [x] Any datasource query execution respects `--disable-query` (and `--disable-write` too, if the query is passed through unfiltered)

### DIFF
--- a/README.md
+++ b/README.md
@@ -435,9 +435,9 @@ The `mcp-grafana` binary supports various command-line flags for configuration:
 
 **HTTP Transport Security (SSE / streamable-http only):**
 
-`Host`/`Origin` validation is enforced on *every* route on the listener — `/sse`, `/mcp`, `/healthz`, and `/metrics` — so a DNS-rebinding browser cannot reach any of them. Stdio transport is unaffected.
+`Host`/`Origin` validation is enforced on *every* route on the MCP listener — `/sse`, `/mcp`, and `/healthz` / `/metrics` when they share that listener — so a DNS-rebinding browser cannot reach any of them. Stdio transport is unaffected. `--healthz-address` and `--metrics-address` start a separate listener that is not wrapped.
 
-- `--allowed-hosts`: Comma-separated allowlist of `Host` header values. Defaults to loopback variants of `--address` (e.g. `localhost:8000,127.0.0.1:8000,[::1]:8000`). A value that parses to empty (unset, `,`, ` , `, etc.) also falls back to the defaults so a typo cannot silently disable the check. Requests with a `Host` header outside the allowlist are rejected with `403`. Pass `*` to disable the check — only safe when running behind a trusted reverse proxy that rewrites `Host`, or in an isolated network. K8s `httpGet` probes and external `/metrics` scrapes will need either an explicit hostname in this list, `*`, or a `tcpSocket` probe / a separate metrics port (`--metrics-address`).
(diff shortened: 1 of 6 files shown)


sd2k (approved):
Thanks!

leoswing:
> Thanks!

Glad I could help! Happy to contribute to Grafana community ~

PR #1139, opened 2026-09-02
fix(datasources): don't report frontend-only plugins as unhealthy

### LINKED ISSUE #1069: `check_datasources_health` reports datasource types that cannot have a backend (e.g. `alertmanager`) as unhealthy
## What happens

`check_datasources_health` reports the core **Alertmanager** datasource as unhealthy and counts it in its `unhealthy` tally, on a Grafana instance where that datasource is working normally:

```json
{"results":[
  {"uid":"alertmanager","name":"Alertmanager","type":"alertmanager",
   "error":"check datasource health alertmanager: HTTP 500: {\"statusCode\":500,\"messageId\":\"plugin.unavailable\",\"message\":\"Plugin unavailable\"}"},
  {"uid":"loki","name":"Loki","type":"loki","status":"OK","message":"Data source successfully connected."},
  {"uid":"prometheus","name":"Prometheus","type":"prometheus","status":"OK","message":"Successfully queried the Prometheus API."},
  {"uid":"victoriametrics","name":"VictoriaMetrics","type":"prometheus","status":"OK","message":"Successfully queried the Prometheus API."}
 ],"total":4,"checked":4,"healthy":3,"unhealthy":1,"hasMore":false}
```

The practical cost is that an agent reads this as a real outage and reports it as one — which is what prompted this investigation. There is no way for a caller to tell this apart from a genuinely unreachable datasource, because the result carries the same shape as one.

## Why it isn't a real failure

The Alertmanager datasource is **frontend-only**. In `grafana/grafana:13.1.3`, `public/app/plugins/datasource/alertmanager/plugin.json` declares neither `backend` nor `executable`, and the plugin directory holds only TypeScript sources — while `loki`, `prometheus`, `elasticsearch`, `graphite`, `jaeger` and `zipkin` all declare `"backend": true`. The same is true of `dashboard` and `mixed`.

`GET /api/datasources/uid/:uid/health` dispatches to the plugin's **backend** (`pkg/api/datasources.go` → `hs.pluginClient.CheckHealth(...)`, error → `translatePluginRequestErrorToAPIError`), so for these types it can only ever fail. Grafana's own "Save & test" for this datasource does not use that endpoint at all: `datasource/alertmanager/DataSource.ts` implements `testDatasource()` in TypeScript and issues a proxied `GET /api/v2/status`.

Both paths, same Grafana instance, same service-account token, same datasource:

| path | result |
|--------------------------------------------------------------|--------------------------------------------------|
| `GET /api/datasources/uid/alertmanager/health`                 | `500 {"messageId":"plugin.unavailable"}`         |
| `GET /api/datasources/proxy/uid/alertmanager/api/v2/status`    | `200`, Alertmanager 0.33.1, healthy              |
| `GET /api/datasources/uid/loki/health` (has a backend)         | `200 {"status":"OK"}`                            |

## Where it comes from in this repo

`tools/datasources.go` (v1.1.0) treats every non-200 as an error:

```go
if resp.StatusCode != http.StatusOK {
    body, _ := io.ReadAll(io.LimitReader(resp.Body, 1024))
    return nil, fmt.Errorf("check datasource health %s: HTTP %d: %s", args.UID, resp.StatusCode, strings.TrimSpace(string(body)))
}
```

and the bulk wrapper folds any error straight into `unhealthy`:

```go
if r.Error != "" || r.Status != "OK" { unhealthy++ } else { healthy++ }
```

`DatasourceHealthCheckResult.Status` is already commented `// "OK", "ERROR", or "UNKNOWN"`, but nothing in this path ever produces `UNKNOWN`.

## This is unlikely to be fixed on the Grafana side

grafana/grafana#83794 ("Support datasource health check for alertmanager datasource") asked Grafana to implement it, quoting the identical response body. It received no maintainer response, was marked stale after a year, and was auto-closed as `not planned` in May 2025. So the endpoint's behaviour for these types should be treated as fixed for the foreseeable future, and handled here.

## Suggested fix

Distinguish "unhealthy" from "cannot be health-checked", and detect it from plugin **capability** rather than from the error.

Classifying on the error body (`messageId: plugin.unavailable`) looks tempting but is unsafe: Grafana returns the same error for a registered backend plugin that is decommissioned or otherwise unavailable — a real failure (e.g. grafana/grafana#96980). Treating that as "not applicable" would silence genuine outages.

`GET /api/frontend/settings` already carries the capability per datasource, and is readable by a Viewer-role service account. On the instance above:

| datasource   | `meta.backend` | `meta.executable` |
|--------------|----------------|-------------------|
| Alertmanager | `false`        | —                 |
| Loki         | `true`         | `gpx_loki`        |
| Prometheus   | `true`         | —                 |

So `check_datasources_health` could skip — or report as `UNKNOWN`/not-applicable — any datasource whose plugin has `meta.backend == false`, and keep reporting everything else exactly as it does now. That endpoint is already being introduced for datasource resolution in #1015, so the two may want to share a fetch (and a cache).

Whichever shape you prefer, the summary counters should stop implying an outage: a caller acting on `unhealthy > 0` currently gets a false positive on any instance provisioning an Alertmanager datasource — including every kube-prometheus-stack deployment, where it is enabled by default.

Two questions before anyone writes code, since they affect consumers rather than internals:

1. Should these become a distinct `Status: "UNKNOWN"` with a new `unknown`/`skipped` counter, or simply be excluded from `unhealthy`? Either changes what `unhealthy > 0` means for existing callers.
2. What should happen when `/api/frontend/settings` is unavailable or forbidden (the case #1015 addresses)? Fall back to today's behaviour, or to error-based classification with the false-negative risk noted above?

Happy to open a PR if this is wanted and those two are settled.

## Environment

- mcp-grafana `1.1.0` (container image `grafana/mcp-grafana:1.1.0`), `--disable-write`, `--enabled-tools=search,datasource,prometheus,loki,alerting,dashboard,navigation`
- Grafana OSS `13.1.3`, authenticated as a service account with the **Viewer** role
- Also reproduced on Grafana OSS `12.4.1`; and `public/app/plugins/datasource/alertmanager/` still declares no `backend` on `grafana/grafana` `main`, so no released version behaves differently
- Alertmanager datasource provisioned by kube-prometheus-stack defaults (`uid: alertmanager`, `implementation: prometheus`, `handleGrafanaManagedAlerts: false`)

Ashfaqbs:
## Summary

Fixes #1069. `check_datasources_health` reports the built-in **Alertmanager** datasource (and any other frontend-only plugin) as unhealthy on every deployment that provisions one — including every `kube-prometheus-stack` default install — because `GET /api/datasources/uid/{uid}/health` always dispatches to a plugin **backend**, and that plugin type has none (`public/app/plugins/datasource/alertmanager/plugin.json` declares neither `backend` nor `executable`). The endpoint can only ever fail for it, regardless of whether the real service is reachable and working:

```json
{"uid":"alertmanager","name":"Alertmanager","type":"alertmanager",
 "error":"check datasource health alertmanager: HTTP 500: {\"messageId\":\"plugin.unavailable\",...}"}
```

An agent reading `unhealthy > 0` from this tool currently gets a false-positive outage report on any such deployment, with no way to tell it apart from a genuinely unreachable datasource.

grafana/grafana#83794 asked Grafana to implement a working health check for this plugin type and was closed `not planned`, so this needs handling on our side.

## Fix

`GET /api/frontend/settings` already reports `meta.backend` per datasource and is readable by a Viewer-role service account. `check_datasources_health` now fetches it once per bulk call and skips the health-endpoint request entirely for any datasource it marks `backend: false`, reporting `"status": "UNKNOWN"` for that entry instead — tallied in a new `unknown` field on the bulk result (`BulkDatasourceHealthResult`) rather than folding into `unhealthy`, so `unhealthy > 0` keeps meaning what callers already expect it to mean.

Classifying on the error body (e.g. `messageId: plugin.unavailable`) was considered and rejected, per the issue discussion: Grafana returns the same error for a registered backend plugin that's genuinely decommissioned or unavailable (a real failure), so that would silence genuine outages too. Capability-based classification (`meta.backend`) doesn't have that false-negative risk.

When `/api/frontend/settings` itself is unavailable to the token, every datasource falls back to today's unchanged behavior (nothing here is a hard dependency).

## Testing Plan

- Added `TestCheckDatasourcesHealth_FrontendOnlyPluginReportsUnknownNotUnhealthy`: a mock server where the health endpoint would report a frontend-only plugin as ERROR if asked; asserts the fix never calls it, reports `UNKNOWN`, and `unhealthy` stays 0 with `unknown` at 1.
- Added `TestCheckDatasourcesHealth_FrontendSettingsUnavailableFallsBackToNormalCheck`: no frontend/settings handler at all; asserts the bulk check still succeeds and behaves exactly as before.
- Verified RED: the two new tests fail to compile against pre-fix code (`result.Unknown` doesn't exist), confirming the behavior didn't already exist.
- Verified GREEN: all `check_datasources_health` tests pass (11/11), including every pre-existing test unchanged.
- Full `go test ./... -tags unit` run: one pre-existing failure (`TestVersionOutput`, both subtests) reproduces identically on unmodified `main` — a Windows-sandbox `exec: ... executable file not found in %PATH%` issue unrelated to this change.
- `go vet ./...` and `gofmt` clean on all touched files.
- Bug fix to an existing tool's internal classification logic — no new tool, no change to any tool's name, input shape, or purpose — so this is exempt from `CONTRIBUTING.md`'s new-tool proposal gate.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - `--loki-enforced-matchers`: operator-configured LogQL label matchers AND-ed into every native-Loki query (logs, stats, patterns, and label enumeration) to restrict which streams the server can read. Fails closed on unparseable queries and refuses VictoriaLogs datasources while set. Pair with `--disable-api` so the raw datasource proxy cannot bypass it. A companion `--loki-label-enumeration-fallback` controls label-enumeration behaviour under purely-negative matchers ([#978](https://github.com/grafana/mcp-grafana/pull/978))
 - `--instructions-append` flag to append operator-supplied text to the server instructions returned to MCP clients on initialize, so every connecting agent sees it — e.g. to explain that Loki reads are restricted by `--loki-enforced-matchers` ([#978](https://github.com/grafana/mcp-grafana/pull/978))
 
+### Fixed
+
+- `check_datasources_health` no longer reports a frontend-only datasource plugin (e.g. the built-in Alertmanager datasource) as unhealthy: those plugins have no backend to serve the health endpoint, so it always failed for them. Such datasources now report `"status": "UNKNOWN"` and are counted separately in a new `unknown` field on the bulk result, rather than inflating `unhealthy` ([#1069](https://github.com/grafana/mcp-grafana/issues/1069))
+
 ## [1.3.0] - 2026-08-28
 
 ### Added
--- a/tools/datasources.go
+++ b/tools/datasources.go
@@ -851,6 +851,13 @@ func checkDatasourcesHealth(ctx context.Context, args BulkCheckDatasourceHealthP
 	withoutBackend := datasourcesWithoutBackend(ctx)
 
 	results := make([]DatasourceHealthCheckResult, len(targets))
+	// noBackend tracks which results were skipped for lacking a backend
+	// component, distinct from a backend plugin's own health check
+	// legitimately returning a "UNKNOWN" status: both currently surface as
+	// Status == "UNKNOWN", but only the former belongs in the unknown tally
+	// below — folding a real inconclusive/failed backend check into it would
+	// hide it from Unhealthy.
+	noBackend := make([]bool, len(targets))
 	var wg sync.WaitGroup
 	for i, ds := range targets {
 		wg.Add(1)
@@ -861,6 +868,7 @@ func checkDatasourcesHealth(ctx context.Context, args BulkCheckDatasourceHealthP
 				r.Status = "UNKNOWN"
 				r.Message = "This datasource's plugin has no backend component, so it cannot be health-checked."
 				results[i] = r
+				noBackend[i] = true
 				return
 			}
 			health, err := checkDatasourceHealth(ctx, CheckDatasourceHealthParams{UID: ds.UID})
@@ -876,9 +884,9 @@ func checkDatasourcesHealth(ctx context.Context, args BulkCheckDatasourceHealthP
 	wg.Wait()
 
 	healthy, unhealthy, unknown := 0, 0, 0
-	for _, r := range results {
+	for i, r := range results {
 		switch {
-		case r.Status == "UNKNOWN":
+		case noBackend[i]:
 			unknown++
 		case r.Error != "" || r.Status != "OK":
 			unhealthy++
--- a/tools/datasources.go
+++ b/tools/datasources.go
@@ -769,7 +769,34 @@ type BulkDatasourceHealthResult struct {
 	Checked   int                           `json:"checked"` // Number of datasources health-checked in this page
 	Healthy   int                           `json:"healthy"`
 	Unhealthy int                           `json:"unhealthy"`
-	HasMore   bool                          `json:"hasMore"` // Whether more datasources exist beyond this page
+	// Unknown counts datasources whose plugin has no backend component (so
+	// the health endpoint can only ever fail for them, e.g. the built-in
+	// Alertmanager datasource type) and were therefore not asked at all,
+	// rather than genuinely unreachable. See issue #1069.
+	Unknown int  `json:"unknown"`
+	HasMore bool `json:"hasMore"` // Whether more datasources exist beyond this page
+}
+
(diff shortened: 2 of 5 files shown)


sd2k (approved):
LGTM!

PR #1142, opened 2026-09-06
fix(server): honor --allowed-hosts behind loopback proxies

kahirokunn:
## What this changes

Since #1097, a proxy connecting over loopback with `Host: mcp.example.com` gets 403 even with `--allowed-hosts mcp.example.com`, because the SDK rejects the Host after our middleware accepts it.

This lets the existing middleware enforce explicit `--allowed-hosts` values for SSE and streamable HTTP; unset or empty values keep the SDK's localhost protection.

## How you tested it

The regression test runs the binary over loopback for both transports. Verified that it fails before the fix and passes after it. It also covers `*`, empty values, Origin checks, and bearer authentication.

- `make build`
- `make test-unit`
- `make lint` with golangci-lint v2.11.4, matching CI

## Checklist

- [x] Commits are signed
- [x] CLA signed


sd2k (approved):
Looks good, thanks!

PR #1150, opened 2026-09-07
fix(oncall): handle user objects in on_call_now from IRM proxy API

petergreen86:
## Summary
- `get_current_oncall_users` could return an empty user list even when someone was on call, when routed through the IRM plugin proxy (OBO auth path).
- Root cause: the internal OnCall API returns `on_call_now` entries as user objects (e.g. `{"pk": "U1A2B3C4D5E6F", "username": "..."}`), not plain ID strings. `extractUserIDs` only handled `[]string`/`[]any` of strings, so every object entry was silently dropped.
- `extractUserIDs` now also extracts the `pk` field from object entries.

## Test plan
- [x] `go test ./tools/... -run TestExtractUserIDs -v` — added a case covering the object-entry shape
- [x] `go build ./...`
- [x] `go test ./tools/... -run "OnCall|OncallProxy|Proxy" -v`
- [x] Manually verified against a real schedule on irmplayground.grafana.net where a user was on call but the tool previously returned an empty list

### DIFF
--- a/tools/oncall_proxy.go
+++ b/tools/oncall_proxy.go
@@ -499,7 +499,8 @@ func proxyGetCurrentOnCallUsers(ctx context.Context, scheduleID string) (*Curren
 }
 
 // extractUserIDs extracts user ID strings from the on_call_now field,
-// which can be []string or []any depending on the API.
+// which can be []string or []any depending on the API. Each []any entry may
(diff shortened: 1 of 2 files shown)


sd2k (approved):
Thanks!

PR #1152, opened 2026-09-07
fix(server): apply --base-path/--endpoint-path to /healthz and /metrics

### LINKED ISSUE #1021: Apply `--base-path` to `/healthz` and `/metrics` so a single reverse-proxy rule covers the whole service

Ashfaqbs:
## What this changes

`/healthz` and `/metrics` were always mounted at the server root regardless of `--base-path` (sse) or `--endpoint-path` (streamable-http), so a single path-routed reverse-proxy rule for the configured base couldn't cover the whole service — only the MCP endpoint itself respected the prefix. This additionally mounts `/healthz` and `/metrics` under the configured base path (for streamable-http, the directory of `--endpoint-path`, e.g. `/my-base/mcp` → base `/my-base`), while keeping the existing root routes for backward compatibility, so nothing changes for anyone not using a base path.

Fixes #1021

## How you tested it

- `go build ./...` and `go test -tags unit ./cmd/mcp-grafana/...` (added `TestRegisterAtBasePath` covering the new helper: empty base, root base, a real base, and a base with a trailing slash).
- Manually ran the built binary for both transports and curled the endpoints:
  - `sse` with `--base-path /my-custom-base --metrics`: `/my-custom-base/healthz` → 200, `/my-custom-base/metrics` → 200, and `/healthz` / `/metrics` at root still 200.
  - `streamable-http` with `--endpoint-path /my-custom-base/mcp --metrics`: `/my-custom-base/healthz` → 200, root `/healthz` still 200.

## Checklist

- [x] `make lint` and `make test-unit` pass locally (ran `gofmt`, the repo's jsonschema/openapi linters, and `go test -tags unit ./cmd/mcp-grafana/...` directly; `golangci-lint` isn't installed in this environment but the change doesn't touch anything it would flag)
- [ ] README tool table updated, with RBAC permissions and scopes, if this adds a tool
- [x] Any write operation respects `--disable-write` — no write operation added
- [x] Any datasource query execution respects `--disable-query` — no query added

### DIFF
--- a/cmd/mcp-grafana/main.go
+++ b/cmd/mcp-grafana/main.go
@@ -9,6 +9,7 @@ import (
 	"net/http"
 	"os"
 	"os/signal"
+	"path"
 	"regexp"
 	"slices"
 	"strconv"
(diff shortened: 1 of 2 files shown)


gerboland (changes requested):
In my experience, it is convention to serve internal-only `/healthz` and `/metrics` from the root path only, while allowing the application endpoints to use a base prefix.

Those internal endpoints are consumed by internal infrastructure only, and should not be made available externally. It's harder to configure a reverse proxy to not expose them, if they're on the same base path as the application endpoints.


I must also flag this is extremely similar to existing PR https://github.com/grafana/mcp-grafana/pull/1033

Ashfaqbs:
Good catch on both counts, thank you.

On the convention: fair, and on reflection I weighted "single proxy rule for the whole service" (what the issue asked for) over "ops endpoints shouldn't ride the same externally-exposed prefix as the app" without really stress-testing that tradeoff. The second one is the more defensible default.

On #1033: you're right that it's the same fix, and it's clearly the more complete version of it — path-traversal-safe mount resolution, actual streamable-http `--base-path` support (not just riding `--endpoint-path`'s directory the way I did it), flag-parse-time validation instead of a runtime panic, and a month of review already spent getting the edge cases right. Splitting your attention across two PRs solving the same issue doesn't help you, so I'll close this one rather than ask you to referee between them.

Happy to help get #1033 across the line if that'd be useful once conflicts are resolved there — otherwise I'll leave it to @Tarasusrus and step back from this one.

PR #1153, opened 2026-09-08
fix(sift): add --enable-sift-investigations to keep Sift usable under --disable-write

### LINKED ISSUE #744: Sift investigation tools blocked by --disable-write despite being read-analysis operations
## Problem

When running with `--disable-write` but without `--disable-sift`, the Sift category registers only its read-only tools (`list_sift_investigations`, `get_sift_investigation`, `get_sift_analysis`) but not the tools that create investigations (`find_error_pattern_logs`, `find_slow_requests`).

This leaves users with list/get tools that have nothing to list or get — a broken half-set. A user who explicitly opts into Sift but opts out of writes gets no practical Sift functionality.

I understand that creating an investigation is technically a write operation (per 48a5979). The issue is that `--disable-sift` and `--disable-write` interact in a way that produces a useless tool surface rather than a clean all-or-nothing.

## Suggestion

When `--disable-write` is active, also suppress the read-only Sift tools (treat Sift as a unit), OR give Sift investigation-creation its own gate independent of `--disable-write` since it doesn't mutate Grafana resources (dashboards, alerts, datasources).

## Relevant code

https://github.com/grafana/mcp-grafana/blob/main/tools/sift.go#L417-L424

```go
func AddSiftTools(mcp *server.MCPServer, enableWriteTools bool) {
	GetSiftInvestigation.Register(mcp)
	GetSiftAnalysis.Register(mcp)
	ListSiftInvestigations.Register(mcp)
	if enableWriteTools {
		FindErrorPatternLogs.Register(mcp)
		FindSlowRequests.Register(mcp)
	}
}
```

Ashfaqbs:
## What this changes

`--disable-write` gates `find_error_pattern_logs`/`find_slow_requests` alongside the read-only Sift tools (`list_sift_investigations`, `get_sift_investigation`, `get_sift_analysis`), so anyone who opts into Sift but also wants read-only mode ends up with list/get tools that have nothing to list or get — a broken half-set.

These two tools only create ephemeral Sift investigation records via the Sift API (`siftClient.createSiftInvestigation`) — they never touch a Grafana dashboard, alert, or datasource, which is a materially different risk profile than the write operations `--disable-write` actually protects against. This adds `--enable-sift-investigations`, mirroring the existing `--enable-query` override for the raw-SQL query tools: purely additive, opt-in, and the default behavior for existing `--disable-write` users is unchanged.

Fixes #744

## How you tested it

- `go build ./...` and `go test -tags unit ./cmd/mcp-grafana/...`.
- Added 4 tests mirroring the existing `--enable-query` coverage: `--disable-write` removes the investigation tools but not the read tools, `--enable-sift-investigations` restores just those two without re-enabling real write tools, the flag alone is a no-op, and `--disable-sift` wins over it.
- Confirmed via the existing `registerAllCategories` test helper, which exercises the same `processTools`/`toolEntries` registration path the running server uses.

## Checklist

- [x] `make lint` and `make test-unit` pass locally (gofmt, the repo's jsonschema/openapi linters, `go vet`, and `go test -tags unit ./cmd/mcp-grafana/...`; `golangci-lint` isn't installed in this environment but the change doesn't touch anything it would flag)
- [ ] README tool table updated, with RBAC permissions and scopes, if this adds a tool
- [x] Any write operation respects `--disable-write` — `find_error_pattern_logs`/`find_slow_requests` still default to gated by `--disable-write`; the new flag is an explicit opt-in override, same pattern as `--enable-query`
- [x] Any datasource query execution respects `--disable-query` — no query added

### DIFF
--- a/README.md
+++ b/README.md
@@ -485,6 +485,7 @@ Caller authentication is enforced only when `--server-auth-token` is set. When i
 - `--disable-write`: Disable write tools (create/update operations)
 - `--disable-query`: Disable query tools (tools that execute a query against a datasource); metadata and discovery tools stay available
 - `--enable-query`: Keep the raw-SQL query tools (`query_clickhouse`, `query_snowflake`, `query_athena`, `query_influxdb`) registered even under `--disable-write`
+- `--enable-sift-investigations`: Keep the Sift investigation-creation tools (`find_error_pattern_logs`, `find_slow_requests`) registered even under `--disable-write`
 - `--disable-loki`: Disable loki tools
 - `--disable-elasticsearch`: Disable elasticsearch and opensearch tools
 - `--disable-quickwit`: Disable quickwit tools
@@ -548,6 +549,8 @@ When `--disable-write` is enabled, the following write operations are disabled:
 **Sift Tools:**
 - `find_error_pattern_logs` (creates investigations)
 - `find_slow_requests` (creates investigations)
+
+These only create ephemeral Sift investigation records via the Sift API — they never touch a Grafana dashboard, alert, or datasource. Without them, `list_sift_investigations`/`get_sift_investigation`/`get_sift_analysis` have nothing to list or get. Pass `--enable-sift-investigations` to keep them registered under `--disable-write`.
 
 **Snapshot Tools:**
 - `create_snapshot`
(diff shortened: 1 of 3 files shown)


gerboland (commented):
Thanks for the contribution! I do not want to introduce more single-use command line flags for each tool however.

What do you think of this approach instead? https://github.com/grafana/mcp-grafana/pull/1157
It should enable your usecase, and be more general.

Ashfaqbs:
That's a better design than a dedicated flag — closing this in favor of #1157, which I've pulled and verified covers this exact case (see my comment there).

PR #1154, opened 2026-09-08
feat(search): add folderUid, tag, and starred filters to search_dashboards

lixiuxiu:
## Summary

- Add optional `folderUid`, `tag`, and `starred` parameters to the existing `search_dashboards` tool and forward them to Grafana `/api/search` (`folderUIDs`, `tag`, `starred`).
- Always send `type=dash-db`, so a folder-only search still returns dashboards rather than mixing in folders.
- This extends an existing tool (no new default-on tool). It is complementary to #907, which builds a client-side catalog; this PR uses the server-side search API instead.

Grafana `/api/search` has no dashboard-version query parameter or version field on search hits, so `version` is not exposed.

## Test plan

- [x] Unit tests assert `folderUIDs`, `tag`, and `starred=true` are forwarded, omitted filters are not sent, and `type=dash-db` is set even when `query` is empty
- [x] `go test -tags unit ./tools -run TestSearchDashboards`
- [x] `make lint-jsonschema`
- [ ] `make lint` / `make test-unit` in CI after maintainer approval (fork workflows need approval)
- [ ] Sign the Grafana CLA if this is a first contribution


gerboland (approved):
Thank you!

PR #1156, opened 2026-09-08
feat(prometheus): add query_prometheus_batched to run multiple queries in one call

### LINKED ISSUE #964: Feature Request: Batched prometheus queries
Currently the `query_prometheus` tool only does one query at a time. This can lead to increased latency when many queries need to be done at once.

It would be helpful to add a `query_prometheus_batched` tool to do multiple queries in one go.

It would take in a list of annotated `query_prometheus` calls and output their results.

#### Example input

```json
{
  "query_foo": {
     "datasourceUid": ...
     "expr": ...
     ...
  },
  "query_bar": { ... }
}
```

#### Example output

```json
{
  "query_foo": {
    "data": [...]
  },
  "query_bar": {
    "error": ...
  }
}
```

Ashfaqbs:
## What this changes

`query_prometheus` only runs one query at a time, so an agent that needs several PromQL queries together (e.g. CPU and memory for the same panel) pays one full tool round-trip per query. This adds `query_prometheus_batched`: a map from a caller-chosen key to a `query_prometheus`-shaped query, run concurrently (bounded to 8 at once, capped at 20 queries per call), returning a map keyed the same way where each entry holds either that query's result or its own error — one failing query never fails the batch.

Fixes #964

## New or changed tools

**Proposal issue:** #964

| Tool name | New or changed | Proposed tier |
| --- | --- | --- |
| `query_prometheus_batched` | new | default-on (registered alongside `query_prometheus` whenever query tools are enabled) |

**Token cost:** I don't have `mcp-tokens` installed in this environment (it's a Rust binary via `cargo`/`dist`, not `go install`-able, and I didn't want to add a full Rust toolchain just to run one check), so I can't paste its exact output. As an honest proxy: the serialized `tools/list` entry for `query_prometheus_batched` is 2289 bytes of JSON (~570 tokens at a rough chars/4 estimate) versus 1826 bytes for the existing `query_prometheus` entry it reuses the schema from — so the added cost is roughly one extra tool listing of about the same size as `query_prometheus` itself, not a large new schema surface, since the batch's per-query shape is exactly `QueryPrometheusParams` (no new types). Happy to run the real `make token-check` number if a maintainer can point me at a prebuilt `mcp-tokens` binary, or I can install Rust and do it properly if that's preferred over this estimate.

**Why not extend `query_prometheus` itself?** Overloading it to accept either a single query or a map of queries would make its schema (and every caller's happy path) more complex for the common case, and would break the existing single-result response shape for every current caller. A separate tool keeps `query_prometheus` exactly as it is today.

**Read-only safety:** Purely a fan-out over the existing read-only `query_prometheus` path — no new write capability, gated behind `enableQueryTools` the same as `query_prometheus`/`query_prometheus_histogram` (excluded by `--disable-query`, and README's query-free-mode list is updated accordingly).

## How you tested it

- `go build ./...`, `go vet ./...`, `go test -tags unit ./...` (whole repo) — all clean.
- New unit tests in `tools/prometheus_unit_test.go` covering the concurrency/aggregation core (`runPrometheusQueryBatch`) with a fake query function, independent of any live datasource: per-key result/error isolation (one failing query doesn't affect others), exactly one result per input key, and — run under `-race` — that concurrency is actually bounded to the configured limit and actually reaches it (not silently serialized). Plus validation tests for the empty-batch and over-the-20-query-cap error paths.
- Manual end-to-end smoke test: built the binary, drove it over stdio with a raw `initialize`/`tools/list` JSON-RPC exchange, and confirmed `query_prometheus_batched` registers with the expected description, annotations, and a correctly-derived nested schema for `queries` (map value schema matches `query_prometheus`'s own params, generated automatically — no schema duplicated by hand).
- `gofmt`, the repo's `jsonschema`/`openapi` linters — clean. `golangci-lint` isn't installed in this environment, but the change doesn't touch anything it would flag.

## Checklist

- [x] README tool table updated, with RBAC permissions and scopes — added `query_prometheus_batched` to the tools table (same scopes as `query_prometheus`) and to the `--disable-query` exclusion list
- [x] Any write operation respects `--disable-write` — no write operation added
- [x] Any datasource query execution respects `--disable-query` — gated behind the same `enableQueryTools` as `query_prometheus`

### DIFF
--- a/README.md
+++ b/README.md
@@ -340,6 +340,7 @@ Scopes define the specific resources that permissions apply to. Each action requ
 | `get_datasource`                  | Datasources               | Get a datasource by UID or name                                                                              | `datasources:read`                                     | `datasources:uid:prometheus-uid`                    |
 | `get_query_examples`              | Examples*                 | Get example queries for a datasource type                                                                    | `datasources:read`                                     | `datasources:*`                                     |
 | `query_prometheus`                | Prometheus                | Execute a query against a Prometheus datasource                                                              | `datasources:query`                                    | `datasources:uid:prometheus-uid`                    |
+| `query_prometheus_batched`        | Prometheus                | Run multiple `query_prometheus`-shaped queries in one call                                                   | `datasources:query`                                    | `datasources:uid:prometheus-uid`                    |
 | `list_prometheus_metric_metadata` | Prometheus                | List metric metadata                                                                                         | `datasources:query`                                    | `datasources:uid:prometheus-uid`                    |
 | `list_prometheus_metric_names`    | Prometheus                | List available metric names                                                                                  | `datasources:query`                                    | `datasources:uid:prometheus-uid`                    |
 | `list_prometheus_label_names`     | Prometheus                | List label names matching a selector                                                                         | `datasources:query`                                    | `datasources:uid:prometheus-uid`                    |
@@ -589,6 +590,7 @@ When `--disable-query` is enabled, the following tools are not registered:
 
 **Prometheus Tools:**
 - `query_prometheus`
+- `query_prometheus_batched`
 - `query_prometheus_histogram`
 
 **Loki Tools:**
--- a/tools/prometheus.go
+++ b/tools/prometheus.go
@@ -15,6 +15,7 @@ import (
 	promv1 "github.com/prometheus/client_golang/api/prometheus/v1"
 	"github.com/prometheus/common/model"
 	"github.com/prometheus/prometheus/model/labels"
+	"golang.org/x/sync/errgroup"
 )
 
 var (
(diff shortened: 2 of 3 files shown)


sd2k:
Hi! I missed this issue when it came in but I'm going to close it and this PR, if this is needed then the harness/agent can just issue parallel tool calls.

If we were to revisit I would want to avoid adding a new tool and instead update the existing tool to accept multiple queries, but for now I don't think it's needed. Thanks though!

Ashfaqbs:
Makes sense, thanks for the quick look — parallel tool calls do cover this, and I hadn't weighed that against adding a new tool. Appreciate the steer on the preferred direction if this comes up again.

PR #979, opened 2026-07-02
feat(tools):add get_dashboard_version and list_dashboard_versions tools

lixiuxiu:
## What

Adds two new read-only MCP tools for inspecting Grafana dashboard version history:

- **`list_dashboard_versions`** – Returns a compact list of saved versions for a dashboard (version number, author, timestamp, save message, parent version). The full dashboard JSON is intentionally omitted to keep response sizes manageable.
- **`get_dashboard_version`** – Retrieves a specific version by number, including the full dashboard JSON stored in that version.

Both tools are registered in `AddDashboardTools`.

## Why

Dashboard version history is useful in agentic workflows, for example:
- Auditing what changed and who changed it
- Comparing the current dashboard against a previous version
- Restoring context when diagnosing a regression

## Changes

- `tools/dashboard_versions.go` – implementation of both tools
- `tools/dashboard_versions_unit_test.go` – unit tests covering path params, pagination, response structure, and input validation
- `tools/dashboard.go` – register the two new tools in `AddDashboardTools`

## Testing

Unit tests run with the `unit` build tag:
```bash
go test -tags unit ./tools/... -run TestListDashboardVersions
go test -tags unit ./tools/... -run TestGetDashboardVersion

PR #1158, opened 2026-09-09
feat(dashboard): optional version on get_dashboard_by_uid and list_dashboard_versions

lixiuxiu:
## Summary

- Adds an optional `version` argument to `get_dashboard_by_uid`. Omit it to fetch the current dashboard (existing k8s/legacy path). When set, the tool calls Grafana's `GET /api/dashboards/uid/{uid}/versions/{version}` API and returns that saved snapshot in the same `dashboard`/`meta` shape.
- Adds read-only `list_dashboard_versions` so agents can discover available versions. It returns compact metadata only (`version`, `createdBy`, `created`, `message`) — not the full dashboard JSON.
- This supersedes #979, following [@sd2k](https://github.com/sd2k)'s suggestion to fold `get_dashboard_version` into the existing get tool rather than adding a second get tool.

Restore (`POST .../restore`) is intentionally out of scope (write-gated).

## Test plan

- [ ] Unit tests: omitted `version` does not hit the versions API; a positive `version` hits `/versions/{n}` and skips the current dashboard endpoint; `version <= 0` and empty `uid` error without HTTP; `list_dashboard_versions` forwards `limit`/`start` and omits dashboard `data`.
- [ ] `make lint-jsonschema` and `make lint-openapi`
- [ ] Against a Grafana instance: `list_dashboard_versions` for a dashboard with history, then `get_dashboard_by_uid` with one of those version numbers, then without `version` to confirm current still works.


sd2k (commented):
## Review Summary

Well-structured PR — clean file separation, good Grafana 11/12 backward compat with the dual-format reader, thorough tests, and consistently updated docs. A few suggestions inline, nothing blocking.

**What's good:**
- The first-byte dispatch in `decodeDashboardVersionList` is clean and robust for the Grafana 11/12 compat.
- Folding version into `get_dashboard_by_uid` rather than a separate tool is the right call for agent ergonomics.
- Test coverage is thorough: path routing, parameter forwarding, edge cases (empty UID, zero version), Grafana 11 compat, data-stripping assertion.
- Read-only/idempotent/non-destructive annotations are correct.

tools/dashboard_versions.go
```diff
@@ -0,0 +1,199 @@
...
+	"List saved versions of a Grafana dashboard. Returns compact metadata: version number, author, timestamp, and save message. Use get_dashboard_by_uid with version to fetch a snapshot.",
+	listDashboardVersions,
+	mcp.WithTitleAnnotation("List dashboard versions"),
+	mcp.WithIdempotentHintAnnotation(true),
+	mcp.WithReadOnlyHintAnnotation(true),
+	mcp.WithDestructiveHintAnnotation(false),
+	mcp.WithOpenWorldHintAnnotation(false),
+)
+
+func fetchDashboardVersion(ctx context.Context, uid string, version int64) (*DashboardResponse, error) {
```
sd2k, line 94: **Note:** `fetchDashboardVersion` always uses the legacy `/api/dashboards/uid/.../versions/N` endpoint and never populates `APIVersion`, while the normal (non-versioned) path through `fetchDashboard` may use the Kubernetes API and return `apiVersion: "v1beta1"`. This means the response shape differs depending on whether `version` is passed.

Probably fine since the versions API is legacy-only, but a doc comment here explaining the intentional asymmetry would help a future reader not try to "fix" it.

tools/dashboard_versions.go
```diff
@@ -0,0 +1,199 @@
...
+	}
+
+	return &DashboardResponse{
+		Dashboard: spec,
+		Meta: &models.DashboardMeta{
+			Version:   v.Version,
+			CreatedBy: v.CreatedBy,
+			Created:   v.Created,
+		},
+		IsV2: isV2DashboardJSON(spec),
```
sd2k, line 127: nit: The `Meta` returned here only has `Version`, `CreatedBy`, and `Created`. The normal path populates the full `DashboardMeta` (slug, folder UID, URL, provisioned status, etc.). An agent using the versioned response and expecting `meta.folderUid` or `meta.url` would get zero values.

This is inherent to the versions API, but consider mentioning in the tool description that meta is partial for versioned fetches.

tools/dashboard.go
```diff
@@ -34,7 +34,8 @@ const dashboardAPIGroup = "dashboard.grafana.app"
 const dashboardReadVersion = "v1beta1"
 
 type GetDashboardByUIDParams struct {
-	UID string `json:"uid" jsonschema:"required,description=The UID of the dashboard"`
+	UID     string `json:"uid" jsonschema:"required,description=The UID of the dashboard"`
+	Version *int64 `json:"version,omitempty" jsonschema:"description=Optional saved version. Omit for the current dashboard"`
```
sd2k: suggestion: The `version` field has no `minimum` constraint in the jsonschema tag. Adding `minimum=1` would let MCP clients validate before calling, rather than relying on the server-side `version <= 0` check in `fetchDashboardVersion`.

tools/dashboard_versions_unit_test.go
```diff
@@ -0,0 +1,277 @@
...
+		server := httptest.NewServer(http.HandlerFunc(func(w http.ResponseWriter, r *http.Request) {
+			t.Error("unexpected HTTP call when version is zero")
+		}))
+		defer server.Close()
+
+		ctx := mockSearchCtx(server)
+		version := int64(0)
+		_, err := getDashboardByUID(ctx, GetDashboardByUIDParams{UID: "my-uid", Version: &version})
+		require.Error(t, err)
+		assert.Contains(t, err.Error(), "version must be a positive integer")
```
sd2k, line 103: nit: The guard in `fetchDashboardVersion` is `version <= 0`, which correctly rejects negatives, but this test only covers `version == 0`. Consider adding a `version == -1` subtest — an LLM could easily pass a negative value.

tools/dashboard_versions.go
```diff
@@ -0,0 +1,199 @@
...
+		if v == nil {
+			continue
+		}
+		summaries = append(summaries, DashboardVersionSummary{
+			Version:   v.Version,
+			CreatedBy: v.CreatedBy,
+			Created:   formatDashboardVersionTime(v.Created),
+			Message:   v.Message,
+		})
+	}
```
sd2k, line 75: Not a blocker, just worth noting: Grafana returns the full dashboard JSON in the `data` field of each version entry. The entire response is read into memory and deserialized — `data` is only discarded during the summary conversion. The optimization is presentation-only (context-window savings), not network-level. Could matter for dashboards with many versions and large panel configs.

lixiuxiu:
HI, sd2k, I fix the code according to your suggestion, Could you please review again, thank you very much!

sd2k:
@lixiuxiu It looks like there are some merge conflicts that need resolving before checks can run.

lixiuxiu:
Sure，I resolved the merge conflicts

tools/dashboard_versions.go
```diff
@@ -0,0 +1,203 @@
...
+		Meta: &models.DashboardMeta{
+			Version:   v.Version,
+			CreatedBy: v.CreatedBy,
+			Created:   v.Created,
+		},
+		IsV2: isV2DashboardJSON(spec),
+	}, nil
+}
+
+func dashboardVersionData(data models.JSON) (map[string]any, error) {
```
sd2k, line 131: This marshal/unmarshal round-trip isn't needed — `models.JSON` is `interface{}`, and `encoding/json` always deserializes JSON objects as `map[string]interface{}` recursively, so the runtime value is already the target type. A direct type assertion would avoid the extra allocation + reparse for large dashboards:

```go
spec, ok := data.(map[string]any)
if !ok {
    // fall back to marshal/unmarshal
}
```

The same pattern is already used at `dashboard.go:205`.

tools/dashboard.go
```diff
@@ -666,7 +670,7 @@ func sortArrayRemovesDescending(operations []PatchOperation) ([]PatchOperation,
...
+	"Retrieves the complete dashboard, including panels, variables, and settings, for a specific dashboard identified by its UID. Pass optional version to load a saved snapshot instead of the current dashboard. Saved snapshots are fetched through Grafana's legacy versions API, so 'apiVersion' is omitted and 'meta' is partial: only version, createdBy, and created are populated. The response includes 'apiVersion' and 'isV2': when 'isV2' is true the dashboard uses the v2 schema (panels live under 'elements' keyed by name, arranged by 'layout'; variables under 'variables'), otherwise it is classic v1 ('panels[]' with 'templating.list'). WARNING: Large dashboards can consume significant context window space. Consider using get_dashboard_summary for overview or get_dashboard_property for specific data instead.",
```
sd2k: Nit: the description reads a bit contradictory for LLM consumers — it first says "'apiVersion' is omitted" (for versioned fetches), then immediately says "The response includes 'apiVersion' and 'isV2'". Without a qualifier on the second sentence, an agent could assume `apiVersion` is always populated.

Suggestion: prefix the second sentence to make the scope explicit, e.g.:
> "For current (non-versioned) fetches, the response includes 'apiVersion' and 'isV2': ..."

tools/dashboard_versions.go
```diff
@@ -0,0 +1,203 @@
...
+func listDashboardVersions(ctx context.Context, args ListDashboardVersionsParams) ([]DashboardVersionSummary, error) {
+	if args.UID == "" {
+		return nil, fmt.Errorf("uid is required")
+	}
+
+	c := mcpgrafana.GrafanaClientFromContext(ctx)
+	params := dashboards.NewGetDashboardVersionsByUIDParamsWithContext(ctx).
+		WithUID(args.UID)
+
+	if args.Limit > 0 {
```
sd2k, line 45: Minor: `limit` and `start` accept negative values silently (the `> 0` guard treats them as unset). The `version` field on `GetDashboardByUIDParams` already has `minimum=1` in its jsonschema tag — adding the same here would let MCP clients reject invalid values before the call reaches the server:

```go
Limit int `json:"limit,omitempty" jsonschema:"minimum=1,description=..."`
Start int `json:"start,omitempty" jsonschema:"minimum=1,description=..."`
```

sd2k (commented):
A few review comments from going through the latest revision.

tools/dashboard.go
lixiuxiu: @sd2k Thank you for your feedback, fix the three part according to your comments, please help review

sd2k (approved):
LGTM, thanks!

PR #1165, opened 2026-09-10
feat(sift): add labelSelector param for pattern matching on label values

### LINKED ISSUE #1164: Add support for pattern matching with label values for find_error_pattern_logs tool

Being able to pattern match on a set of label values used when submitted to the find_error_pattern_logs tool using the MCP server. Currently only exact matching is supported with label values.

Functionality for pattern matching with label values submitted to the find_error_pattern_logs tool since it currently only supports exact match on label values.

Manually creating investigations or using LogQL queries with the query_loki_logs tool which isn't an exact replacement.

_No response_

sd2k:
## Summary
- Adds a `labelSelector` parameter to `find_error_pattern_logs` and `find_slow_requests` tools that accepts PromQL/LogQL stream selector syntax (e.g. `{namespace=~"prod.*", cluster="us-east-1"}`)
- Enables all four match operators: `=` (exact), `!=` (not equal), `=~` (regex), `!~` (negative regex)
- Parses the selector with the existing `prometheus/promql/parser` dependency and sends typed `inputs` directly to the Sift backend, bypassing its `ToInputs()` which hardcodes exact match — no Sift backend changes needed
- Keeps the existing `labels` map for backwards compatibility; `labelSelector` takes precedence when both are provided

## Test plan
- [x] Unit tests for selector parsing (exact, regex, negative regex, not-equal, multiple matchers, `__name__` filtering, invalid input, empty braces)
- [x] Unit tests for `resolveSiftInputs` (precedence, fallback, validation)
- [x] Cloud integration test for `labelSelector` (alongside existing `labels` map test)
- [ ] `go vet ./...` passes
- [ ] CI green

Closes #1164


sd2k:
cursor review

gitdoluquita (approved):
LGTM!

PR #1167, opened 2026-09-10
fix(server): skip default-org warning when dynamic multi-org is enabled

### LINKED ISSUE #1162: Suppress WARN log noise from --dynamic-multi-org default-org fallback

When running MCP v1.3.0 with the `--dynamic-multi-org` flag enabled, the following WARN log is emitted repeatedly:
 
```
No org ID found in request headers or environment variables, using default org. Set GRAFANA_ORG_ID or pass X-Grafana-Org-Id header to target a specific org.
```
 
This happens even though falling back to the default org is expected, normal behavior in this mode — not an error condition. The volume of these logs adds noise and makes it harder to spot genuinely actionable warnings.

Suppress this log message specifically when `--dynamic-multi-org` is enabled and no org ID is provided, since falling back to the default org is the intended behavior in that case rather than a misconfiguration worth warning about on every request.

No workaround

None

lazerg:
## What this changes

With `--dynamic-multi-org`, the org is picked per tool call via the `orgId` argument, so a connection that carries no `GRAFANA_ORG_ID` and no `X-Grafana-Org-Id` header is the normal setup rather than a misconfiguration. Both HTTP client-extraction paths still logged a WARN on every request in that case, drowning out warnings that are actually worth reading.

The warning is now skipped when `DynamicMultiOrgEnabled` is set. Everything else is unchanged: without the flag, a missing org still warns exactly as before.

Fixes #1162

## How you tested it

Unit tests only. `TestDefaultOrgWarningSuppressedWithDynamicMultiOrg` covers all three cases: the warning still fires with the flag off, and is silent with the flag on for both `ExtractGrafanaClientFromHeaders` and the cached `extractGrafanaClientCached` path. It fails on the current code and passes with the fix.

## Checklist

- [x] README tool table updated, with RBAC permissions and scopes, if this adds a tool — no new tool
- [x] Any write operation respects `--disable-write` — no write path touched
- [x] Any datasource query execution respects `--disable-query` (and `--disable-write` too, if the query is passed through unfiltered) — no query path touched

### DIFF
--- a/client_cache.go
+++ b/client_cache.go
@@ -358,7 +358,9 @@ func extractGrafanaClientCached(cache *ClientCache) httpContextFunc {
 	return func(ctx context.Context, req *http.Request) context.Context {
 		config := GrafanaConfigFromContext(ctx)
 		logger := config.LoggerOrDefault()
-		if config.OrgID == 0 {
+		// Under dynamic multi-org the org is chosen per tool call, so an unset
+		// connection-level org is the expected starting point, not a misconfiguration.
+		if config.OrgID == 0 && !DynamicMultiOrgEnabled {
 			logger.Warn("No org ID found in request headers or environment variables, using default org. Set GRAFANA_ORG_ID or pass X-Grafana-Org-Id header to target a specific org.")
 		}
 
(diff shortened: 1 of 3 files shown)


sd2k (approved):
Thanks 👍

PR #1168, opened 2026-09-11
fix(logging): demote missing-org-ID warning to Debug under dynamic multi-org

### LINKED ISSUE #1162: Suppress WARN log noise from --dynamic-multi-org default-org fallback

Ashfaqbs:
Fixes #1162.

## Problem

`ExtractGrafanaClientFromHeaders` and `extractGrafanaClientCached` (both in the HTTP context-func chain, so they run per request) log:

```
No org ID found in request headers or environment variables, using default org. Set GRAFANA_ORG_ID or pass X-Grafana-Org-Id header to target a specific org.
```

at `Warn` whenever `config.OrgID == 0`. With `--dynamic-multi-org` enabled, omitting the per-call `orgId` argument is the normal, expected way to target the connection's default org — it's not a misconfiguration — so this fires on essentially every call and drowns out genuinely actionable warnings, exactly as reported.

## Fix

Factored both call sites into a shared `warnOnMissingOrgID(logger, orgID)` helper in `orgid.go` (next to the `DynamicMultiOrgEnabled` flag it reads):

- `DynamicMultiOrgEnabled == false` (the default): unchanged — still logs at `Warn`, since a missing org ID in that mode may still indicate a misconfigured connection.
- `DynamicMultiOrgEnabled == true`: logs the same message at `Debug` instead, so it's still available for troubleshooting but doesn't spam normal operation.

## Tests

- `orgid_test.go`: new `TestWarnOnMissingOrgID` — a resolved org ID logs nothing in either mode; a missing org ID logs at `Warn` with the flag off and at `Debug` (never `Warn`) with it on.
- `go test .` — full package passes. `TestExtractGrafanaClientFromHeaders` (unchanged) continues to see the `Warn` in its default (non-dynamic) scenarios, confirming existing behavior is preserved.
- `go build ./...` clean.
- `go test ./cmd/mcp-grafana/...` has a pre-existing, unrelated failure (`TestHTTPAllowedHostsLoopbackProxy` needs a pre-built `mcp-grafana` binary on `PATH`) — confirmed present on unmodified `upstream/main` via `git stash`, not caused by this change.

### DIFF
--- a/client_cache.go
+++ b/client_cache.go
@@ -358,9 +358,7 @@ func extractGrafanaClientCached(cache *ClientCache) httpContextFunc {
 	return func(ctx context.Context, req *http.Request) context.Context {
 		config := GrafanaConfigFromContext(ctx)
 		logger := config.LoggerOrDefault()
-		if config.OrgID == 0 {
-			logger.Warn("No org ID found in request headers or environment variables, using default org. Set GRAFANA_ORG_ID or pass X-Grafana-Org-Id header to target a specific org.")
-		}
+		warnOnMissingOrgID(logger, config.OrgID)
 
 		u, apiKey, basicAuth, _ := extractKeyGrafanaInfoFromReq(req, logger)
 		key := cacheKeyFromRequest(u, apiKey, basicAuth, config.OrgID, req)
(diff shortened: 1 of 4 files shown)


sd2k:
Thanks, I think #1167 beat you to it though!

PR #1169, opened 2026-09-11
fix(run_panel_query): route on the datasource's real type to preserve Loki enforcement

madaraszg-tulip:
## What this changes

`run_panel_query` chose its executor from the caller-supplied `datasourceType` instead of the datasource's actual type. Since `--loki-enforced-matchers` enforcement lives only in the native Loki backend, a caller could declare a SQL/BigQuery/CloudWatch type for a **Loki** datasource UID and have the panel's LogQL run through `executeSQLPanelQuery` / `executeCloudWatchPanelQuery`, which POST to `/api/ds/query` directly and bypass enforcement. The docs even stated `run_panel_query` was unconditionally safe.

The fix resolves the type authoritatively from the datasource UID whenever the caller overrode the datasource (or no type is known yet) and routes on that. A caller-supplied type is used only as a fallback when the datasource can't be read; when enforcement is active and an unverified fallback type isn't `loki`, it fails closed rather than proxy — mirroring the existing VictoriaLogs guard in `lokiBackendForDatasource`. A type declared in the panel JSON with no override is still trusted as-is, since it comes from the dashboard rather than the caller.

## How you tested it

- New unit tests (`tools/run_panel_query_enforce_unit_test.go`, `-tags unit`): one asserts the resolved Postgres type wins over a spoofed `grafana-bigquery-datasource` override (and that the type sent to `/api/ds/query` is the resolved one); one asserts that an unreadable datasource under active enforcement with a non-Loki override fails closed and never reaches `/api/ds/query`.
- Reproduced the original bypass against a live Grafana before the fix: a `grafana-bigquery-datasource`-declared query over a Loki datasource UID returned `agent_safe="false"` streams that the native `query_loki_logs` path (enforced) refused. With the fix the override no longer diverts to the proxy path.
- `gofmt -l`, `make lint-jsonschema`, `make lint-openapi`, and `go test -tags unit ./...` all clean locally. (`golangci-lint` not run locally — no binary on hand; deferring to CI.)

## Checklist

- [x] CLA signed (required to merge)
- [x] README tool table updated, with RBAC permissions and scopes, if this adds a tool 
- [x] Any datasource query execution respects `--disable-query` (and `--disable-write` too, if the query is passed through unfiltered)

### DIFF
--- a/docs/sources/configure/command-line-flags.md
+++ b/docs/sources/configure/command-line-flags.md
@@ -177,7 +177,7 @@ Enforcement applies only to the Loki query tools. Other tools can reach Loki log
 - `--disable-sift`: Sift investigations analyze Loki logs server-side across all streams.
 - `--disable-assistant`: `ask_assistant` delegates to Grafana Assistant, which reads Loki server-side across all streams. It is only registered when write tools are enabled, so `--disable-write` closes it too.
 
-The server logs a warning at startup naming each of these that is still enabled. `run_panel_query` is safe (it reuses the enforced query path). Proxied tools currently expose only Tempo (traces), not Loki logs, so they are not a bypass today. Dashboard snapshots (`--disable-snapshot`) can embed log-panel data captured outside enforcement.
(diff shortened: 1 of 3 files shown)


sd2k (approved):
👍 thanks again!

PR #1178, opened 2026-09-14
fix(cmd): apply --base-path to /healthz and /metrics on the shared listener

### LINKED ISSUE #1021: Apply `--base-path` to `/healthz` and `/metrics` so a single reverse-proxy rule covers the whole service

Ashfaqbs:
Fixes #1021.

## What

`registerOps` mounted `/healthz` and `/metrics` at hardcoded root paths regardless of `--base-path`. Under a single path-routed reverse-proxy rule (e.g. `https://host/grafana-mcp/*` → this server), `/grafana-mcp/sse` worked but `/grafana-mcp/healthz` and `/grafana-mcp/metrics` 404'd, because the backend only served them at root — exactly as described in the issue.

## Fix

`registerOps` now takes `basePath` and prefixes both routes with it — but **only** when they stay on the shared main listener. A side listener started via `--healthz-address`/`--metrics-address` is meant to be hit directly (e.g. a container health check hitting the pod's port directly, bypassing the proxy entirely), so it intentionally keeps the unprefixed root path regardless of `basePath` — prefixing it too would break that direct-hit use case for no benefit.

`streamable-http` is left unprefixed on purpose: `--endpoint-path` defaults to `/mcp` and is a full mount path, not a directory-style prefix like SSE's `--base-path` — there's no clean, unambiguous prefix to reuse there. Happy to revisit if there's an appetite for adding an equivalent base-path concept to that transport, but didn't want to invent one unreviewed.

## Test plan

Added 3 new tests to the existing `TestRegisterOps_*` suite in `main_test.go`:
- `TestRegisterOps_BasePathPrefixesMainMuxRoutes` — `/my-custom-base/healthz` and `/my-custom-base/metrics` now work; the unprefixed root path correctly stops working once a base path is set.
- `TestRegisterOps_RootBasePathDoesNotDoubleSlash` — the default `basePath="/"` (set by `run()` when `--base-path` is unset) doesn't produce `//healthz`.
- `TestRegisterOps_BasePathDoesNotAffectSideListener` — `--healthz-address` keeps serving at root even when `--base-path` is also set.

```
go test ./cmd/mcp-grafana/... -run TestRegisterOps -v
```

All 9 `TestRegisterOps_*` tests pass (6 existing + 3 new). Also ran the full `cmd/mcp-grafana` suite; unrelated to this change, `TestHTTPAllowedHostsLoopbackProxy` and `TestVersionOutput` fail in this Windows dev environment on unmodified `main` too (a pre-existing `exec.Command` binary-path resolution issue on Windows, confirmed by stashing this change and re-running) — everything else passes.


Ashfaqbs:
Fixed the trailing-slash issue Bugbot flagged — `pathFor` now trims a trailing slash off `basePath` before joining, so `--base-path=/foo` and `--base-path=/foo/` both resolve to `/foo/healthz`/`/foo/metrics` correctly. Added a regression test (`TestRegisterOps_BasePathTrailingSlashDoesNotDoubleSlash`) covering it.

gerboland:
Thank you for the contribution but https://github.com/grafana/mcp-grafana/pull/1033 did appear first.

PR #1182, opened 2026-09-14
fix(loki): fail closed on Loki datasources reached outside the enforced path

madaraszg-tulip:
## What this changes

`--loki-enforced-matchers` is applied only in the native Loki backend (`loki.go`, via the datasource proxy). Any *other* route to a Loki datasource bypasses it. #1169 fixed one such route (`run_panel_query` dispatching on a caller-supplied `datasourceType` and reaching Loki through `/api/ds/query`); this change closes the whole **class** at the two shared chokepoints so a future tool can't reopen it.

**This is hardening, not an active-leak fix.** With #1169 merged there is no known remaining route that leaks Loki data among the built-in tools: `query_prometheus` against a Loki datasource just 404s, and the other `/api/ds/query` tools build their own query models (`rawSql`, CloudWatch dimensions, ES DSL) that carry no LogQL `expr` for the Loki plugin to execute. The point here is to make that safety structural rather than incidental — so the next tool that builds an `/api/ds/query` payload with an `expr`, or a future Loki version that answers the Prometheus query endpoint, can't silently reopen the bypass.

- **`doDSQuery` (`/api/ds/query`)** — this endpoint routes by datasource UID to the target plugin, which for Loki executes the query's `expr` field. The enforced Loki path never uses `/api/ds/query`, so when enforcement is active, any `loki`/`victorialogs` datasource reaching this helper is by definition an unenforced route and is refused. The type is resolved from the **UID**, never the caller-supplied payload `type` (that's what #1169 exploited). A query naming no UID is checked against the org default datasource; an unresolvable type fails closed. One guard covers every `/api/ds/query` tool at once — cloudwatch, influxdb, elasticsearch, sql, run_panel_query, and any future one.
- **`backendForDatasource`** — reject `loki`/`victorialogs` the same way `tempo` already is, so `query_prometheus` returns a clear "not a Prometheus-compatible datasource" error instead of falling through to the Prometheus client (the route that currently 404s).

Not in scope: `grafana_api_request` (the `api` category) POSTs a raw body straight to `/api/ds/query` without going through `doDSQuery`, so this guard doesn't see it — it remains the documented full bypass gated by `--disable-api`.

## How you tested it

- New unit tests (`-tags unit`) for both chokepoints: Loki/VictoriaLogs refused before any request is sent; non-log datasources still pass through under enforcement; no-enforcement is a no-op (no lookup); unverifiable datasource fails closed; no-UID query decided by the org default datasource (refused when default is a log DS, allowed otherwise); `backendForDatasource` rejects Loki.
- `gofmt`, `go build ./...`, `go test -tags unit ./...`, and golangci-lint (CI-pinned v2.11.4) all clean.
- Full `-tags integration ./...` against the docker-compose stack: no new failures vs. `main` (touched paths — cloudwatch/influxdb/sql/prometheus/run_panel_query/datasource — 61/61 green; pre-existing unrelated failures confirmed identical on `main`).

## Checklist

- [x] Commits are signed
- [x] CLA signed
- [x] README tool table updated, with RBAC permissions and scopes, if this adds a tool (no tool added)
- [x] Any datasource query execution respects `--disable-query`


tools/ds_query.go
```diff
@@ -24,6 +24,81 @@ func dsQueryPayload(from, to time.Time, queries ...map[string]interface{}) map[s
...
+			}
+			continue
+		}
+		if checked[uid] {
+			continue
+		}
+		checked[uid] = true
+		ds, err := getDatasourceByUID(ctx, GetDatasourceByUIDParams{UID: uid})
+		if err != nil {
+			return fmt.Errorf("loki label-matcher enforcement is active: refusing /api/ds/query for datasource %q whose type could not be verified: %w", uid, err)
```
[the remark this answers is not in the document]
madaraszg-tulip: Good catch — fixed in 01f9865. The guard now treats Grafana's magic `default` UID the same as an absent UID: it resolves the org default datasource and decides on its real type, so a non-log default is allowed through and only a log default is refused. Added unit coverage for the `default` UID against both a log and a non-log org default.

tools/ds_query.go
```diff
@@ -24,6 +24,83 @@ func dsQueryPayload(from, to time.Time, queries ...map[string]interface{}) map[s
...
+		return fmt.Errorf("loki label-matcher enforcement is active but the /api/ds/query payload could not be inspected for its target datasources")
+	}
+	defaultChecked := false
+	checked := make(map[string]bool, len(queries))
+	for _, q := range queries {
+		uid := dsQueryDatasourceUID(q)
+		if uid == "" || uid == "default" {
+			// No explicit datasource, or Grafana's magic "default" UID: the
+			// query resolves against the org default. Check that default's type
+			// once.
```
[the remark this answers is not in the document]
madaraszg-tulip: Fixed in b61f59b. I verified against Grafana 13: `/api/ds/query` treats `default` as a literal datasource UID (404s when none exists), so you're right that a literal `default` Loki datasource would have been checked against the org default's type. The guard now resolves `default`/absent the way Grafana does — a datasource whose literal UID is `default` wins when present, otherwise the org default — and checks that datasource's real type. Added a test for a literal `default`-UID log datasource with a non-log org default.

tools/datasources.go
```diff
@@ -72,6 +72,52 @@ type ListDatasourcesResult struct {
...
+// check always reflects the datasource the query will actually hit.
+func effectiveDefaultDatasourceType(ctx context.Context, uid string) (string, error) {
+	list, err := allDatasources(ctx)
+	if err != nil {
+		return "", err
+	}
+	defType := ""
+	haveDefault := false
+	for _, ds := range list {
+		if uid == "default" && ds.UID == "default" {
```
sd2k: **medium: a literal `"default"`-UID datasource makes the guard skip the org default, re-opening the bypass.**

This returns as soon as it finds a datasource whose UID is literally `default`, so only *one* of the two resolutions Grafana might perform is checked. The added test covers the case where the literal `default` DS is Loki and the org default is Prometheus, but not the inverse.

Probe confirmed: with `[{uid: "default", type: "prometheus"}, {uid: "lokids", type: "loki", isDefault: true}]` and a payload naming `uid: "default"`, the guard returns `nil` and the request is sent to `/api/ds/query` — so on any Grafana that resolves `"default"` to the org default (the behaviour the doc comment itself attributes to older versions) the LogQL `expr` reaches Loki unenforced.

To actually fail closed, refuse if *either* candidate is a log datasource: the literal `default` DS **or** the org default.

tools/ds_query.go
```diff
@@ -24,6 +24,84 @@ func dsQueryPayload(from, to time.Time, queries ...map[string]interface{}) map[s
...
+// lokiLikeDatasourceType reports whether a resolved datasource type is a log
+// datasource whose reads must go through the enforced Loki path.
+func lokiLikeDatasourceType(t string) bool {
+	return t == "loki" || t == victoriaLogsDatasourceType
+}
+
+// dsQueryDatasourceUID extracts the datasource UID from one /api/ds/query query
+// object, tolerating both the map[string]string and map[string]interface{}
+// shapes the various tools build.
+func dsQueryDatasourceUID(q map[string]interface{}) string {
```
sd2k: **medium: this fails *open* for any `datasource` shape it doesn't recognise.**

An unrecognised shape returns `""`, which the caller below can't distinguish from "this query names no datasource" — so it gets checked against the *org default* and allowed if that isn't a log DS.

Probe confirmed: a query with `"datasource": "loki-uid"` (the pre-8.3 bare-string form — Grafana's query service still falls back to `query.Get("datasource").MustString()` as the UID when `datasource.uid` is empty) passes the guard and reaches `/api/ds/query` while the org default is Prometheus. Same hole for `datasource` as a map whose `uid` is a non-string or absent.

Suggest returning `(uid string, ok bool)` and having the guard fail closed when the `datasource` key is present but unparseable.

tools/ds_query.go
```diff
@@ -24,6 +24,84 @@ func dsQueryPayload(from, to time.Time, queries ...map[string]interface{}) map[s
...
+		return nil
+	}
+	queries, ok := payload["queries"].([]map[string]interface{})
+	if !ok {
+		return fmt.Errorf("loki label-matcher enforcement is active but the /api/ds/query payload could not be inspected for its target datasources")
+	}
+	defaultChecked := false
+	checked := make(map[string]bool, len(queries))
+	for _, q := range queries {
+		uid := dsQueryDatasourceUID(q)
```
sd2k: **low/medium: the legacy `datasourceId` field is never inspected.**

Grafana's `/api/ds/query` resolver falls back to the numeric `datasourceId` on a query when no `datasource.uid` is present. Probe confirmed a payload of `{"refId": "A", "datasourceId": 7, "expr": "{app=\"secret\"}"}` sails through the guard (checked against the Prometheus org default) and the request is sent.

Not reachable via today's built-in tools — but `run_panel_query` forwards raw panel targets verbatim (`substituteTemplateVariablesInMapWithValues(panelData.RawTarget, …)`, overriding only `target["datasource"]`), and this PR's stated purpose is to close the class structurally for future tools. Worth refusing outright when a query carries `datasourceId` under enforcement.

tools/ds_query.go
```diff
@@ -24,6 +24,84 @@ func dsQueryPayload(from, to time.Time, queries ...map[string]interface{}) map[s
...
+// type cannot be resolved fails closed rather than being forwarded. A query that
+// names no UID — or the magic "default" UID — is resolved the way Grafana
+// resolves it (a literal "default" datasource if one exists, otherwise the org
+// default) and its real type is checked the same way; an unresolvable default
+// also fails closed.
+func guardEnforcedLokiDSQuery(ctx context.Context, payload map[string]interface{}) error {
+	if len(enforcedMatchers(ctx)) == 0 {
+		return nil
+	}
+	queries, ok := payload["queries"].([]map[string]interface{})
```
sd2k: **low: `[]map[string]interface{}` is a brittle concrete-type assertion that fails *all* traffic, not just Loki.**

Every caller happens to go through `dsQueryPayload`, so this holds today. But the guard is billed as future-proofing, and the failure mode for a future caller that builds the envelope as `[]interface{}`/`[]any` (e.g. after a JSON round-trip) or as a typed slice is that *every* `/api/ds/query` tool — CloudWatch, SQL, InfluxDB, OpenSearch, `run_panel_query` — hard-fails with "the payload could not be inspected"… but only in deployments with `--loki-enforced-matchers` set, i.e. the configuration least likely to be exercised in tests.

Handling `[]any` too (extracting each element as `map[string]any`) removes that cliff.

sd2k (commented):
Reviewed the guard for fail-open paths. Four findings, each confirmed with a throwaway probe test against this branch (since removed). The three medium/low-medium ones are cases where a payload still reaches a log datasource with enforcement active; the last is a robustness cliff.

Things I checked and found clean:
- All six existing `/api/ds/query` payload builders (cloudwatch, influxdb, sql_tools, es_backend_opensearch, cloudmonitoring, victoriametrics) set `datasource.uid`, so the guard introduces no false positives on the default-datasource branch.
- `backendForDatasource`'s new loki/victorialogs rejection: only `tools/prometheus.go` calls it (5 sites), all PromQL tools — no path legitimately routed Loki through the Prometheus client.
- `getDatasourceByUID`'s 403 → frontend-settings fallback means query-only RBAC tokens still resolve types (Grafana filters `/api/frontend/settings` datasources by *query* permission), so failing closed shouldn't break restricted-token deployments.
- `go vet -tags unit ./tools/` and the new tests pass.

tools/datasources.go
madaraszg-tulip: Fixed in 44c396f. Default resolution now checks *both* candidates — a literal `default` datasource and the org default — and refuses if either is a log datasource, rather than returning on the first match. Added the inverse test (literal `default` = Prometheus, org default = Loki → refused).

tools/ds_query.go
madaraszg-tulip: Fixed in 44c396f. `dsQueryDatasourceUID` now returns `(uid, ok)`; the guard fails closed when `ok` is false. A `datasource` object present but unparseable (missing or non-string `uid`) is `ok=false`, and the bare-string form is honoured as a UID. Tests added for the bare-string Loki case and the non-string-uid shape.

tools/ds_query.go
madaraszg-tulip: Fixed in 44c396f. A query carrying a numeric `datasourceId` returns `ok=false` from the extractor and the guard fails closed (we can't cheaply map the id to a type). Test added.

tools/ds_query.go
madaraszg-tulip: Fixed in 44c396f. The query list is now read from both `[]map[string]interface{}` and `[]interface{}` (post-JSON-round-trip) via `dsQueryQueries`; a genuinely unknown shape still fails closed. Test added for the `[]interface{}` envelope.

tools/ds_query.go
```diff
@@ -24,6 +24,147 @@ func dsQueryPayload(from, to time.Time, queries ...map[string]interface{}) map[s
...
+//
+// ok is false when a datasource reference is present but not understood, so an
+// unrecognised shape fails closed instead of being treated as "no datasource"
+// and waved through against the org default.
+func dsQueryDatasourceUID(q map[string]interface{}) (uid string, ok bool) {
+	// A legacy numeric datasourceId can name any datasource, including a log
+	// one, and we cannot resolve it to a type cheaply. Refuse under enforcement.
+	if _, hasLegacyID := q["datasourceId"]; hasLegacyID {
+		return "", false
+	}
```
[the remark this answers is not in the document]
madaraszg-tulip: Fixed in 05ef951. Resolution now follows Grafana's order: `datasource.uid` (or the bare-string form) is used first when non-empty, and `datasourceId` is only consulted when the datasource field yields no uid (still failing closed there, since the numeric id can't be mapped to a type here). So a query carrying both a real uid and a leftover `datasourceId` — as `run_panel_query`'s copied panel targets do — resolves on the uid. Added a test for that case.

PR #1188, opened 2026-09-15
feat(usagestats): add anonymous usage statistics reporting

dafydd-t:
## Summary

- New `usagestats/` package that reports anonymous usage statistics about the MCP server itself. It emits one event per server process, sent every four hours and once on shutdown.
- Counters are deltas (`calls` / `errors` per tool) that reset on send, not on acknowledgement. Raw counts, no buckets. 
- Users can set usage stats to `enabled|disabled|log`. Theres a  `--usage-stats` flag, a nd then `GRAFANA_USAGE_STATS` env var. log prints the event to **stderr** only.
- **Ships default-disabled.** The receiver isn't live yet, so `DefaultMode` is `disabled` and a code comment at its definition says the flip to enabled belongs to a separate later release.
- **Nothing an MCP client reports about itself is collected** — not its name, not its version. Everything else that comes from outside the binary is clamped to a vocabulary in it: `auth_method` to a fixed list, and every proxied tool to the single pseudo-name `proxied` — never a proxied tool's real name and never the datasource type.
- A field with nothing to say is absent rather than empty, so it reads as NULL at the receiver instead of `''` or `{}`. `grafana_version` and `auth_method` are additionally omitted when the process resolved more than one distinct value, rather than inventing a "mixed" sentinel.
- Never on the wire: the Grafana URL/hostname/stack slug/org ID/org name, the value of `--server-name`, `--instructions-append` or the address flags, anything from `GRAFANA_EXTRA_HEADERS` / `GRAFANA_FORWARD_HEADERS`, error messages, tool arguments, credentials.
- The allowlist clamp and the `other` sentinel are now exported from `observability` (`BoundedValue`, `ValueSet`, `ValueOther`) rather than duplicated. A second copy of an allowlist is a privacy bug waiting to happen: the copy that stops being updated is the one that starts emitting user-supplied strings.
- New docs page `docs/sources/anonymous-usage-statistics.md`, cross-referenced from `docs/sources/configure/command-line-flags.md` and `README.md`.

- Nothing is per user, per session or per client: no session identifier is on the wire, no field says which clients connected, and a tool call cannot be attributed to a particular client.

The receiver endpoint (`https://stats.grafana.org/mcp-grafana-usage-report`) lands separately in `grafana/usage-stats`.


README.md
```diff
@@ -1190,6 +1193,27 @@ curl http://127.0.0.1:8080/healthz
...
+# Turn it off (or GRAFANA_USAGE_STATS=disabled)
+mcp-grafana --usage-stats=disabled
+
+# Print what would be sent, to stderr, and send nothing
+GRAFANA_USAGE_STATS=log mcp-grafana
+```
+
+`GRAFANA_USAGE_STATS_ENDPOINT` changes the destination. It is not an opt-out.
+
+For the full field list, what is never sent, how to read the data and its limitations, see [Anonymous usage statistics](https://grafana.com/docs/mcp-grafana/latest/anonymous-usage-statistics/).
```
sd2k: Need to make sure this URL is correct before merging

sd2k: would be something like this for now

```suggestion
For the full field list, what is never sent, how to read the data and its limitations, see [Anonymous usage statistics](https://grafana.com/docs/grafana/latest/developer-resources/mcp/anonymous-usage-statistics/).
```

usagestats/event.go
```diff
@@ -0,0 +1,203 @@
...
+	ProcessID       string `json:"process_id"`
+	ReportReason    string `json:"report_reason"`
+	ProcessUptimeMS int64  `json:"process_uptime_ms"`
+
+	// Tool usage since the previous flush, aggregated over the whole process.
+	// Both are omitted when no tool was called in this window, so "no tools
+	// called" reads as NULL rather than as an empty string and an empty
+	// object.
+	ToolsCalled string               `json:"tools_called,omitempty"`
+	ToolCalls   map[string]ToolCount `json:"tool_calls,omitempty"`
```
sd2k, line 117: I wonder what the query will have to look like to convert this into something useful, probably some fancy BigQuery `JSON_EXTRACT` and `UNNEST`? Not quite sure what the alternative is though, maybe we just have to accept that.

usagestats/event.go
```diff
@@ -0,0 +1,203 @@
...
+// The unit is the PROCESS, not the MCP session. Protocol version 2026-07-28
+// removed protocol-level sessions (SEP-2567), so mcp-go registers no session
+// for a client using it and a per-session event would never be built at all on
+// the streamable-http transport — see the Reporter doc comment. Tool-call
+// hooks still fire for those requests, so the counts below are aggregated over
+// every session and every sessionless request the process served.
+//
+// Fields are flat scalars apart from ToolCalls, which the receiver stores as a
+// JSON column.
+type Event struct {
```
sd2k, line 95: It'd be nice to have a `reported_at` field to account for any late-delivered events or queued shutdowns, to make the timestamps more accurate (delivery timestamp != reported timestamp).

cmd/mcp-grafana/main.go
```diff
@@ -1158,7 +1262,21 @@ func main() {
 	}
 
 	setFlags := map[string]bool{}
-	flag.Visit(func(f *flag.Flag) { setFlags[f.Name] = true })
+	setFlagNames := []string{}
+	flag.Visit(func(f *flag.Flag) {
+		setFlags[f.Name] = true
+		setFlagNames = append(setFlagNames, f.Name)
+	})
```
sd2k, line 1295: I think we miss configuration set using env vars here right? which will likely be the case for docker/k8s deployed servers, so we'll be underreporting things for those.

usagestats/event.go
sd2k, line 95: It'd also be nice to have a monotonic `report_seq` counter to identify missed reports. Not a blocker at all though.

usagestats/event.go
```diff
@@ -0,0 +1,203 @@
...
+	Version    string
+	AuthMethod string
+}
+
+// grafanaCloudHostSuffix is the only positive signal for a Grafana Cloud
+// stack that can be read from configuration alone. A Cloud instance reached
+// through a custom domain therefore reports self_hosted; see the docs page,
+// which states the misclassification rather than implying target_kind is
+// authoritative.
+const grafanaCloudHostSuffix = ".grafana.net"
```
sd2k: worth including .grafana-ops.net and .grafana-dev.net here too.

sd2k: also of note, this misses Grafana Cloud instances behind vanity URLs, but I'm not sure there's much we can do about that!

usagestats/event.go
```diff
@@ -0,0 +1,203 @@
...
+// is the honest answer — a "mixed" sentinel would sit in the same column as
+// real values and be counted as one.
+func soleValue(values map[string]struct{}) string {
+	if len(values) != 1 {
+		return ""
+	}
+	for v := range values {
+		return v
+	}
+	return ""
```
sd2k, line 220: this is unreachable I think?

usagestats/reporter.go
```diff
@@ -0,0 +1,570 @@
...
+				// tools capability is unsupported or the tools/call payload
+				// fails to unmarshal. No tool was named, so there is nothing
+				// to attribute; counting it would report a phantom proxied
+				// call, since toolKey("") falls through to the sentinel.
+				if req.Params.Name == "" {
+					return
+				}
+				r.recordToolCall(ctx, req.Params.Name, true)
+			},
+		},
```
sd2k, line 376: I don't think it's possible right now, but is there a chance that these hooks could both fire in future, resulting in double counts for erroring tool calls?

OTOH the only way I can think to avoid that is deduplicating by request ID which would mean keeping hold of all request IDs, which feels unreasonable.

sd2k (commented):
Looks good in general, minimal telemetry sent which I approve of!

I wonder if we should obey `DO_NOT_TRACK=1` and treat that as disabling telemetry, too? https://donottrack.sh/

sd2k:
Some additional comments from Codex:

  - [P2] Honor query and write gates in enabled category reports — /home/ben/repos/grafana/mcp-grafana/cmd/mcp-grafana/main.go:456-457
    With --enabled-tools=assistant --disable-write, this reports assistant as enabled even though no tools are registered. Likewise,
    --disable-query removes every tool from elasticsearch, quickwit, influxdb and runpanelquery while these categories remain reported
    as enabled. This skews dashboard comparisons of tool availability versus usage. Apply the effective gates already handled by
    buildInstructions, rather than checking only category selection.

  - [P2] Derive proxied_enabled from the actual server toggle — /home/ben/repos/grafana/mcp-grafana/cmd/mcp-grafana/main.go:1056-1056
    With --enabled-tools=search and no --disable-proxied, this records proxied_enabled=false. However, newServer configures
    WithProxiedTools(!dt.proxied), and both transport paths enable discovery based on that disable flag alone. Reports therefore claim
    proxied tools are disabled while the server can use them. Mirror the actual toggle so dashboard configuration breakdowns remain
    accurate.

  - [P2] Validate the transport before starting usage reporting — /home/ben/repos/grafana/mcp-grafana/cmd/mcp-grafana/main.go:1044-1044
    When reporting is enabled, an invalid --transport value reaches the reporter before the final transport switch rejects startup. The
    deferred shutdown flush then emits that value verbatim; --transport=https://private-tenant.example reproduces this in log mode. This
    bypasses the promised fixed vocabulary and can disclose mistakenly supplied configuration. Validate the transport before
    constructing the reporter, or clamp the wire value.

usagestats/event.go
dafydd-t: There's a 1s cap on sending these (needs a response in that time too), I don't think there's a queue in place really, so I'm not sure its worth it?

usagestats/event.go
dafydd-t: agree for -dev and -ops. Not sure what we can do about the others!

usagestats/event.go
dafydd-t: it is, but I think it needs some kind of return statement after the for loop

usagestats/event.go
dafydd-t: we have a flat `tools_called` for a list of tools, I guess the `toolCalls` count will need something more complicated, yeah.

README.md
dafydd-t: yeah `/docs/grafana/latest/developer-resources/mcp/anonymous-usage-statistics/` looks legit. I'm going to update the copy in this doc file

cmd/mcp-grafana/main.go
dafydd-t: Added `env_set` to capture the names of the configuration environment variables that are set (not values).

usagestats/event.go
dafydd-t: added `report_seq`, numbering each process's reports from 1.

usagestats/event.go
dafydd-t: `target_kind` now matches `.grafana-ops.net` and `.grafana-dev.net` alongside `.grafana.net`.

dafydd-t:
Thanks - addressed the codex issues too.

Implemented `DO_NOT_TRACK` as well, we have this in gcx too its consistent.

dafydd-t:
Reworded the docs files 👍

usagestats/reporter.go
dafydd-t: Fair, i think it might be the best of two bad choices like this though?

sd2k (dismissed):
LGTM!

grafsean:
@dafydd-t I reviewed and made minor edits to docs LGTM 👍

grafsean (approved):
Docs approved LGTM

ioanarm (approved):

PR #1196, opened 2026-09-17
fix(loki): preserve structured metadata in compact format

### LINKED ISSUE #1195: `query_loki_logs` `format: "compact"` silently drops all structured metadata (pod, node, detected_level ...)
Follow-up to #990, which added `format` to `query_loki_logs`. I left this as a [comment on that PR](https://github.com/grafana/mcp-grafana/pull/990#issuecomment-5587249620) after it was merged; opening it as its own issue so it has a place to live.

**This is not blocking us** — we handle it in our own agent's code, which refuses the argument before the call. I'm filing it anyway in case it's useful to anyone else running Loki with structured metadata. Still present as of 1.4.2.

The tool description recommends compact for broad queries, so our agent can pick it on its own. We never opted in.

In our Loki, `pod`, `node` and `detected_level` are structured metadata, not stream labels — so compact drops them. The agent then cannot answer "which pod is failing?". Nothing fails and no warning is returned; the answer just quietly loses the thing the question was about. In one test, four different distributor pods collapsed into a single stream with no pod anywhere in the response. And broad queries — the case compact is meant for — cover more pods per label set, so they lose more, not less.

Emitting the metadata once per stream would not work either: some keys are constant within a group (`deployment`, `detected_level`), but `pod` and `node` change line by line.

Could compact split it instead? Constant keys once in the stream header, changing keys on the line. Still much smaller, but nothing lost. I have measurements from our data if that helps.

If that is too much, could compact be disabled server-side, so operators decide instead of the model?

Thanks!

sd2k:
## Summary

- `query_loki_logs` with `format: "compact"` was silently dropping all structured metadata and parsed labels (e.g. `pod`, `node`, `detected_level`), causing agents to lose critical context
- Compact format now partitions metadata: constant keys (same value on every line) are hoisted to the stream header, varying keys are kept per-line
- This preserves compact's size benefit while losing no data

Fixes #1195

## Changes

- **`tools/loki.go`**: Added `StructuredMetadata` and `Parsed` fields to `CompactStream` and `CompactLine`. Rewrote `compactLogEntries` with a two-pass partition: identify constant vs varying keys per stream, then place them accordingly. Added helper functions `partitionConstant`, `cloneMap`, `pruneChanged`, `varyingOnly`.
- **`tools/loki_unit_test.go`**: Updated existing compact tests and added new test cases covering: constant metadata hoisted to stream, varying metadata kept per-line, mixed constant/varying, single-entry streams, and metadata keys present on some lines but not others.

## Test plan

- [x] Unit tests pass (`TestCompactLogEntries` — 8 subtests)
- [x] `go vet ./tools/` clean
- [ ] CI passes
- [ ] Integration tests with real Loki (existing test covers compact format normalization)


sd2k:
cursor review

gerboland (approved):

PR #1208, opened 2026-09-17
test(cmd): build the test binary with the host OS executable suffix

comedianhhh:
## What this changes

`TestVersionOutput` and `TestHTTPAllowedHostsLoopbackProxy` build the CLI into a temp dir and then exec it. The output path was hardcoded as `mcp-grafana`, so on Windows `exec` fails with `executable file not found in %PATH%` because the file has no `.exe` suffix. This adds a small `testBinaryPath(t)` helper that appends `.exe` on `windows` (and joins with `filepath`), and uses it at the three call sites. No test logic changes.

The release pipeline ships Windows binaries, so `go test ./...` should pass there too.

## How you tested it

- Windows 11, Go 1.26.5 (the version in `go.mod`): before the change `go test ./...` had 2 failures in `cmd/mcp-grafana` (both subtests of the two functions above); after the change the full suite passes.
- `go vet ./...` and `gofmt` clean on the changed file.
- Linux/macOS behaviour is unchanged (the suffix is empty there).

## Checklist

- [x] `make lint` and `make test-unit` pass locally (`go vet` + `go test ./...` on Windows; `make` targets are Linux-shaped)
- [x] Commits are signed (SSH signature; the key is being registered as a signing key on my account)
- [x] README tool table updated, with RBAC permissions and scopes, if this adds a tool — n/a
- [x] Any write operation respects `--disable-write` — n/a
- [x] Any datasource query execution respects `--disable-query` — n/a


sd2k (approved):

sd2k:
Thanks!

PR #1217, opened 2026-09-21
fix(prometheus): bound metric name discovery requests

dafydd-t:
`list_prometheus_metric_names` currently downloads every metric name before applying its regex and pagination. Large catalogs can therefore produce large temporary allocations even when the caller requests only 10 results.

We see this in the form of memory spikes on the cloud mcp server, which causes OOMKills:

Screenshot 2026-09-21 at 11 37 41

We can fix this by putting more controls into the request, to be handled by the server:

- Sending regex filtering, using `match[]`. The regex matching behaviour should be the same as that in the previous client-side regex filtering (see **regex matching** below)
- Request at most `page * limit` names with `promv1.WithLimit`
- Put a reasonable limit on the number of things requested: Reject requests with over 10,000 names (using page * limit query params). We also reject invalid values (< 0) for page and limit
- Cloud Monitoring queries still use local filtering because its descriptor does not support `limit` and `match` params.

Prometheus doesn't really support offsets, so to get the "second page" of results when there are 10 results per page, then we just need to fetch 20 results. 10 is the default page size.

For example, `regex=http`, `limit=10`, `page=2` now requests the first 20 matching names upstream and returns names 11–20. Previously it fetched the entire catalog first.

The memory reduction depends on the datasource honouring the upstream `limit` parameter. We still enforce trimming to `limit` on the client, but we won't get any performance benefit if the server ignores these query parameters. **Regex matching is not re-done on the client, so the server MUST implement regex filtering.**

**regex matching**

Go’s MatchString (used in the old client-side filtering) searches anywhere in the name, whereas Prometheus regex matchers require the whole name to match. This PR accounts for that while preserving explicit anchors and flags:

  - http still matches my_http_requests_total
  - ^http only matches names starting with http
  - http$ only matches names ending with http
  - (?i)http still matches case-insensitively

  without that adjustment, sending http directly to Prometheus would only match a metric named exactly http

**Adding some metrics**

This PR also adds two native histograms to count:
- number of metric names requested (page * limit)
- response size (bytes)

These will help us understand if our limits on total metrics names is reasonable. It needs https://github.com/grafana/grafana-assistant-app/pull/10845 to get the provider wired in, but either can be merged first

Cloud Monitoring histogram coverage includes plain/chunked/gzip responses, empty catalogs, malformed JSON, incomplete reads, and HTTP errors. The response-size and count histograms use backend-neutral names because both backends now contribute.


tools/prometheus.go
```diff
@@ -188,6 +188,8 @@ var QueryPrometheus = mcpgrafana.MustTool(
 	mcp.WithOpenWorldHintAnnotation(false),
 )
 
+const maxPrometheusMetricNames = 10_000
```
dafydd-t, line 191: kind of picked at random, we dont have any data for historical response sizes here

dafydd-t: added some metrics so we can count this in the future

dafydd-t:
cursor review

sd2k (dismissed):
LGTM! ty

sd2k (commented):
hmm agent came back with some things:

       ●  42 [correctness] metricNamesResponseBody only records the byte histogram in Read() when err == io.EOF, but Close() is not
                           overridden, so closing the body early (context cancellation, timeout) drops the measurement entirely.
     tools/prom_backend_cloudmonitoring.go
       ● 187 [correctness] recordPrometheusMetricNames is only called from prometheusBackend.MetricNames (prom_backend.go:191).
                           cloudMonitoringBackend.MetricNames at this line never calls it, creating an observability blind spot for CM
                           datasources.
     tools/prometheus.go
       ● 216 [correctness] When limit > maxPrometheusMetricNames (e.g. 15000) and page=1, integer division makes the check pass (1 >
                           10000/15000 = 0), but the error says 'page * limit must not exceed 10000; narrow the regex or request an
                           earlier page' -- misleading because the limit alone is the problem.
       ● 257 [correctness] The list_prometheus_metric_names tool description says it supports 'an optional time range to restrict
                           results to metrics active within that window' but cloudMonitoringBackend.MetricNames ignores both time
                           parameters (named _ in the signature), returning all metrics.
     tools/prom_backend.go
       ● 122 [efficiency]  prometheusMetricNamesTransport is applied to the entire Prometheus API client, so every Query, LabelNames,
                           LabelValues call passes through its path-suffix check. The overhead is negligible but a bug in the
                           transport would break all Prometheus tools.

dafydd-t:
> so closing the body early (context cancellation, timeout) drops the measurement entirely.

Thats what we want, I think. We're not recording status here, just the size of (successful) responses


>  ● 187 [correctness] recordPrometheusMetricNames is only called from prometheusBackend.MetricNames (prom_backend.go:191).
                       cloudMonitoringBackend.MetricNames at this line never calls it, creating an observability blind spot for CM
                       datasources

I'll look at adding it there

>  ● 216 [correctness] When limit > maxPrometheusMetricNames (e.g. 15000) and page=1, integer division makes the check pass (1 >
                       10000/15000 = 0), but the error says 'page * limit must not exceed 10000; narrow the regex or request an
                       earlier page' -- misleading because the limit alone is the problem.

I'll improve the error

>    ● 257 [correctness] The list_prometheus_metric_names tool description says it supports 'an optional time range to restrict
                       results to metrics active within that window' but cloudMonitoringBackend.MetricNames ignores both time
                       parameters (named _ in the signature), returning all metrics.

True, but this is not new. maybe for another PR

>    ● 122 [efficiency]  prometheusMetricNamesTransport is applied to the entire Prometheus API client, so every Query, LabelNames,
                       LabelValues call passes through its path-suffix check. The overhead is negligible but a bug in the
                       transport would break all Prometheus tools.

It feels more obvious to wrap all calls, so I think this is a reasonable tradeoff IMO

dafydd-t:
addressed the Cloud Monitoring telemetry point from https://github.com/grafana/mcp-grafana/pull/1217#pullrequestreview-5266201399 in 4868265

both backends now record native histograms under `mcp_metric_names_response_size_bytes` and `mcp_metric_names_count`, with `backend="prometheus"` or `backend="cloud_monitoring"`. Cloud Monitoring records the full descriptor response before local filtering/pagination; metadata queries using the same endpoint are excluded. Added tests for that exclusion, backend labels, and successful/error/incomplete responses. `make lint` and `make test-unit` pass

dafydd-t:
addressed the two wording points from https://github.com/grafana/mcp-grafana/pull/1217#pullrequestreview-5266201399 in 148233b:

- pagination error now says `page * limit must not exceed 10000; reduce page or limit`.
- the tool description now explicitly says Cloud Monitoring ignores the time range

sd2k (approved):
Thanks 🙏

